PR #4, opened 2020-12-07
Modified to allow changes to the existing StorageCluster

kesavanvt:
**Description**

New (fine-tuned) StorageCluster is fetched from the configmap

- User with cluster admin access creates a configmap called `storagecluster` with new StorageCluster yaml in openshift-storage namespace. 

     `kubectl create configmap storagecluster --from-file storagecluster.yml`

- The ocs-osd-deployer pod fetches the configmap for applying StorageCluster if exists. If not, existing StorageCluster is applied.

### DIFF
--- a/deployer/deploy.sh
+++ b/deployer/deploy.sh
@@ -22,9 +22,16 @@ osds=$((${QUOTA_COUNT:-1} * 3))
 
 echo "Quota count: ${QUOTA_COUNT} -- OSD count: ${osds}"
 
+#-- The path mapped to StorageCluster configmap used for overriding.
+overridePath="/sc-override/storagecluster.yml"
+
 #-- Patch the OSD count into the StorageCluster & apply it
+#-- If config map containing StorageCluster exists, apply it. 
+#-- Else, apply the default StorageCluster.
 while true; do
-    sed "s/STORAGE_NODES/${osds}/" storagecluster.yml | \
+    storageClusterPath="storagecluster.yml"
+    [ -f $overridePath ] && storageClusterPath=$overridePath
+    sed "s/STORAGE_NODES/${osds}/" $storageClusterPath | \
       kubectl -n openshift-storage apply -f -
     sleep 60
 done
--- a/patcher/csv_fragment.yml
+++ b/patcher/csv_fragment.yml
@@ -30,4 +30,12 @@ spec:
                     name: deployer
                     image: quay.io/johnstrunk/ocs-osd-deployer:latest
                     imagePullPolicy: Always
+                    volumeMounts:
+                      - mountPath: /sc-override
+                        name: config-volume
                 serviceAccountName: ocs-osd-deployer
+                volumes:
+                  - configMap:
+                      name: sc-config
+                      optional: true
+                    name: config-volume


deployer/deploy.sh
```diff
@@ -23,9 +23,19 @@ osds=$((${QUOTA_COUNT:-1} * 3))
...
-    sed "s/STORAGE_NODES/${osds}/" storagecluster.yml | \
-      kubectl -n openshift-storage apply -f -
+    sc_content=`kubectl get configmaps storagecluster -n openshift-storage -oyaml 2> /dev/null`
+    if [ $? -ne 0 ]
+    then
+        sed "s/STORAGE_NODES/${osds}/" storagecluster.yml | \
+        kubectl -n openshift-storage apply -f -
+    else
+        echo "configmap is used to merge StorageCluster properties"
+        echo "$sc_content" | yq r - "data.*" | yq merge - storagecluster.yml | \
```
JohnStrunk, line 36: This appears to be missing the substitution of storage node count from L32.

Do you really want to merge with the storagecluster file in the container or just provide a full override (i.e., is the CM just going to have a subset of the storagecluster object)?

nb-ohad: @JohnStrunk we agreed to go with override instead of merge because it is much easier to understand (from a usage perspective). So this will not be merged in the current form

nb-ohad (changes requested):
I will suggest mapping the config map entry to an optional local mount. This way you just read it from disk. If you can mount it on top of the existing file (same location) you will not even have to touch the loop code

deployer/Dockerfile
```diff
@@ -8,6 +8,8 @@ RUN microdnf update -y && \
     && microdnf clean all && \
     rm -rf /var/cache/yum
 
+RUN mkdir -p sc-override
```
nb-ohad, line 11: Do we really need this, doesn't the config map to mount mapping take care of it?

deployer/deploy.sh
```diff
@@ -23,8 +23,12 @@ osds=$((${QUOTA_COUNT:-1} * 3))
 echo "Quota count: ${QUOTA_COUNT} -- OSD count: ${osds}"
 
 #-- Patch the OSD count into the StorageCluster & apply it
+#-- If config map containing StorageCluster exists, apply it. 
+#-- Else, apply the default StorageCluster.
 while true; do
-    sed "s/STORAGE_NODES/${osds}/" storagecluster.yml | \
+    storageClusterPath="storagecluster.yml"
+    [ -f /sc-override/storagecluster.yml ] && storageClusterPath="/sc-override/storagecluster.yml"
```
nb-ohad, line 30: Can we move `/sc-override/storagecluster.yml` into an env var? it is used more than once

nb-ohad (changes requested):

deployer/Dockerfile
kesavanvt: Thanks. Yes, the config map takes care of it.

deployer/deploy.sh
kesavanvt: fixed!

nb-ohad (approved):

deployer/deploy.sh
```diff
@@ -22,9 +22,16 @@ osds=$((${QUOTA_COUNT:-1} * 3))
...
 
+#-- The path mapped to StorageCluster configmap used for overriding.
+overridePath="/sc-override/storagecluster.yml"
+
 #-- Patch the OSD count into the StorageCluster & apply it
+#-- If config map containing StorageCluster exists, apply it. 
+#-- Else, apply the default StorageCluster.
 while true; do
-    sed "s/STORAGE_NODES/${osds}/" storagecluster.yml | \
+    storageClusterPath="storagecluster.yml"
```
nbalacha, line 32: Does this need to be done inside the loop?

kesavanvt: Yes, it should fallback to the default StorageCluster if the config map is removed

nbalacha: This could lead to issues - what if the OSD size is changed in the config map , the cluster is created and then the config map is deleted?

nb-ohad: The ability to use a config map to override the `storagecluster` is not meant to be available to anybody but the SREs who are supporting OCS on OSD, as such this is a crude tool that does not cover all cases.
It will be used with extreme caution and only after consultation.

Another point here is that this is just a temporary solution and a more wholesome solution, in the form of an operator based on operator-sdk, is coming in the near future.

@nbalacha I will just like to point out that the same issue can be reproduced by directly editing a `sotragecluster` CR, so the risk is already present on every OCS installation.

nbalacha: That is true but deleting a ConfigMap seems a lot easier than editing the file, especially when there is a UI in the OCP console to prevent it.

nb-ohad: In principle, I agree but please take into account that the barrier will not be deleting the config map but getting admin access to the `openshift-sotrage` namespace in the first place.
In an OSD environment, there is literally no admin access to any openshift namespace.

PR #10, opened 2021-01-04
Fix broken make commands

nb-ohad:
- Update `kustomize` to fix unwanted namespace name change (see https://github.com/kubernetes-sigs/kustomize/pull/2742)
- Copy new folders (utils and templates) in docker file
- Rename the default image from `controller` to `ocs-osd-deployer`
- Rename the resource prefix from `ocs-osd-deployer-` to ocs-osd-`
- Change the deploy namespace from `ocs-osd-deployer-system` to `openshift-storage`
- Remove `foo:bar` from ManagedOCS example resource

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -9,7 +9,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: openshift-storage
+  namespace: system
   labels:
     control-plane: controller-manager
 spec:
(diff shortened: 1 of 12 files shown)


config/manager/manager.yaml
```diff
@@ -9,7 +9,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: controller-manager
-  namespace: openshift-storage
+  namespace: system
```
JohnStrunk, line 12: Shouldn't this be `openshift-storage` to match the Namespace, above?

nb-ohad: no the namespace here is modified/overriden by `kustomize` so it should stay `system`

PR #12, opened 2021-01-27
Watch the addon secret for config changes

nbalacha:
Signed-off-by: N Balachandran <nibalach@redhat.com>

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -41,4 +41,6 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: ADDON_PARAMS_SECRET
+          value: "addon-ocs-converged-parameters"
       terminationGracePeriodSeconds: 10
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,18 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
   - ocs.openshift.io
   resources:
   - managedocs
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -11,12 +11,8 @@ rules:
   resources:
   - secrets
   verbs:
-  - create
-  - delete
   - get
   - list
-  - patch
-  - update
   - watch
 - apiGroups:
   - ocs.openshift.io
@@ -38,3 +34,9 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - ocs.openshift.io
+  resources:
+  - storageclusters
+  verbs:
+  - '*'
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -26,7 +26,6 @@ import (
 	"github.com/openshift/ocs-osd-deployer/templates"
 	"github.com/openshift/ocs-osd-deployer/utils"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -40,15 +39,8 @@ import (
 )
 
 const (
-	storageClusterName = "ocs-storagecluster"
-)
-
-// AddonSecretName is the secret name
-const (
-	AddonSecretName string = "addon-ocs-converged-parameters"
-)
-
-const (
+	storageClusterName         = "ocs-storagecluster"
+	addonSecretName     string = "addon-ocs-converged-parameters"
 	storageClassSizeKey string = "size"
 )
 
@@ -72,32 +64,16 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	mapFn := handler.ToRequestsFunc(
 		func(obj handler.MapObject) []reconcile.Request {
-			return []reconcile.Request{
-				{NamespacedName: types.NamespacedName{
-					Name:      "managedocs",
-					Namespace: obj.Meta.GetNamespace(),
-				}},
+			if obj.Meta.GetName() == r.ConfigSecretName {
+				return []reconcile.Request{
+					{NamespacedName: types.NamespacedName{
+						Name:      "managedocs",
+						Namespace: obj.Meta.GetNamespace(),
+					}},
+				}
 			}
+			return nil
 		})
-
-	/* 	p := predicate.Funcs{
-		UpdateFunc: func(e event.UpdateEvent) bool {
-			// The object name is not the secret name, so the event will be
-			// ignored.
-			name := e.MetaOld.GetName()
-			if name != AddonSecretName {
-				return false
-			}
-			return e.ObjectOld != e.ObjectNew
-		},
-		CreateFunc: func(e event.CreateEvent) bool {
-			name := e.Meta.GetName()
-			if name != AddonSecretName {
-				return false
-			}
-			return true
-		},
-	} */
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1.ManagedOCS{}).
@@ -114,7 +90,7 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) 
 	log := r.Log.WithValues("req.Namespace", req.Namespace, "req.Name", req.Name)
 	log.Info("Reconciling ManagedOCS")
 
-	var err error
+	//var err error
 	r.ctx = context.Background()
 
 	// Load the managed ocs resource
@@ -124,7 +100,7 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) 
 	}
 
 	// Run the reconcile phases
-	err = r.reconcilePhases(req)
+	err := r.reconcilePhases(req)
 
 	// Ensure status is updated once even on failed reconciles
 	statusErr := r.Status().Update(r.ctx, r.managedOCS)
@@ -147,15 +123,6 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
 	}
 	r.managedOCS.Status.ReconcileStrategy = reconcileStrategy
 
-	configSecret := &corev1.Secret{}
-	err := r.Get(r.ctx, types.NamespacedName{Namespace: req.Namespace, Name: r.ConfigSecretName}, configSecret)
-	if err != nil {
-		if !errors.IsNotFound(err) {
-			return err
-		} else {
-			configSecret = nil
-		}
-	}
 	// Create or update an existing storage cluster
 	storageCluster := &ocsv1.StorageCluster{
 		ObjectMeta: metav1.ObjectMeta{
@@ -164,7 +131,7 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
 		},
 	}
 	if _, err := ctrlutil.CreateOrUpdate(r.ctx, r, storageCluster, func() error {
-		return r.setDesiredStorageCluster(reconcileStrategy, storageCluster, configSecret)
+		return r.setDesiredStorageCluster(reconcileStrategy, storageCluster)
 	}); err != nil {
 		return err
 	}
@@ -175,8 +142,7 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
 // Set the desired stats for the storage cluster resource
 func (r *ManagedOCSReconciler) setDesiredStorageCluster(
 	reconcileStrategy v1.ReconcileStrategy,
-	sc *ocsv1.StorageCluster,
-	cs *corev1.Secret) error {
+	sc *ocsv1.StorageCluster) error {
 	r.Log.Info("Reconciling storagecluster", "ReconcileStrategy", reconcileStrategy)
 
 	// Ensure ownership on the storage cluster CR
@@ -186,53 +152,28 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
 
 	// Handle strict mode reconciliation
 	if reconcileStrategy == v1.ReconcileStrategyStrict {
+		configSecret := &corev1.Secret{}
+		if err := r.Get(r.ctx, types.NamespacedName{Namespace: sc.ObjectMeta.Namespace, Name: r.ConfigSecretName}, configSecret); err != nil {
+			// Do not create the StorageCluster if the addon param secret is missing
+			r.Log.Info("Failed to get the addon param secret. Not reconciling the storage cluster.", "SecretName", r.ConfigSecretName)
+
+			return err
+		}
+		sdata := configSecret.Data
+
 		// Get an instance of the desired state
 		desired := utils.ObjectFromTemplate(templates.StorageClusterTemplate, r.Scheme).(*ocsv1.StorageCluster)
-		r.updateStorageClusterWithParams(desired, cs)
+
+		PersistentVolume := &desired.Spec.StorageDeviceSets[0].DataPVCTemplate
+		PersistentVolume.Spec.Resources = corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				"storage": resource.MustParse(string(sdata[storageClassSizeKey])),
+			},
+		}
 		// Override storage cluster spec with desired spec from the template.
 		// We do not replace meta or status on purpose
 		sc.Spec = desired.Spec
-		r.Log.Info("StorageCluster", "desiredSpec", desired.Spec)
-		r.Log.Info("StorageCluster", "Current spec", sc.Spec)
 	}
 
 	return nil
-}
-
-func (r *ManagedOCSReconciler) updateStorageClusterWithParams(
-	sc *ocsv1.StorageCluster,
-	cs *corev1.Secret) error {
-
-	if cs == nil {
-		return nil
-	}
-	r.Log.Info("Updating storagecluster")
-
-	sdata := cs.Data
-
-	for key, element := range sdata {
-		r.Log.Info("Secret keys:", "key", key, "value", string(element))
-		PersistentVolume := &sc.Spec.StorageDeviceSets[0].DataPVCTemplate
-		if isValidStorageClusterParam(key, string(sdata[key])) {
-			PersistentVolume.Spec.Resources = corev1.ResourceRequirements{
-				Requests: corev1.ResourceList{
-					"storage": resource.MustParse(string(sdata[key])),
-				},
-			}
-		}
-		r.Log.Info("Post updation Persistent Volume ", "PV", PersistentVolume)
-	}
-
-	return nil
-}
-
-func isValidStorageClusterParam(key string, val string) bool {
-
-	if key == storageClassSizeKey {
-		if val != "1Ti" && val != "4Ti" {
-			return false
-		}
-		return true
-	}
-	return false
 }
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -39,25 +40,25 @@ import (
 )
 
 const (
-	storageClusterName         = "ocs-storagecluster"
-	addonSecretName     string = "addon-ocs-converged-parameters"
-	storageClassSizeKey string = "size"
+	storageClusterName   = "ocs-storagecluster"
+	addonParamSecretName = "addon-ocs-converged-parameters"
+	storageClassSizeKey  = "size"
 )
 
 // ManagedOCSReconciler reconciles a ManagedOCS object
 type ManagedOCSReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
-	ctx    context.Context
-
+	Log              logr.Logger
+	Scheme           *runtime.Scheme
+	ctx              context.Context
 	managedOCS       *v1.ManagedOCS
 	ConfigSecretName string
 }
 
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs/status,verbs=get;update;patch
-// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusters,verbs=*
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;
 
 // SetupWithManager TODO
 func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
@@ -80,7 +81,6 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&ocsv1.StorageCluster{}).
 		Watches(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{
 			ToRequests: mapFn},
-		//p,
 		).
 		Complete(r)
 }
@@ -90,7 +90,6 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) 
 	log := r.Log.WithValues("req.Namespace", req.Namespace, "req.Name", req.Name)
 	log.Info("Reconciling ManagedOCS")
 
-	//var err error
 	r.ctx = context.Background()
 
 	// Load the managed ocs resource
@@ -147,6 +146,7 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
 
 	// Ensure ownership on the storage cluster CR
 	if err := ctrlutil.SetControllerReference(r.managedOCS, sc, r.Scheme); err != nil {
+		r.Log.Error(err, "Failed to set controller reference on the managedOCS resource")
 		return err
 	}
 
@@ -154,21 +154,14 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
 	if reconcileStrategy == v1.ReconcileStrategyStrict {
 		configSecret := &corev1.Secret{}
 		if err := r.Get(r.ctx, types.NamespacedName{Namespace: sc.ObjectMeta.Namespace, Name: r.ConfigSecretName}, configSecret); err != nil {
-			// Do not create the StorageCluster if the addon param secret is missing
-			r.Log.Info("Failed to get the addon param secret. Not reconciling the storage cluster.", "SecretName", r.ConfigSecretName)
-
+			// Do not create the StorageCluster if the we fail to get the addon param secret
+			r.Log.Error(err, "Failed to get the addon param secret. Not reconciling the storage cluster.", "SecretName", r.ConfigSecretName)
 			return err
 		}
-		sdata := configSecret.Data
 
-		// Get an instance of the desired state
-		desired := utils.ObjectFromTemplate(templates.StorageClusterTemplate, r.Scheme).(*ocsv1.StorageCluster)
-
-		PersistentVolume := &desired.Spec.StorageDeviceSets[0].DataPVCTemplate
-		PersistentVolume.Spec.Resources = corev1.ResourceRequirements{
-			Requests: corev1.ResourceList{
-				"storage": resource.MustParse(string(sdata[storageClassSizeKey])),
-			},
+		desired, err := r.generateStorageClusterWithParams(configSecret)
+		if err != nil {
+			return err
 		}
 		// Override storage cluster spec with desired spec from the template.
 		// We do not replace meta or status on purpose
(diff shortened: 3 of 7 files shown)


controllers/managedocs_controller.go
```diff
@@ -63,6 +114,7 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	log := r.Log.WithValues("req.Namespace", req.Namespace, "req.Name", req.Name)
 	log.Info("Reconciling ManagedOCS")
 
+	var err error
```
nb-ohad, line 117: I would like to have an err var defined (every time) when needed and not globally, it allows us to track the right var when watching dumps (each instance will have a diff memory address)

controllers/managedocs_controller.go
```diff
@@ -95,6 +147,15 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
 	}
 	r.managedOCS.Status.ReconcileStrategy = reconcileStrategy
 
+	configSecret := &corev1.Secret{}
+	err := r.Get(r.ctx, types.NamespacedName{Namespace: req.Namespace, Name: r.ConfigSecretName}, configSecret)
+	if err != nil {
```
nb-ohad, line 152: Two notes here:
1.  Can we merge these into the if? It's a go best practice 
```if err := r.Get(r.ctx, types.NamespacedName{Namespace: req.Namespace, Name: r.ConfigSecretName}, configSecret); err != nil {```

2. Can we move this inside the `setDesiredStorageCluster` func to prevent this logic (and reading of the secret) in case the reconcile strategy is `none`?

File: controllers/managedocs_controller.go [hunk 3]
```diff
@@ -95,6 +147,15 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
...
+		if !errors.IsNotFound(err) {
+			return err
```
nb-ohad, line 154: Let's add a print to log here (error)

controllers/managedocs_controller.go
```diff
@@ -95,6 +147,15 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
...
+		} else {
+			configSecret = nil
```
nb-ohad, line 156: Let's add a print to log here (info) and return to prevent reconciliation if the secret does not exists

controllers/managedocs_controller.go
```diff
@@ -126,11 +188,51 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
 }
+
+func (r *ManagedOCSReconciler) updateStorageClusterWithParams(
+	sc *ocsv1.StorageCluster,
+	cs *corev1.Secret) error {
+
+	if cs == nil {
+		return nil
+	}
+	r.Log.Info("Updating storagecluster")
```
nb-ohad, line 209: This is a misleading log, as we are not updating the storage cluster here at all, the updating happens when we replace the sc spec

controllers/managedocs_controller.go
```diff
@@ -126,11 +188,51 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
 	return nil
 }
+
+func (r *ManagedOCSReconciler) updateStorageClusterWithParams(
+	sc *ocsv1.StorageCluster,
+	cs *corev1.Secret) error {
+
+	if cs == nil {
+		return nil
+	}
```
nb-ohad, line 208: Should not happen, we should bail out of a reconcile if the secret does not exist earlier

controllers/managedocs_controller.go
```diff
@@ -126,11 +188,51 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+		r.Log.Info("Secret keys:", "key", key, "value", string(element))
+		PersistentVolume := &sc.Spec.StorageDeviceSets[0].DataPVCTemplate
+		if isValidStorageClusterParam(key, string(sdata[key])) {
+			PersistentVolume.Spec.Resources = corev1.ResourceRequirements{
+				Requests: corev1.ResourceList{
+					"storage": resource.MustParse(string(sdata[key])),
+				},
+			}
+		}
+		r.Log.Info("Post updation Persistent Volume ", "PV", PersistentVolume)
```
nb-ohad, line 223: Why do we need this generic code? can we just do a simple read of predefined keys from the secret? 
It will be easier to read, understand and even handle special cases (for future params) more easily (for example 2 params that depend on each other)

```
storageClassSize := cs.data[storageClassSizeKey]
if storageClassSize != "1TiB" || storageClassSize != "4TiB" {
    return r.Log.Ferrorf("Invalid storage class size, %s = %s", storageClassSizeKey, storageClassSize)
}
PersistentVolume := &sc.Spec.StorageDeviceSets[0].DataPVCTemplate
PersistentVolume.Spec.Resources["storage"] = resource.MustParse(storageClassSize)
```

For that matter, I suggest that this whole code section should be folded back into the `setDesiredStorageCluster` which is the place to update the storage cluster

main.go
```diff
@@ -119,6 +121,16 @@ func getNamespace() (string, error) {
...
+// The name of the secret containing the addon params
+func getConfigName() string {
+
+	configName, found := os.LookupEnv(configSecretEnvVar)
+	if !found {
+		configName = controllers.AddonSecretName
+	}
+	return configName
+}
+
```
nb-ohad, line 133: This is the wrong place to do it as the value should be read from the `managedOCS` CR (with a default value, of course)

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -20,41 +20,92 @@ import (
...
 // ManagedOCSReconciler reconciles a ManagedOCS object
 type ManagedOCSReconciler struct {
 	client.Client
 	Log    logr.Logger
 	Scheme *runtime.Scheme
 	ctx    context.Context
 
-	managedOCS *v1.ManagedOCS
+	managedOCS       *v1.ManagedOCS
+	ConfigSecretName string
```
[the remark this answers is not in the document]
nbalacha: It is not there in the managedOCS spec.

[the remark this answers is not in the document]
nb-ohad: The original idea was to have it inside the ManagedOCS spec, but, if you remember, last week we decided to move it to an env var, which means that this comment is obsolete

controllers/managedocs_controller.go
nbalacha: This was done in this function in order to get the req.namespace to get the secret.
The other option is to pass the request as an argument to setDesiredStorageCluster.

controllers/managedocs_controller.go
nbalacha: This was part of the code I was using to understand the way the secret worked as part of the draft code.

main.go
nbalacha: It was decided to use an env value to store this instead.

nb-ohad: This is correct, we decided to have 2 secrets, the first one will point at the secret, the second one will be used to get the storage unit per OSD

main.go
```diff
@@ -68,12 +74,19 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
+	var envVars EnvVars
+
 	namespace, err := getNamespace()
```
nbalacha, line 79: Did not put the namespace in the envVars struct. Can be done if required.

controllers/managedocs_controller.go
nb-ohad: or same the namespace on the reconciler
```
r.namespace = req.namespace
```

controllers/managedocs_controller.go
nb-ohad: Ack

nbalacha:
The updates do not include changes to update the managedOCS status with the params.

controllers/managedocs_controller.go
```diff
@@ -20,21 +20,28 @@ import (
...
+	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 const (
-	storageClusterName = "ocs-storagecluster"
+	storageClusterName         = "ocs-storagecluster"
+	addonSecretName     string = "addon-ocs-converged-parameters"
```
nb-ohad, line 43: I think this is a leftover from the old code. I don't see any code using it.

controllers/managedocs_controller.go
```diff
@@ -44,17 +51,37 @@ type ManagedOCSReconciler struct {
...
 	ctx    context.Context
 
-	managedOCS *v1.ManagedOCS
+	managedOCS       *v1.ManagedOCS
+	ConfigSecretName string
 }
 
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
nb-ohad: I second that and believe we only need `get`, `list`, and `watch`

controllers/managedocs_controller.go
```diff
@@ -124,9 +152,24 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
 
 	// Handle strict mode reconciliation
 	if reconcileStrategy == v1.ReconcileStrategyStrict {
+		configSecret := &corev1.Secret{}
+		if err := r.Get(r.ctx, types.NamespacedName{Namespace: sc.ObjectMeta.Namespace, Name: r.ConfigSecretName}, configSecret); err != nil {
+			// Do not create the StorageCluster if the addon param secret is missing
+			r.Log.Info("Failed to get the addon param secret. Not reconciling the storage cluster.", "SecretName", r.ConfigSecretName)
```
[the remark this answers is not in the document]
nb-ohad: Also can we please incorporate the error message into the log entry?

controllers/managedocs_controller.go
```diff
@@ -124,9 +152,24 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+
 		// Get an instance of the desired state
 		desired := utils.ObjectFromTemplate(templates.StorageClusterTemplate, r.Scheme).(*ocsv1.StorageCluster)
 
+		PersistentVolume := &desired.Spec.StorageDeviceSets[0].DataPVCTemplate
+		PersistentVolume.Spec.Resources = corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
+				"storage": resource.MustParse(string(sdata[storageClassSizeKey])),
+			},
+		}
```
nb-ohad, line 172: The agreed way to have a specific size is by provisioning more 1TiB OSDS not by asking for a bigger device set. 
This also mandates the following missing checks:
1. Validate that the size is a multiply of 1TiB
2. Validate that the size is not already set in the actual storage cluster (we do not support resize at the moment)

Both conditions should log a message and terminate the reconcile loop with an  error

controllers/managedocs_controller_test.go
```diff
@@ -61,6 +63,23 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				secret := &corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      ConfigSecretName,
+						Namespace: TestNamespace,
+					},
+					Data: data,
+				}
+				Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+				Expect(k8sClient.Get(ctx, getResourceKey(secret), secret)).Should(Succeed())
+
```
nb-ohad, line 82: This should be a separate test case that runs even before the  first one

controllers/managedocs_controller_test.go
```diff
@@ -18,7 +19,8 @@ import (
 var _ = Describe("ManagedOCS controller", func() {
 	// Define utility constants for object names and testing timeouts/durations and intervals.
 	const (
-		ManagedOCSName = "test-managedocs"
+		ManagedOCSName   = "test-managedocs"
+		ConfigSecretName = "test-secret"
```
nb-ohad, line 23: This const should be defined on the test_suite to be shared with all tests. 
Also, the name is misleading, let's call it something like  `AddOnParamsSecretName`

controllers/managedocs_controller.go
```diff
@@ -44,17 +51,37 @@ type ManagedOCSReconciler struct {
 	Scheme *runtime.Scheme
 	ctx    context.Context
 
-	managedOCS *v1.ManagedOCS
+	managedOCS       *v1.ManagedOCS
+	ConfigSecretName string
```
nb-ohad, line 55: I think this is not a good name for the field. The field is not a general config nor is it the only configuration source.
As such I would prefer something more specific like `AddOnParamsSecretName`

controllers/suite_test.go
```diff
@@ -81,9 +81,10 @@ var _ = BeforeSuite(func(done Done) {
 	Expect(err).ToNot(HaveOccurred())
 
 	err = (&ManagedOCSReconciler{
-		Client: k8sManager.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("ManagedOCS"),
-		Scheme: scheme.Scheme,
+		Client:           k8sManager.GetClient(),
+		Log:              ctrl.Log.WithName("controllers").WithName("ManagedOCS"),
+		Scheme:           scheme.Scheme,
+		ConfigSecretName: "test-secret",
```
nb-ohad, line 87: I would suggest moving the value into a public const at the top of the file to be used both here and on every test suite where it is needed.

main.go
```diff
@@ -40,14 +40,20 @@ import (
...
 
 var (
 	scheme   = runtime.NewScheme()
 	setupLog = ctrl.Log.WithName("setup")
 )
 
+// EnvVars struct holds environment vars
+type EnvVars struct {
+	configName string
+}
```
nb-ohad, line 55: Why a dedicated struct?, why not just a `map[string]string`.

main.go
```diff
@@ -40,14 +40,20 @@ import (
 )
 
 const (
-	namespaceEnvVar = "NAMESPACE"
+	namespaceEnvVar    = "NAMESPACE"
+	configSecretEnvVar = "CONFIG"
```
nb-ohad, line 44: I think the naming here is problematic. 

First of all the vars are not env vars but env vars name so a better prefix should be `EnvVarName` (I did not understand it was the name until I read the rest of the code which is a bad sign for readability)

Second, the "CONFIG" env var is incorrectly named, this is not the name of the env var that holds some general config, or the only config details for that matter. It is the env var that specifies the name of the AddOnParamsSecret and should be named as such, for example 
```addonParamsSecretEnvVarName = "ADDON_PARAMS_SECRET"```

But I have another more "radical" suggestion: Why not just hold an env var for the add-on name, like
```addonNameEnvVarName="ADDON_NAME"```
Now we can use it to build the secret name below and moving forward also use it for other tasks that need the addon name (For example reading the config-map that is used to indicate uninstallation)

File: main.go [hunk 21]
```diff
@@ -68,12 +74,19 @@ func main() {
...
 
+	var envVars EnvVars
+
 	namespace, err := getNamespace()
 	if err != nil {
 		setupLog.Error(err, "unable to get namespace name")
 		os.Exit(1)
 	}
 
+	if err = getEnvVars(&envVars); err != nil {
```
nb-ohad, line 85: Can we just return the envVars from the function, I do not see the need to send a struct to the function

main.go
```diff
@@ -119,6 +133,19 @@ func getNamespace() (string, error) {
 	return ns, nil
 }
 
+// The name of the secret containing the addon params
+func getEnvVars(envVars *EnvVars) error {
```
nb-ohad, line 137: Would prefer to have this method read all env vars, this includes the namespace.

main.go
```diff
@@ -68,12 +74,19 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
+	var envVars EnvVars
+
 	namespace, err := getNamespace()
 	if err != nil {
 		setupLog.Error(err, "unable to get namespace name")
 		os.Exit(1)
 	}
```
nb-ohad, line 83: It is just another env var, let's merge

main.go
```diff
@@ -68,12 +74,19 @@ func main() {
...
 
+	if err = getEnvVars(&envVars); err != nil {
+		setupLog.Error(err, "unable to get required paramaters")
```
nb-ohad, line 86: The error is sufficient, you can remove the added text

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
nbalacha: In that case should that be done for the other names as well (namespace, managedOcs etc)

controllers/suite_test.go
nbalacha: Done

controllers/managedocs_controller.go
nbalacha: I don't think we need the second check - the storageCluster cannot be created without a size

nb-ohad: the second check is to validate that the requested size is not different than what was set previously.  In that case we should abort

controllers/managedocs_controller_test.go
nb-ohad: If we need to use a const both on more than one test file or on a test file and the test_suite it should be defined in a single location

controllers/managedocs_controller.go
nbalacha: In that case, should we just check if the storagecluster already exists and abort without checking for the secret?
If we cannot change the sc, there doesn't seem to be a point to watching the secret.

nb-ohad: The `CreateOrUpdate` callback does not and should not know if it exists or not. What we can do is check if the loaded resource has a size or not. Of course, this assumes that we do not have a default size in the template (which we should not).

This way it is easier to implement, more consistent with how reconciles are designed to work and, solves the problem at least as good as checking the existence of the storage cluster (I am aware that both methodologies are not perfect, but they have the same drawbacks)

This is also better groundwork  for the upcoming change that will enable us to support upscaling

controllers/managedocs_controller.go
```diff
@@ -20,41 +20,92 @@ import (
...
 
+// AddonSecretName is the secret name
+const (
+	AddonSecretName string = "addon-ocs-converged-parameters"
+)
+
+const (
+	storageClassSizeKey string = "size"
+)
+
```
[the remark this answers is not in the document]
nbalacha: done

controllers/managedocs_controller.go
nbalacha: done

controllers/managedocs_controller.go
nbalacha: done

controllers/managedocs_controller.go
```diff
@@ -20,21 +20,28 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 const (
-	storageClusterName = "ocs-storagecluster"
+	storageClusterName         = "ocs-storagecluster"
```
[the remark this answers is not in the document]
nbalacha: done

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nbalacha: done

controllers/managedocs_controller.go
```diff
@@ -44,17 +51,37 @@ type ManagedOCSReconciler struct {
...
+			}
+			return nil
+		})
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1.ManagedOCS{}).
 		Owns(&ocsv1.StorageCluster{}).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: mapFn},
+		//p,
```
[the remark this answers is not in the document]
nbalacha: done.

controllers/managedocs_controller.go
```diff
@@ -63,6 +90,7 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	log := r.Log.WithValues("req.Namespace", req.Namespace, "req.Name", req.Name)
 	log.Info("Reconciling ManagedOCS")
 
+	//var err error
```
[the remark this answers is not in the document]
nbalacha: done.

controllers/managedocs_controller.go
nbalacha: We could get the storagecluster object itself before calling CreateOrUpdate to check if it exists - if it does, don't do anything.

controllers/managedocs_controller_test.go
nbalacha: done.

main.go
nbalacha: done.

main.go
nbalacha: done.

config/rbac/role.yaml
```diff
@@ -38,3 +34,9 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - ocs.openshift.io
+  resources:
+  - storageclusters
+  verbs:
+  - '*'
```
nb-ohad, line 42: This will conflict with the recent add-on PR for RBAC changes. 
If we can wait for it to be merged this changed will be obsolete.

controllers/managedocs_controller.go
```diff
@@ -39,25 +40,25 @@ import (
 )
 
 const (
-	storageClusterName         = "ocs-storagecluster"
-	addonSecretName     string = "addon-ocs-converged-parameters"
-	storageClassSizeKey string = "size"
+	storageClusterName   = "ocs-storagecluster"
+	addonParamSecretName = "addon-ocs-converged-parameters"
```
nb-ohad, line 44: The actual add-on name should not be a const in the code as this codebase will back all the add-ons for our offering. Even now this const will conflict when deploying with the `ocs-converged-dev` add-on

controllers/managedocs_controller.go
```diff
@@ -39,25 +40,25 @@ import (
...
+	Scheme           *runtime.Scheme
+	ctx              context.Context
 	managedOCS       *v1.ManagedOCS
 	ConfigSecretName string
 }
 
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs/status,verbs=get;update;patch
-// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,resources=storageclusters,verbs=*
```
nb-ohad, line 60: Same as my last RBAC related comment

controllers/managedocs_controller.go
```diff
@@ -177,3 +170,34 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+func (r *ManagedOCSReconciler) generateStorageClusterWithParams(configSecret *corev1.Secret) (*ocsv1.StorageCluster, error) {
+	sdata := configSecret.Data
+	requestedStorageCapacity, err := resource.ParseQuantity(string(sdata[storageClassSizeKey]))
+	if err != nil {
+		r.Log.Error(err, "Missing or incorrect parameter in the addonparam secret", "Key", storageClassSizeKey)
+		return nil, err
+	}
+	r.Log.Info("Addon param data", "Storage Cluster Size", requestedStorageCapacity)
+
+	osdSize, _ := resource.ParseQuantity("1Ti")
```
nb-ohad, line 183: Can we please move the 1Ti into a const

controllers/managedocs_controller.go
```diff
@@ -177,3 +170,34 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+
+	// The request must be an exact multiple of the base OSD size
+	rem := requestedStorageCapacity.Value() % osdSize.Value()
+
+	if rem != 0 {
+		err := fmt.Errorf("Storage cluster size must be a multiple of %v", osdSize)
+		return nil, err
+	}
+	count := requestedStorageCapacity.Value() / osdSize.Value()
```
nb-ohad, line 192: Why aren't we just treating the size param as a number which is a multiplication of 1Tib. If we do we will have no fail conditions

controllers/managedocs_controller.go
```diff
@@ -177,3 +170,34 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+	desired := utils.ObjectFromTemplate(templates.StorageClusterTemplate, r.Scheme).(*ocsv1.StorageCluster)
+
+	desired.Spec.StorageDeviceSets[0].Count = int(count)
+	persistentVolume := &desired.Spec.StorageDeviceSets[0].DataPVCTemplate
+	persistentVolume.Spec.Resources = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			"storage": osdSize,
+		},
+	}
```
nb-ohad, line 201: Why are we bundling this code in the same function as the size treatment? These are 2 separate concerns and should be kept separate. I would suggest creating the template from the outside and passing it to a function that updates the size. 

Currently, we have only one param but with the introduction of more we could just pass the storage cluster to a function for each param which will be responsible to update the template according to the specific param value

controllers/managedocs_controller_test.go
```diff
@@ -46,14 +45,67 @@ var _ = Describe("ManagedOCS controller", func() {
 	}
 
 	Context("reconcile()", func() {
-		When("there is no storage cluster resource in the cluster", func() {
-			It("should create a new storage cluster", func() {
+		When("There is no addonParam secret in the cluster", func() {
```
nb-ohad, line 48: Test descriptions should be human-readable, as such let's change 'addonParam secret"  to "add-on parameters secret"

controllers/managedocs_controller_test.go
```diff
@@ -46,14 +45,67 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				// No storage cluster should be created
+				scList = &ocsv1.StorageClusterList{}
+
+				Expect(k8sClient.List(ctx, scList, client.InNamespace(TestNamespace))).Should(Succeed())
+				Expect(scList.Items).Should(HaveLen(0))
+
+			})
+		})
+		When("there is incorrect data in the secret", func() {
+			It("should not create a new storage cluster", func() {
```
nb-ohad, line 88: This should never be a case in regards to size unless you are treating a string that does no contain a number a case.

main.go
```diff
@@ -40,20 +40,17 @@ import (
 )
 
 const (
-	namespaceEnvVar    = "NAMESPACE"
-	configSecretEnvVar = "CONFIG"
+	namespaceEnvVarName         = "NAMESPACE"
+	addonParamsSecretEnvVarName = "ADDON_PARAMS_SECRET"
```
nb-ohad, line 44: I was under the impression we agreed to have an ADDON_NAME env var and not ADDON_PARAMS_SECRET env var as it will also benefit the other add-on features we are working on

main.go
```diff
@@ -104,44 +100,50 @@ func main() {
 		Client:           mgr.GetClient(),
 		Log:              ctrl.Log.WithName("controllers").WithName("ManagedOCS"),
 		Scheme:           mgr.GetScheme(),
-		ConfigSecretName: envVars.configName,
+		ConfigSecretName: EnvVars[addonParamsSecretEnvVarName],
```
nb-ohad, line 103: `ConfigSecretName` - Please see my last review comments regarding naming.
I can see the naming changes in the tests code but not on the actual controller code

main.go
```diff
@@ -104,44 +100,50 @@ func main() {
...
+	ns, found := os.LookupEnv(namespaceEnvVarName)
 	if !found {
-		err := fmt.Errorf("%s must be set", namespaceEnvVar)
+		err := fmt.Errorf("%s must be set", namespaceEnvVarName)
 		return "", err
 	}
 	return ns, nil
 }
-
+*/
```
nb-ohad: Can we delete this

main.go
```diff
@@ -104,44 +100,50 @@ func main() {
...
 // The name of the secret containing the addon params
-func getEnvVars(envVars *EnvVars) error {
+func getEnvVars(m map[string]string) error {
```
nb-ohad, line 133: Why do we need to provide the map from the outside? can we just instantiate it here and return it?

templates/storagecluster.go
```diff
@@ -61,7 +61,7 @@ spec:
       requests: {}
   storageDeviceSets:
     - name: mydeviceset
-      count: 3 
+      count: 1
```
nb-ohad, line 64: Can we just remove it? It should be overridden in all cases

templates/storagecluster.go
```diff
@@ -70,9 +70,10 @@ spec:
           volumeMode: Block
           resources:
             requests:
-              storage: 1000Gi
+              storage: 1Ti
       placement: {}
       portable: true
+      replica: 3
```
nb-ohad, line 76: Same as prev comment

nb-ohad (changes requested):

controllers/managedocs_controller.go
nb-ohad: Of course we can, but to me what you describe sounds backward.

config/manager/manager.yaml
```diff
@@ -41,4 +41,6 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: metadata.namespace
+        - name: ADDON_PARAMS_SECRET
+          value: "addon-ocs-converged-parameters"
```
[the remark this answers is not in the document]
nbalacha: The original discussion was to use the secret name as a variable. Post the discussion today I will add the addon name

nbalacha:
Closing this as per Ohad's request. Will open a new PR.

controllers/managedocs_controller.go
nbalacha: Done

controllers/managedocs_controller.go
nbalacha: Done

controllers/managedocs_controller_test.go
nbalacha: Modified with the change in expected secret input

controllers/managedocs_controller.go
nbalacha: Treating the input as an integer will still fail if a non-integer is passed in. That check will need to be enforced in the UI

controllers/managedocs_controller.go
nbalacha: That will be taken care of when it is rebased.

controllers/managedocs_controller.go
nbalacha: Done

config/rbac/role.yaml
nbalacha: That will be taken care of with a rebase

controllers/managedocs_controller_test.go
nbalacha: Done

controllers/managedocs_controller.go
nbalacha: Done

PR #13, opened 2021-02-11
Modify rbac permissions and service account

kesavanvt:
**Description** 
This PR consists of following changes

- Add necessary rbac permissions.
- Modify to role based access.
- Modify service account from `default` to `ocs-osd-deployer `

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -1,15 +1,17 @@
 
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
+kind: Role
 metadata:
   creationTimestamp: null
   name: manager-role
+  namespace: system
 rules:
 - apiGroups:
   - ocs.openshift.io
   resources:
   - managedocs
+  - managedocs/finalizers
   verbs:
   - create
   - delete
@@ -26,3 +28,15 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - ocs.openshift.io
+  resources:
+  - storageclusters
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
--- a/config/rbac/service_account.yaml
+++ b/config/rbac/service_account.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: deployer
+  namespace: system
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -47,8 +47,10 @@ type ManagedOCSReconciler struct {
 	managedOCS *v1.ManagedOCS
 }
 
-// +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs/status,verbs=get;update;patch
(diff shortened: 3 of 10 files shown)


config/rbac/role.yaml
```diff
@@ -18,6 +19,18 @@ rules:
...
+  resources:
+  - managedocs/finalizers
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
nb-ohad, line 33: Can we fold it into the other instance of the `ocs.openshift.io` group?

controllers/managedocs_controller.go
```diff
@@ -47,8 +47,12 @@ type ManagedOCSReconciler struct {
...
 }
 
-// +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=ocs.openshift.io,resources=managedocs/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=managedocs,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=storageclusters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=managedocs/status,verbs=get;update;patch
+
+// Add necessary rbac permissions for managedocs finalizer in order to set blockOwnerDeletion.
+// +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=managedocs/finalizers,verbs=get;list;watch;create;update;patch;delete
```
nb-ohad, line 55: I am not sure `namespace=system` is the correct place for these RBAC

config/rbac/role.yaml
kesavanvt: Done

controllers/managedocs_controller.go
kesavanvt: Based on kubebuilder markers, the rules will set to Role only when a namespace is mentioned else the rules will set to ClusterRole. The namespace value will be replaced with openshift-storage by kustomize, thus leading to creation of roles in openshift-storage namespace.

nb-ohad: Have you verified this behaviour. If so I will approve the changes

kesavanvt: yes, verified using `make deploy` .

nb-ohad: Can you also check what is written to the bundle directory after `make manifests`?

kesavanvt: Based on my observation, `make manifests` doesn't modify any contents in bundle directory, it mainly updates rbac files in case of any changes.

config/rbac/service_account.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: deployer
+  namespace: system
```
nb-ohad, line 5: Why are creating the new service account in the `system` namespace? or is the make process patch the correct namespace here?

kesavanvt: Yes, the make process will substitute the correct namespace

nb-ohad (approved):

nb-ohad:
@JohnStrunk Can you please merge this PR?

PR #15, opened 2021-02-19
OCS OSD deployer - Addon Parameters 

nbalacha:
OSD stores addon parameters in a secret called
addon-<addon-name>-parameters.The OCS converged addon
controller will watch for this secret and update the
storagecluster resource with the provided parameters.

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -18,45 +18,71 @@ package controllers
 
 import (
 	"context"
+	"strconv"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	"github.com/go-logr/logr"
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 	v1 "github.com/openshift/ocs-osd-deployer/api/v1alpha1"
 	"github.com/openshift/ocs-osd-deployer/templates"
 	"github.com/openshift/ocs-osd-deployer/utils"
+	corev1 "k8s.io/api/core/v1"
 )
 
 const (
-	storageClusterName = "ocs-storagecluster"
+	storageClusterName  = "ocs-storagecluster"
+	storageClassSizeKey = "size"
+	baseOsdSize         = "1Ti"
 )
 
 // ManagedOCSReconciler reconciles a ManagedOCS object
 type ManagedOCSReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
-	ctx    context.Context
-
-	managedOCS *v1.ManagedOCS
+	Log                  logr.Logger
+	Scheme               *runtime.Scheme
+	ctx                  context.Context
+	managedOCS           *v1.ManagedOCS
+	AddonParamSecretName string
 }
 
 // Add necessary rbac permissions for managedocs finalizer in order to set blockOwnerDeletion.
 // +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources={managedocs,managedocs/finalizers},verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=managedocs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=storageclusters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",namespace=system,resources=secrets,verbs=get;list;watch
 
 // SetupWithManager TODO
 func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	mapFn := handler.ToRequestsFunc(
+		func(obj handler.MapObject) []reconcile.Request {
+			if obj.Meta.GetName() == r.AddonParamSecretName {
+				return []reconcile.Request{
+					{NamespacedName: types.NamespacedName{
+						Name:      "managedocs",
+						Namespace: obj.Meta.GetNamespace(),
+					}},
+				}
+			}
+			return nil
+		})
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1.ManagedOCS{}).
 		Owns(&ocsv1.StorageCluster{}).
+		Watches(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: mapFn},
+		).
 		Complete(r)
 }
 
@@ -105,7 +131,7 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
 		},
 	}
 	if _, err := ctrlutil.CreateOrUpdate(r.ctx, r, storageCluster, func() error {
-		return r.setDesiredStorageCluster(reconcileStrategy, storageCluster)
+		return r.generateStorageCluster(reconcileStrategy, storageCluster)
 	}); err != nil {
 		return err
 	}
@@ -114,13 +140,14 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
 }
 
 // Set the desired stats for the storage cluster resource
-func (r *ManagedOCSReconciler) setDesiredStorageCluster(
+func (r *ManagedOCSReconciler) generateStorageCluster(
 	reconcileStrategy v1.ReconcileStrategy,
 	sc *ocsv1.StorageCluster) error {
 	r.Log.Info("Reconciling storagecluster", "ReconcileStrategy", reconcileStrategy)
 
 	// Ensure ownership on the storage cluster CR
 	if err := ctrlutil.SetControllerReference(r.managedOCS, sc, r.Scheme); err != nil {
+		r.Log.Error(err, "Failed to set controller reference on storage cluster")
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -83,16 +86,17 @@ func main() {
 		Namespace:          namespace,
 	})
 	if err != nil {
-		setupLog.Error(err, "unable to start manager")
+		setupLog.Error(err, "Unable to start manager")
```
nb-ohad, line 89: Let's leave the message letter casing the way it was. These messages were generated by operator SDK and we might need to generate more in the future. So let's stick to the operator SDK message casing.

This comment should apply all casing changes no just this specific one

main.go
```diff
@@ -83,16 +86,17 @@ func main() {
...
 	}
 
 	if err = (&controllers.ManagedOCSReconciler{
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("controllers").WithName("ManagedOCS"),
-		Scheme: mgr.GetScheme(),
+		Client:               mgr.GetClient(),
+		Log:                  ctrl.Log.WithName("controllers").WithName("ManagedOCS"),
+		Scheme:               mgr.GetScheme(),
+		AddonParamSecretName: getAddonParamSecretName(),
```
nb-ohad, line 97: I don't see a value in a helper for a single call site, especially where the code inside the helper is less than the helper itself. 
Doing it just forces the reader to jump elsewhere in the file to understand the code which is, in my opinion,  cumbersome and annoying. 

```suggestion
		AddonParamSecretName: fmt.Sprintf("addon-%v-parameters", envVars[addonNameEnvVarName])
```

controllers/managedocs_controller_test.go
```diff
@@ -44,14 +45,24 @@ var _ = Describe("ManagedOCS controller", func() {
 	}
 
 	Context("reconcile()", func() {
-		When("there is no storage clsuter resource in the cluster", func() {
-			It("should create a new storage cluster", func() {
+		When("There is no add-on parameters secret in the cluster", func() {
```
nb-ohad, line 48: The string specified by `When` is concatenated into a full sentence in the test output as such it should always start with a lowercase letter.

controllers/managedocs_controller_test.go
```diff
@@ -61,6 +72,86 @@ var _ = Describe("ManagedOCS controller", func() {
 				Expect(k8sClient.Create(ctx, managedOCS)).Should(Succeed())
 				Expect(k8sClient.Get(ctx, getResourceKey(managedOCS), managedOCS)).Should(Succeed())
 
+				// No storage cluster should be created
+				scList = &ocsv1.StorageClusterList{}
+
+				Expect(k8sClient.List(ctx, scList, client.InNamespace(TestNamespace))).Should(Succeed())
+				Expect(scList.Items).Should(HaveLen(0))
```
nb-ohad, line 79: Testing the existence of a storage cluster immediately after the creation of the ManagedOCS is problematic as you don't know if the storage cluster was not created because of the lack of the secret or just because the reconcile loop did not end yet. 

My suggestion is to wait until the ManagedOCS has a status section to insure full reconciliation then validate that there is no storageclsuter.

On the positive cases I am using `waitForReasource` to a accomplish this

controllers/managedocs_controller_test.go
```diff
@@ -61,6 +72,86 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+			})
+		})
+		When("there is incorrect data in the add-on parameters secret", func() {
+			It("should not create a new storage cluster", func() {
+				ctx := context.Background()
+
+				scList := &ocsv1.StorageClusterList{}
+				Expect(k8sClient.List(ctx, scList, client.InNamespace(TestNamespace))).Should(Succeed())
+				Expect(scList.Items).Should(HaveLen(0))
```
nb-ohad, line 89: Why do we need this? the last case just concluded that there is no storage cluster.

controllers/managedocs_controller_test.go
```diff
@@ -61,6 +72,86 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      ManagedOCSName,
+						Namespace: TestNamespace,
+					},
+				}
+				Expect(k8sClient.Get(ctx, getResourceKey(managedOCS), managedOCS)).Should(Succeed())
+
+				defer func() {
+					Expect(k8sClient.Delete(context.Background(), secret)).Should(Succeed())
+				}()
```
nb-ohad, line 115: Cleanup should be the last task on the test because it will trigger another reconcile which might cause an invalid result when verifying the result

controllers/managedocs_controller_test.go
```diff
@@ -61,6 +72,86 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+				// No storage cluster should be created
+				scList = &ocsv1.StorageClusterList{}
+
+				Expect(k8sClient.List(ctx, scList, client.InNamespace(TestNamespace))).Should(Succeed())
+				Expect(scList.Items).Should(HaveLen(0))
```
nb-ohad, line 121: Same as the comment from the last test, we need to verify full reconciliation (even if a failed one)

controllers/managedocs_controller_test.go
```diff
@@ -61,6 +72,86 @@ var _ = Describe("ManagedOCS controller", func() {
...
+			It("should create a new storage cluster", func() {
+				ctx := context.Background()
+
+				scList := &ocsv1.StorageClusterList{}
+				Expect(k8sClient.List(ctx, scList, client.InNamespace(TestNamespace))).Should(Succeed())
+				Expect(scList.Items).Should(HaveLen(0))
+
+				// Create the secret
+				data := make(map[string][]byte)
+				data["size"] = []byte("2")
```
nb-ohad, line 135: Let use a supported value: 1 or 4

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -129,10 +156,47 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
 		// Get an instance of the desired state
 		desired := utils.ObjectFromTemplate(templates.StorageClusterTemplate, r.Scheme).(*ocsv1.StorageCluster)
 
+		if err := r.setStorageClusterParamsFromSecret(desired); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
nbalacha: For a strict reconcile strategy we are setting it back to the desired cluster so I believe this is correct.
If we wish to update the existing that would be a different reconcile strategy.

controllers/managedocs_controller.go
```diff
@@ -114,13 +140,14 @@ func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
 }
 
 // Set the desired stats for the storage cluster resource
-func (r *ManagedOCSReconciler) setDesiredStorageCluster(
+func (r *ManagedOCSReconciler) generateStorageCluster(
```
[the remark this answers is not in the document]
nbalacha: It does actually generate a storagecluster object that will be set so the name seems correct to me.

controllers/managedocs_controller.go
```diff
@@ -129,10 +156,47 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+	osdSize, _ := resource.ParseQuantity(baseOsdSize)
+
+	r.Log.Info("Storage cluster parameters", "count", sdsCount, "osdSize", osdSize)
+	sc.Spec.StorageDeviceSets[0].Count = sdsCount
+	persistentVolume := &sc.Spec.StorageDeviceSets[0].DataPVCTemplate
+	persistentVolume.Spec.Resources = corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			"storage": osdSize,
+		},
+	}
```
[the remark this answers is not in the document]
nbalacha: Leaving it here will make it easier to change it in future in case we want to change it via a parameter.

controllers/managedocs_controller.go
```diff
@@ -129,10 +156,47 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+
+	sdsCount, err := strconv.Atoi(string(data[storageClassSizeKey]))
+	if err != nil {
+		r.Log.Error(err, "Invalid storage cluster size value", "Size", string(data[storageClassSizeKey]))
+		return err
+	}
+	osdSize, _ := resource.ParseQuantity(baseOsdSize)
+
+	r.Log.Info("Storage cluster parameters", "count", sdsCount, "osdSize", osdSize)
+	sc.Spec.StorageDeviceSets[0].Count = sdsCount
```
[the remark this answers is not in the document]
nbalacha: 1.  That is what the replica count in the template takes care of.
2. The last time it was decided that the code would not perform the validations. These constant decision changes are slowing this down.

controllers/managedocs_controller_test.go
nbalacha: A defer statement defers the execution of a function until the surrounding function returns.
This function will be called only at the end of this test function.

controllers/managedocs_controller_test.go
nbalacha: It can be incremented in 1 TiB so this is a valid value.

controllers/managedocs_controller_test.go
nbalacha: While a valid point , I believe this can be done in a separate PR.

controllers/managedocs_controller_test.go
nbalacha: This can be removed but as it does not harm I don't think we need to hold up this PR for this.

controllers/managedocs_controller_test.go
nbalacha: This would already have reconciled in the first test. To actually be sure the managedOCS should be deleted and recreated.
Something I would prefer not to hold up this pr for.

controllers/managedocs_controller.go
```diff
@@ -129,10 +156,47 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+	// size = 1,  creates a cluster of 1 Ti capacity
+	// size = 2,  creates a cluster of 2 Ti capacity etc
+
+	addonParamSecret := &corev1.Secret{}
+	if err := r.Get(r.ctx, types.NamespacedName{Namespace: r.managedOCS.ObjectMeta.Namespace, Name: r.AddonParamSecretName}, addonParamSecret); err != nil {
+		// Do not create the StorageCluster if the we fail to get the addon param secret
+		r.Log.Error(err, "Failed to get the addon param secret.", "SecretName", r.AddonParamSecretName)
+		return err
+	}
+	data := addonParamSecret.Data
```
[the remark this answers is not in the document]
nbalacha: While a valid point, this need not hold up this PR. It can be handled later.

kesavanvt:
ocs-osd-deployer csv needs to be updated with role and addon-name changes, `make bundle` will do the trick

nbalacha:
> ocs-osd-deployer csv needs to be updated with role and addon-name changes, `make bundle` will do the trick

Done.

nb-ohad:
@nbalacha Your code (commits) was merged into master using another PR (https://github.com/openshift/ocs-osd-deployer/pull/20) (after rebasing and fixing the tests concerns). 

Can you please close this one?

PR #17, opened 2021-03-01
Component Status

renan-campos:
Readiness, uninstallation, and general debugging rely on the
    operator to know the current status of its sub-components. This
    commit adds the logic to get those statuses, and a structure within
    the status field to hold that data.


api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,24 @@ type ManagedOCSSpec struct {
...
+
+const (
+	ComponentReady   ComponentStatus = "Ready"
+	ComponentPending ComponentStatus = "Pending"
+	ComponentNone    ComponentStatus = "None"
+)
+
+type ManagedOCSComponent struct {
+	Name   string          `json:"name"`
+	Status ComponentStatus `json:"status"`
```
nb-ohad, line 51: Aren't we missing the timestamp field?

.gitignore
```diff
@@ -22,3 +22,5 @@ bin
 *.swp
 *.swo
 *~
+
+ocs-osd-deployer
```
nb-ohad, line 26: Is this change by mistake?

api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,24 @@ type ManagedOCSSpec struct {
...
+}
+
 // ManagedOCSStatus defines the observed state of ManagedOCS
 type ManagedOCSStatus struct {
-	ReconcileStrategy ReconcileStrategy `json:"reconcileStrategy,omitempty"`
+	ReconcileStrategy ReconcileStrategy     `json:"reconcileStrategy,omitempty"`
+	Components        []ManagedOCSComponent `json:"components"`
+	ReconcileCount    uint                  `json:"ReconcileCount,omitempty"`
```
nb-ohad, line 58: Not needed as "generation" is already managed by k8s

controllers/managedocs_controller.go
```diff
@@ -77,6 +77,7 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	err := r.reconcilePhases(req)
 
 	// Ensure status is updated once even on failed reconciles
+	r.managedOCS.Status.ReconcileCount++
```
nb-ohad, line 80: Not needed

controllers/managedocs_controller.go
```diff
@@ -136,3 +145,34 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+	component := v1.ManagedOCSComponent{
+		Name: "StorageCluster",
+	}
+
+	var scList ocsv1.StorageClusterList
+
+	if err := r.List(r.ctx, &scList); err != nil {
+		r.Log.Error(err, "error listing StorageCluster")
+		return nil, err
+	}
```
nb-ohad, line 162: I believe `Get` is better in this case than `list`

controllers/managedocs_controller.go
```diff
@@ -136,3 +145,34 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+
+	if len(scList.Items) < 1 {
+		component.Status = v1.ComponentNone
```
nb-ohad, line 165: Why do we need a status as all in this case, can't we just set the proper component status to nil?

controllers/managedocs_controller.go
```diff
@@ -136,3 +145,34 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+	} else {
+		// Assuming only one StorageCluster in namespace.
+		if scList.Items[0].Status.Phase == "Ready" {
```
nb-ohad, line 168: It is better to search for the item in a list by name not by location. future changes to ocs might result in lists with more than 1 item

nb-ohad (changes requested):

api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,24 @@ type ManagedOCSSpec struct {
 	ReconcileStrategy ReconcileStrategy `json:"reconcileStrategy,omitempty"`
 }
 
+type ComponentStatus string
+
+const (
+	ComponentReady   ComponentStatus = "Ready"
+	ComponentPending ComponentStatus = "Pending"
+	ComponentNone    ComponentStatus = "None"
```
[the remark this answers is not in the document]
renan-campos: None means the component isn't there. For example if the StorageCluster isn't present.

.gitignore
renan-campos: no, ocs-osd-deployer is the executable build by calling "go build", but we'd never commit the binary so I wanted to gitignore it.

api/v1alpha1/managedocs_types.go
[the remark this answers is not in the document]
nb-ohad: So let's just not have it in the status instead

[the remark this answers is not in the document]
nb-ohad: Another thought, why do we stick to these values and not just reflect the value of the readiness of the components?
For example, in the case of OCS, we can just reflect the phase field.

api/v1alpha1/managedocs_types.go
renan-campos: generation doesn't keep track of the same thing that ReconcileCount is tracking.

.gitignore
nb-ohad: Ack, was thinking that this was something else.

api/v1alpha1/managedocs_types.go
nb-ohad: Ok, so what does it track and why do we need it? 
The CR is a public API (not an implementation detail) and as such every piece of it should have a reasoning

controllers/managedocs_controller.go
renan-campos: No a Get would fail if the storage cluster isn't present.

controllers/managedocs_controller.go
renan-campos: I'm using it for testing and debugging purposes. Many of the tests require a reconciliation loop to occur before checking if reconcile did what we expected. Plus, I'm interested to know how many times the CR is being reconciled.

nb-ohad: Please see my comment about public API above. You can store something on the reconciler or the manager if needed.

renan-campos: Keeping track of the times the cr was reconciled provides an immediate way for the tests to know when they can check what they need to.

controllers/managedocs_controller.go
nb-ohad: Yes, it will fail with a proper return value. 
But if you think it is better to use list instead, I will not insist

controllers/managedocs_controller.go
nb-ohad: The last update time is already on the CR and is part of the public API. Please use it instead

api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,24 @@ type ManagedOCSSpec struct {
...
+type ManagedOCSComponent struct {
+	Name   string          `json:"name"`
+	Status ComponentStatus `json:"status"`
+}
+
 // ManagedOCSStatus defines the observed state of ManagedOCS
 type ManagedOCSStatus struct {
-	ReconcileStrategy ReconcileStrategy `json:"reconcileStrategy,omitempty"`
+	ReconcileStrategy ReconcileStrategy     `json:"reconcileStrategy,omitempty"`
+	Components        []ManagedOCSComponent `json:"components"`
```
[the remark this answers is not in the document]
renan-campos: Ok, I will change to a non-array structure for this.

api/v1alpha1/managedocs_types.go
renan-campos: "The .metadata.generation value is incremented for all changes, except for changes to .metadata or .status.". I'm tracking all times reconcile was run on this cr.

nb-ohad: I understand, but this should not be part of the public API.
You have these options:
1. ` generation`
2. `resourceVersion`
3. The last update timestamp

controllers/managedocs_controller.go
renan-campos: Where?

controllers/managedocs_controller.go
renan-campos: Where could I find information on the errors that Get returns? If there is a NotFound error type that it returns than I agree that Get would be cleaner, but I can find no documentation on the error these calls make.

controllers/managedocs_controller.go
nb-ohad: The create/update count is captured by `resourceVersion` and the timestamp should be captured somewhere in the metadata (I beleive under ManagedFields). But if not we can add it to the spec - this will be reasonable for a public API

controllers/managedocs_controller.go
nb-ohad: You need to do `if !errors.IsNotFound(err)` on the returned `err` of the `get` operation to differentiate. 
You can look at the implementation of `CreateOrUpdate` as an example.

api/v1alpha1/managedocs_types.go
renan-campos: Why do we need timestamp?

nb-ohad: Because the status of the component will always be a stale status. As such we need to know the time when the observation was true

api/v1alpha1/managedocs_types.go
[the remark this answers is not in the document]
renan-campos: My thought is that this way provides a consistent method of checking the statuses of the components. To check if any component is ready: `component.Status == ComponentReady`.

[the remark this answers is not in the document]
renan-campos: So outside of the code that aggregates the statuses of each components, these are the only statuses that we care about.

controllers/managedocs_controller.go
renan-campos: For debugging purposes I would prefer to see that the reconcile looked for the StorageCluster and found none at the time.

nb-ohad: I see 2 options:
1. Remove it from the status and add it to the log, the reason is that debug purposes should not inflect public API
2. Make it meaningful by changing the value from none to "Uninstalled" which has a valid meaning

api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,27 @@ type ManagedOCSSpec struct {
...
+)
+
+type ManagedOCSComponent struct {
+	Status      ComponentStatus `json:"status"`
+	LastChecked metav1.Time     `json:"lastChecked"`
+}
+
+type ManagedOCSComponents struct {
+	StorageCluster ManagedOCSComponent `json:"storageCluster"`
+}
```
nb-ohad, line 56: I will suggest renaming these structs:
-  The `ManagedOCS` prefix is unneeded as all these structures live under the ManagedOCS API
- `Components` and `Components` are easily mixed while reading (happened to me) 
- The structures' name does not covey the fact these are status structures.

I will suggest something in the line of `ComponentStatusMap` and `ComponentStatus`

bundle.Dockerfile
```diff
@@ -5,7 +5,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=ocs-osd-deployer
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.3.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.2.0
```
nb-ohad, line 8: Please revert

bundle/manifests/ocs-osd-deployer.clusterserviceversion.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
         }
       ]
     capabilities: Basic Install
-    operators.operatorframework.io/builder: operator-sdk-v1.3.0
+    operators.operatorframework.io/builder: operator-sdk-v1.2.0
```
nb-ohad, line 16: Please revert

api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,27 @@ type ManagedOCSSpec struct {
...
+
+const (
+	ComponentReady   ComponentStatus = "Ready"
+	ComponentPending ComponentStatus = "Pending"
+	ComponentNone    ComponentStatus = "None"
+)
+
+type ManagedOCSComponent struct {
+	Status      ComponentStatus `json:"status"`
+	LastChecked metav1.Time     `json:"lastChecked"`
```
nb-ohad, line 51: `LastCheck` is a misleading name for the field because for some resources we can be much more accurate by reading a last update time field on the resource and use it here instead of taking the last check date.

My suggestion is to rename to `LastUpdate` or `LastUpdateTime`

bundle/metadata/annotations.yaml
```diff
@@ -4,7 +4,7 @@ annotations:
   operators.operatorframework.io.bundle.mediatype.v1: registry+v1
   operators.operatorframework.io.bundle.metadata.v1: metadata/
   operators.operatorframework.io.bundle.package.v1: ocs-osd-deployer
-  operators.operatorframework.io.metrics.builder: operator-sdk-v1.3.0
+  operators.operatorframework.io.metrics.builder: operator-sdk-v1.2.0
```
nb-ohad, line 7: Please revert

controllers/managedocs_controller.go
```diff
@@ -146,3 +156,32 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
 	return nil
 }
+
+func (r *ManagedOCSReconciler) updateComponents(req ctrl.Request) (v1.ManagedOCSComponents, error) {
+	var components v1.ManagedOCSComponents
+
+	// Checking the StorageCluster component.
+	var sc ocsv1.StorageCluster
+
+	scNamespaceName := types.NamespacedName{
```
nb-ohad, line 166: `scNamespaceName` -> `scNamespacedName`

api/v1alpha1/managedocs_types.go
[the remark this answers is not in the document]
nb-ohad: Ok for that, But please still change from "None" to "Unknown" and "Uninstalled" in the proper cases.

controllers/managedocs_controller.go
```diff
@@ -146,3 +156,32 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+		} else {
+			components.StorageCluster.Status = v1.ComponentPending
+		}
+	} else if errors.IsNotFound(err) {
+		components.StorageCluster.Status = v1.ComponentNone
+	} else {
+		r.Log.Error(err, "error getting StorageCluster")
+		return components, err
+	}
+	components.StorageCluster.LastChecked = metav1.Now()
```
nb-ohad, line 184: `metav1.Now()` should only be used if we cannot infere the last update time from the actual resource. 
Please check if we can for `stroagecluster` before falling back to `Now()`

controllers/managedocs_controller.go
```diff
@@ -146,3 +156,32 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+		if sc.Status.Phase == "Ready" {
+			components.StorageCluster.Status = v1.ComponentReady
+		} else {
+			components.StorageCluster.Status = v1.ComponentPending
+		}
+	} else if errors.IsNotFound(err) {
+		components.StorageCluster.Status = v1.ComponentNone
+	} else {
+		r.Log.Error(err, "error getting StorageCluster")
+		return components, err
```
nb-ohad, line 182: Why do we need to fail the reconcile in this case. 
Why not just specify the status as 'Unknown' and log the error?

nb-ohad (changes requested):
Does not contain review on test cases

controllers/managedocs_controller.go
renan-campos: How is this field going to be used?

renan-campos: I'm currently using this field to track the last time we checked on this component.

renan-campos: what is the value in infering the last update time of the actual resource, assuming it even reports it.

nb-ohad: We should give the most accurate value we can give, the check time should be a fall back only if we cannot get a more accurate time somehow

nb-ohad: The value is in aggregating and reporting the most accurate time to the entity which consumes this information.  
Think about a case where the storagecluster is not being reconciled by the OCS operator at all. The time should be fixed but If you take the sampling time it will change with each reconcile. 

Now In some cases, we do not have a source to read an actual update time from, in that case, the sample time is the best approximation we have.

renan-campos: storagecluster has no time field.

nb-ohad: Ok, what about `ManagedFields` in the metadata section?
they are not mandatory, but if exists we can use it

renan-campos: there is nothing to indicate that those times are updated by the storagecluster being updated.

nb-ohad: Look at the definition for ManagedField

renan-campos: 272   // ManagedFields maps workflow-id and version to the set of fields                                                                         
 273   // that are managed by that workflow. This is mostly for internal                                                                          
 274   // housekeeping, and users typically shouldn't need to set or                                                                              
 275   // understand this field. A workflow can be the user's name, a                                                                             
 276   // controller's name, or the name of a specific apply path like                                                                            
 277   // "ci-cd". The set of fields is always in the version that the                                                                            
 278   // workflow used when modifying the object.                                                                                                
 279   //                                                                                                                                         
 280   // +optional                                                                                                                               
 281   ManagedFields []ManagedFieldsEntry `json:"managedFields,omitempty" protobuf:"bytes,17,rep,name=managedFields"`                                           

1148   // Time is timestamp of when these fields were set. It should always be empty if Operation is 'Apply'                                      
1149   // +optional                                                                                                                               
1150   Time *Time `json:"time,omitempty" protobuf:"bytes,4,opt,name=time"`

renan-campos: ?

nb-ohad: Ok, so in this case, lets go with the sample time and we can request ocs-operator in the future to provide a proper time.

nb-ohad: I gave another thought about this matter. If the only component we have now is StorageCluster and we have no means to get the update time then you can remove it from the struct, but keep the struct as an extension point.

In the future when we will have more components we can reevaluate

api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,27 @@ type ManagedOCSSpec struct {
...
+
+const (
+	ComponentReady   ComponentStatusMsg = "Ready"
+	ComponentPending ComponentStatusMsg = "Pending"
+	ComponentNone    ComponentStatusMsg = "None"
+	ComponentUnknown ComponentStatusMsg = "Unknown"
+)
+
+type ComponentStatus struct {
+	Status ComponentStatusMsg `json:"status"`
```
nb-ohad, line 51: I believe a more accurate name to this field would be `State` see, 
https://softwareengineering.stackexchange.com/questions/219351/state-or-status-when-should-a-variable-name-contain-the-word-state-and-w#:~:text=status%20is%20used%20to%20describe,(e.g.%20pending%2Fdispatched).

controllers/managedocs_controller.go
```diff
@@ -146,3 +156,31 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
 
 	return nil
 }
+
+func (r *ManagedOCSReconciler) updateComponents(req ctrl.Request) (v1.ComponentStatusMap, error) {
```
nb-ohad, line 160: This function does not return any errors.

controllers/managedocs_controller.go
```diff
@@ -146,3 +156,31 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+	if err == nil {
+		if sc.Status.Phase == "Ready" {
+			components.StorageCluster.Status = v1.ComponentReady
+		} else {
+			components.StorageCluster.Status = v1.ComponentPending
+		}
+	} else if errors.IsNotFound(err) {
+		components.StorageCluster.Status = v1.ComponentNone
+	} else {
+		r.Log.Error(err, "error getting StorageCluster")
```
nb-ohad, line 181:
```suggestion
		r.Log.Error(err, "error getting StorageCluster, setting compoment status to Unknown")
```

controllers/managedocs_controller.go
```diff
@@ -146,3 +156,31 @@ func (r *ManagedOCSReconciler) setDesiredStorageCluster(
...
+
+	err := r.Get(r.ctx, scNamespacedName, &sc)
+	if err == nil {
+		if sc.Status.Phase == "Ready" {
+			components.StorageCluster.Status = v1.ComponentReady
+		} else {
+			components.StorageCluster.Status = v1.ComponentPending
+		}
+	} else if errors.IsNotFound(err) {
+		components.StorageCluster.Status = v1.ComponentNone
```
nb-ohad, line 179: Let's be more specific, can we set the state to `NotFound` as everything else is treated as `Unknown`?

controllers/managedocs_controller_test.go
```diff
@@ -71,8 +71,102 @@ var _ = Describe("ManagedOCS controller", func() {
...
+		When("the storeage cluster is not ready", func() {
+			ctx := context.Background()
+			managedOCS := &v1.ManagedOCS{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      ManagedOCSName,
+					Namespace: TestNamespace,
+				},
+			}
+
+			BeforeEach(func() {
```
nb-ohad, line 83: Why do we need this before each when there is only one test it affects? 
It just creates more convoluted code and a discontinuous stack trace  
Is there a specific reason?

controllers/managedocs_controller_test.go
```diff
@@ -71,8 +71,102 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				// This test, like the ones below it, assume managed-ocs is already created.
+				Expect(k8sClient.Get(ctx, getResourceKey(managedOCS), managedOCS)).Should(Succeed())
+
+				// Using the last time the StorageCluster was checked as a way to tell if the reconcile loop was run.
+				lastCheck := managedOCS.Status.Components.StorageCluster.LastChecked
```
nb-ohad: `LastChecked` is not a field anymore

controllers/managedocs_controller_test.go
```diff
@@ -71,8 +71,102 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				}
+				Expect(k8sClient.Get(ctx, getResourceKey(&sc), &sc)).Should(Succeed())
+
+				// Updating the Status of the StorageCluster should trigger a reconcile
+				// for managed-ocs
+				sc.Status.Phase = "Pending"
+				Expect(k8sClient.Status().Update(ctx, &sc)).Should(Succeed())
+
+				// Wait for operator to reconcile before checking conditions
+				Eventually(func() bool {
```
nb-ohad: Can't you use one of the utils (maybe enhance one of them) to do this check?

controllers/managedocs_controller_test.go
```diff
@@ -71,8 +71,102 @@ var _ = Describe("ManagedOCS controller", func() {
...
+		When("the storeage cluster is ready", func() {
+			ctx := context.Background()
+			managedOCS := &v1.ManagedOCS{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      ManagedOCSName,
+					Namespace: TestNamespace,
+				},
+			}
+
+			BeforeEach(func() {
```
nb-ohad, line 120: Same as a above

nb-ohad (changes requested):

api/v1alpha1/managedocs_types.go
renan-campos: "a choice needs to be made involving the perverted intricacies of English language"- what an epithet! ok `State` it is

controllers/managedocs_controller_test.go
renan-campos: Yes I wrote it this way to separate the code that sets up the state of the test from the code that tests the behaviour when in that state. Now every `It` within that `When` block can assume the cluster is in a state where:
1. Managed ocs has been created
2. StorageCluster has been created
3. StorageCluster is in a pending Phase

controllers/managedocs_controller.go
```diff
@@ -223,3 +227,31 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
...
+
+func (r *ManagedOCSReconciler) updateComponents(req ctrl.Request) v1.ComponentStatusMap {
+	var components v1.ComponentStatusMap
+
+	// Checking the StorageCluster component.
+	var sc ocsv1.StorageCluster
+
+	scNamespacedName := types.NamespacedName{
+		Name:      storageClusterName,
+		Namespace: req.Namespace,
```
nb-ohad, line 239: Please use `r.namespace instead`. I intentionally remove the `req` param from `reconcilePhases`

controllers/managedocs_controller.go
```diff
@@ -131,7 +132,10 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	}
 }
 
-func (r *ManagedOCSReconciler) reconcilePhases() error {
+func (r *ManagedOCSReconciler) reconcilePhases(req ctrl.Request) error {
```
nb-ohad, line 135: The `req` param was removed intentionally in the last commit. Your only use for it is to get access to the namespace, for that you can now use `r.namespace` instead

api/v1alpha1/managedocs_types.go
```diff
@@ -38,9 +38,27 @@ type ManagedOCSSpec struct {
 	ReconcileStrategy ReconcileStrategy `json:"reconcileStrategy,omitempty"`
 }
 
+type ComponentState string
+
+const (
+	ComponentReady    ComponentState = "Ready"
+	ComponentPending  ComponentState = "Pending"
+	ComponentNotFound ComponentState = "Not Found"
```
nb-ohad, line 46: I think somehow the following change got reverted:

```suggestion
	ComponentNotFound ComponentState = "NotFound"
```

nb-ohad (changes requested):

nb-ohad:
@JohnStrunk Can you please merge this one?

PR #18, opened 2021-03-02
using defer to guarantee that status is updated at the end of reconcile.

renan-campos:
We would like to ensure the reconcile loop always updates the CR's
status. As it was written prior, this required the developer to remember
not to modify or return before running that logic in the reconcile
loop. With the use of defer, this requirement is now explicitly
expressed in the code. As reconcile returns, the function defined within
the defer statement is called and will update the status. This will run
no matter when reconcile returns.

### DIFF
--- a/bundle/manifests/ocs-osd-deployer.clusterserviceversion.yaml
+++ b/bundle/manifests/ocs-osd-deployer.clusterserviceversion.yaml
@@ -9,9 +9,6 @@ metadata:
           "kind": "ManagedOCS",
           "metadata": {
             "name": "managedocs-sample"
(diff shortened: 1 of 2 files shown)


nb-ohad:
This is not the classic use for `defer`. 
The common use case for `defer` is release logic for resources, to specify the release logic immediately after the acquiring code, which makes the code more readable. 

In my opinion, this change just makes the code less readable and more convoluted. 
Also, the reader now needs to track the return value of the function in 2 diff places. 

@renan-campos 
I am against merging this one unless a compelling argument can be presented.

PR #21, opened 2021-03-05
Adding readiness probe

renan-campos:
As an OCM add-on, the ocs-osd-operator needs a way to communicate its
status to OCM. OCM is currently limited to using the CSV status of the
add-on, which it gets every 8 minutes through OLM telemetry. In order to
control when the operator is reported as installed, a readiness probe
was added, along with logic in the controller to only report the
operator as ready when all of its components are in their desired state.

### DIFF
--- a/readinessProbe/main.go
+++ b/readinessProbe/main.go
@@ -1,0 +1,71 @@
+// Description: This program creates a web server to verify if the managedOCS
+//              resource is ready. It is used as a readiness probe by the
+//              ocs-osd-deployer operator. For this to be set up, the following
+//              sidecar should be added to the manager CSV:
+//      - name: readinessServer
+//        command:
+//        - /readinessServer
+//        image: ocs-osd-deployer:latest
+//        readinessProbe:
+//          httpGet:
+//            path: /readyz
+//            port: 8081
+//          initialDelaySeconds: 5
+//          periodSeconds: 10
+package main
+
+import (
+	"fmt"
+	"os"
+
+	ocsv1 "github.com/openshift/ocs-operator/pkg/apis"
+	v1 "github.com/openshift/ocs-osd-deployer/api/v1alpha1"
+	"github.com/openshift/ocs-osd-deployer/readinessProbe/readiness"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+)
+
+func main() {
+	// Setup logging
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	log := ctrl.Log.WithName("readiness")
+
+	// Setup client options
+	var options client.Options
+
+	// The readiness must have these schemes to deserialize the k8s objects
+	options.Scheme = runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(options.Scheme))
+	utilruntime.Must(ocsv1.AddToScheme(options.Scheme))
+	utilruntime.Must(v1.AddToScheme(options.Scheme))
+
+	k8sClient, err := client.New(config.GetConfigOrDie(), options)
+	if err != nil {
+		log.Error(err, "error creating client")
+		os.Exit(1)
+	}
+
+	namespace, found := os.LookupEnv(readiness.NamespaceEnvVarName)
+	if !found {
+		log.Error(fmt.Errorf("%q not set", readiness.NamespaceEnvVarName), "error in environment variables")
+		os.Exit(2)
+	}
+
+	managedOCSResource := types.NamespacedName{
+		Name:      "managedocs",
+		Namespace: namespace,
+	}
+
+	log.Info("starting HTTP server...")
+	err = readiness.RunServer(k8sClient, managedOCSResource, log)
+	if err != nil {
+		log.Error(err, "server error")
+	}
+	log.Info("HTTP server terminated.")
+}
--- a/readinessProbe/readiness/server.go
+++ b/readinessProbe/readiness/server.go
@@ -1,0 +1,56 @@
+package readiness
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/go-logr/logr"
+	v1 "github.com/openshift/ocs-osd-deployer/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	listenAddr          string = ":8081"
+	readinessPath       string = "/readyz/"
+	NamespaceEnvVarName string = "NAMESPACE"
+)
+
+func isReady(client client.Client, managedOCSResource types.NamespacedName) (bool, error) {
+
+	var managedOCS v1.ManagedOCS
+
+	if err := client.Get(context.Background(), managedOCSResource, &managedOCS); err != nil {
+		return false, err
+	}
+
+	ready := managedOCS.Status.Components.StorageCluster.State == v1.ComponentReady
+	return ready, nil
+}
+
+func RunServer(client client.Client, managedOCSResource types.NamespacedName, log logr.Logger) error {
+
+	// Readiness probe is defined here.
+	// From k8s documentation:
+	// "Any code greater than or equal to 200 and less than 400 indicates success."
+	// [indicates that the deployment is ready]
+	// "Any other code indicates failure."
+	// [indicates that the deployment is not ready]
+	http.HandleFunc(readinessPath, func(httpw http.ResponseWriter, req *http.Request) {
+		ready, err := isReady(client, managedOCSResource)
+
+		if err != nil {
+			log.Error(err, "error checking readiness\n")
+			httpw.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		if ready {
+			httpw.WriteHeader(http.StatusOK)
+		} else {
+			httpw.WriteHeader(http.StatusServiceUnavailable)
+		}
+	})
+
+	return http.ListenAndServe(listenAddr, nil)
+}
--- a/readinessProbe/readiness/server_test.go
+++ b/readinessProbe/readiness/server_test.go
@@ -1,0 +1,78 @@
+package readiness
+
+import (
+	"context"
+	"net/http"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	v1 "github.com/openshift/ocs-osd-deployer/api/v1alpha1"
+	utils "github.com/openshift/ocs-osd-deployer/testutils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("ManagedOCS Readiness Probe", func() {
+	// Define utility constants for object names and testing timeouts/durations and intervals.
+	const (
+		timeout  = time.Second * 10
+		duration = time.Second * 10
+		interval = time.Millisecond * 250
+	)
+
+	When("managedocs reports its components as \"not ready\"", func() {
+		BeforeEach(func() {
+			ctx := context.Background()
+
+			managedOCS := &v1.ManagedOCS{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      ManagedOCSName,
+					Namespace: TestNamespace,
+				},
+				Status: v1.ManagedOCSStatus{
+					Components: v1.ComponentStatusMap{
+						StorageCluster: v1.ComponentStatus{
+							State: v1.ComponentPending,
+						},
+					},
+				},
+			}
+
+			Expect(k8sClient.Create(ctx, managedOCS)).Should(Succeed())
+
+			utils.WaitForResource(k8sClient, ctx, managedOCS, timeout, interval)
+		})
+
+		It("should cause the readiness logic to report \"not ready\"", func() {
+			status, err := utils.ProbeReadiness()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(status).To(Equal(http.StatusServiceUnavailable))
+		})
+	})
+
+	When("managedocs reports its components as \"ready\"", func() {
+		BeforeEach(func() {
+			ctx := context.Background()
+			managedOCS := &v1.ManagedOCS{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      ManagedOCSName,
+					Namespace: TestNamespace,
+				},
+			}
+
+			// This test expects the ManagedOCS resource to already have been made.
+			err := k8sClient.Get(ctx, utils.GetResourceKey(managedOCS), managedOCS)
+			Expect(err).ToNot(HaveOccurred())
+
+			managedOCS.Status.Components.StorageCluster.State = v1.ComponentReady
+
+			Expect(k8sClient.Status().Update(ctx, managedOCS)).Should(Succeed())
+		})
+
+		It("should cause the readiness logic to report \"ready\"", func() {
+			status, err := utils.ProbeReadiness()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(status).To(Equal(http.StatusOK))
+		})
+	})
+})
--- a/readinessProbe/readiness/suite_test.go
+++ b/readinessProbe/readiness/suite_test.go
@@ -1,0 +1,100 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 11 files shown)


controllers/managedocs_controller.go
```diff
@@ -136,6 +136,13 @@ func (r *ManagedOCSReconciler) reconcilePhases() error {
 	// Update the status of the components
 	r.managedOCS.Status.Components = r.updateComponents()
 
+	// Update the readiness of the operator.
+	if r.managedOCS.Status.Components.StorageCluster.State != v1.ComponentReady {
+		utils.SetNotReady()
+	} else {
+		utils.SetReady()
+	}
+
```
nb-ohad, line 145: I would prefer the controller itself will have no relation to the prob, so that future PR could easily move it into its own binary.  

We have reflected the components state on the `ManagedOCS` status in order to provide the prob code a way to query the status without any need for in-memory communication between the controller and the prob code

utils/readiness.go
```diff
@@ -0,0 +1,48 @@
...
+	readinessServer.Lock()
+	readinessServer.ready = true
+	readinessServer.Unlock()
+}
+
+func SetNotReady() {
+	readinessServer.Lock()
+	readinessServer.ready = false
+	readinessServer.Unlock()
+}
```
nb-ohad, line 48: @renan-campos as we agreed when we did the code design discussion, there should be no dependency on shared structures between the controller and the prob. 

All of the above code is unnecessary

main.go
```diff
@@ -60,8 +61,10 @@ func init() {
 
 func main() {
 	var metricsAddr string
+	var probeAddr string
 	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "ready-probe-bind-address", ":8081", "The address the readiness probe endpoint binds to.")
```
nb-ohad, line 67: What does this row suppose to do? Do we really need it?

nb-ohad (changes requested):
Reviewed with test code

controllers/managedocs_controller.go
renan-campos: Addressed in new commit.

utils/readiness.go
renan-campos: Addressed in new commit.

renan-campos:
Tested on OCP on AWS using OLM, behaves as expected!

vendor/modules.txt
```diff
@@ -259,6 +259,7 @@ gopkg.in/tomb.v1
 # gopkg.in/yaml.v2 v2.3.0
 gopkg.in/yaml.v2
 # k8s.io/api v0.19.3 => k8s.io/api v0.19.3
+## explicit
```
nb-ohad, line 262: This might have implications. was this auto-generated or added manually?

vendor/github.com/onsi/gomega/go.mod
```diff
@@ -1,7 +1,5 @@
 module github.com/onsi/gomega
 
-go 1.15
-
```
nb-ohad, line 4: Is there a reason to remove the go version? it caused up problems in the past

readinessProbe/readiness/server.go
```diff
@@ -0,0 +1,59 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	listenAddr          string = ":8081"
+	readinessPath       string = "/readyz/"
+	NamespaceEnvVarName string = "NAMESPACE"
+)
+
+func Logic(client client.Client, managedOCSResource types.NamespacedName, w io.Writer) (bool, error) {
```
nb-ohad, line 19: [Suggestion] Have a better descriptive name, also make is private

```suggestion
func IsReady(client client.Client, managedOCSResource types.NamespacedName, w io.Writer) (bool, error) {
```

readinessProbe/readiness/server.go
```diff
@@ -0,0 +1,59 @@
...
+	// From k8s documentation:
+	// "Any code greater than or equal to 200 and less than 400 indicates success."
+	// [indicates that the deployment is ready]
+	// "Any other code indicates failure."
+	// [indicates that the deployment is not ready]
+	http.HandleFunc(readinessPath, func(httpw http.ResponseWriter, req *http.Request) {
+		ready, err := Logic(client, managedOCSResource, w)
+
+		if err != nil {
+			w.Write([]byte("error in readiness logic\n"))
```
nb-ohad, line 47: You are already logging the error in the logic function then again here. This just creates an over cluttered log. 

Instead, you can drop the log from `Logic` and just return it as the actual error using `fmt.Errorf()` then you can log here providing the returned error as part of the message

readinessProbe/readiness/server.go
```diff
@@ -0,0 +1,59 @@
...
+			return
+		}
+
+		if ready {
+			httpw.WriteHeader(200)
+		} else {
+			httpw.WriteHeader(418)
```
nb-ohad, line 54: You are returning "I'm a teapot" response (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418).

Unless I am missing something the readiness server is serving even in the case the pod is not ready which means that both `ready` and `non-ready` should return HTTP 200 and the result should be part of the body.

The only case, I can identify,  where another  HTTP code is needed is in the err path

readinessProbe/readiness/server.go
```diff
@@ -0,0 +1,59 @@
...
+	// "Any code greater than or equal to 200 and less than 400 indicates success."
+	// [indicates that the deployment is ready]
+	// "Any other code indicates failure."
+	// [indicates that the deployment is not ready]
+	http.HandleFunc(readinessPath, func(httpw http.ResponseWriter, req *http.Request) {
+		ready, err := Logic(client, managedOCSResource, w)
+
+		if err != nil {
+			w.Write([]byte("error in readiness logic\n"))
+			return
```
nb-ohad, line 48: Missing an HTTP response for the err case. 
I think a good match for this case is HTTP 500 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500

readinessProbe/main.go
```diff
@@ -0,0 +1,65 @@
+// Description: This program creates a web server to verify if the managedOCS
+//              resource is ready. It is used as a readiness probe by the
+//              ocs-osd-deployer operator. For this to be set up, the following
+//              sidecar should be added to the manager CSV:
+//      - name: readinessServer
+//        command:
+//        - /readinessServer
+//        image: quay.io/rcampos/ocs-osd-deployer-readiness:latest
```
nb-ohad, line 8:
```suggestion
//        image: ocs-osd-deployer:latest
```

readinessProbe/main.go
```diff
@@ -0,0 +1,65 @@
...
+	// Setup client options
+	var options client.Options
+
+	// The readiness must have these schemes to deserialize the k8s objects
+	options.Scheme = runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(options.Scheme))
+	utilruntime.Must(ocsv1.AddToScheme(options.Scheme))
+	utilruntime.Must(v1.AddToScheme(options.Scheme))
+
+	k8sClient, err := client.New(config.GetConfigOrDie(), options)
```
nb-ohad, line 43: Are we ok with a non-cached client? I think we should use a cached client for performance reasons

readinessProbe/main.go
```diff
@@ -0,0 +1,65 @@
...
+	v1 "github.com/openshift/ocs-osd-deployer/api/v1alpha1"
+	"github.com/openshift/ocs-osd-deployer/readinessProbe/readiness"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
```
nb-ohad, line 32: I prefer if we use the controller-runtime log instead of golang log or writing directly to stdout

```suggestion
var (
	log = ctrl.Log.WithName("readiness")
)
```

readinessProbe/main.go
```diff
@@ -0,0 +1,65 @@
...
+	// The readiness must have these schemes to deserialize the k8s objects
+	options.Scheme = runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(options.Scheme))
+	utilruntime.Must(ocsv1.AddToScheme(options.Scheme))
+	utilruntime.Must(v1.AddToScheme(options.Scheme))
+
+	k8sClient, err := client.New(config.GetConfigOrDie(), options)
+	if err != nil {
+		log.Fatal(err)
```
nb-ohad, line 45: I think it will be beneficial for us to have  please  surrounding message on top of the error to make it easier to track in the logs

readinessProbe/main.go
```diff
@@ -0,0 +1,65 @@
...
+		os.Exit(2)
+	}
+
+	managedOCSResource := types.NamespacedName{
+		Name:      "managedocs",
+		Namespace: namespace,
+	}
+
+	fmt.Println("Running Readiness Server...")
+	err = readiness.RunServer(k8sClient, managedOCSResource, os.Stdout)
```
nb-ohad, line 61: I highly discourage writing directly to stdout. Can we pass in `log` instead?

readinessProbe/main.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		log.Fatal(err)
```
nb-ohad, line 63: I think it will be beneficial for us to have please surrounding message on top of the error to make it easier to track in the logs

controllers/suite_test.go
```diff
@@ -78,7 +78,8 @@ var _ = BeforeSuite(func(done Done) {
 	// +kubebuilder:scaffold:scheme
 
 	k8sManager, err := ctrl.NewManager(cfg, ctrl.Options{
-		Scheme: scheme.Scheme,
+		Scheme:                 scheme.Scheme,
+		HealthProbeBindAddress: ":8081",
```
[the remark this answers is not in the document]
renan-campos: No sorry I missed that one!

vendor/modules.txt
renan-campos: This was auto generated, I'm not sure what that means actually

readinessProbe/main.go
renan-campos: I used the non-cached client to ensure we get the freshest cr status, and I don't think there will be any significant  performance gain from using the cached client since this code only does one get operation every 10 seconds.

vendor/modules.txt
renan-campos: If I try to manually change it, I get the following error: `k8s.io/api@v0.19.3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt`

readinessProbe/main.go
nb-ohad: Even with a cached client, you will get an updated CRs as the cache is push/invalidated upon changes.
But I get your point. You can resolve this thread

readinessProbe/readiness/server_test.go
```diff
@@ -0,0 +1,98 @@
...
+					Namespace: TestNamespace,
+				},
+			}
+
+			Expect(k8sClient.Create(ctx, managedOCS)).Should(Succeed())
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, getResourceKey(managedOCS), managedOCS)
+				return err == nil
+			}, timeout, interval).Should(BeTrue())
```
nb-ohad, line 49: This code is bassicly `waitForResource` from `managedocs_controller_test.go`.
As of the last comment, I suggest moving this shared functionality to  `test/util.go`

readinessProbe/readiness/server_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			managedOCS.Status.Components.StorageCluster.State = v1.ComponentPending
+
+			Expect(k8sClient.Status().Update(ctx, managedOCS)).Should(Succeed())
```
nb-ohad, line 53: Can't we just specify the status as part of the managedOCS creation at line 44 and remove the extra update?

readinessProbe/readiness/server_test.go
```diff
@@ -0,0 +1,98 @@
...
+					Name:      ManagedOCSName,
+					Namespace: TestNamespace,
+				},
+			}
+
+			// This test expects the ManagedOCS resource to already have been made.
+			err := k8sClient.Get(ctx, getResourceKey(managedOCS), managedOCS)
+			Expect(err).ToNot(HaveOccurred())
+
+			managedOCS.Status.Components.StorageCluster.State = v1.ComponentReady
```
nb-ohad, line 82: Same as the comment on the last test

readinessProbe/readiness/server_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			Expect(k8sClient.Status().Update(ctx, managedOCS)).Should(Succeed())
+		})
+
+		It("should cause the readiness logic to report \"not ready\"", func() {
+			managedOCSResource := types.NamespacedName{
+				Name:      ManagedOCSName,
+				Namespace: TestNamespace,
+			}
+			ready, err := isReady(k8sClient, managedOCSResource)
```
nb-ohad, line 61: I would prefer to test an actual response from the HTTP server (with an HTTP code) and not function invocation.
Calling the function directly you lose coverage on all surrounding code (including the code governing the returned HTTP status)

readinessProbe/readiness/server_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+			Expect(k8sClient.Status().Update(ctx, managedOCS)).Should(Succeed())
+		})
+
+		It("should cause the readiness logic to report \"ready\"", func() {
+			managedOCSResource := types.NamespacedName{
+				Name:      ManagedOCSName,
+				Namespace: TestNamespace,
+			}
+			ready, err := isReady(k8sClient, managedOCSResource)
```
nb-ohad, line 92: Same as last test

nb-ohad (changes requested):

vendor/modules.txt
renan-campos: Should I leave this in and resolve this?

readinessProbe/main.go
```diff
@@ -0,0 +1,70 @@
...
+	managedOCSResource := types.NamespacedName{
+		Name:      "managedocs",
+		Namespace: namespace,
+	}
+
+	log.Info("server running...")
+	err = readiness.RunServer(k8sClient, managedOCSResource, log)
+	if err != nil {
+		log.Error(err, "server error")
+	}
```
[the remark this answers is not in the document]
renan-campos: I can't find a simple way to log that the server is running after actually starting the server through the ListenAndServe() call. I'l change the log to say that the server is starting, and add a log at the end indicating that it terminated.

readinessProbe/readiness/server_test.go
renan-campos: We can skip the extra update in the first test, but in this test the resource already exists, so the update is needed to change the cluster state.

testutils/utils.go
```diff
@@ -0,0 +1,45 @@
...
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	return key
+}
+
+func ReadinessProbe() int {
+	resp, err := http.Get("http://localhost:8081/readyz")
+	if err != nil {
+		return -1
+	}
+	return resp.StatusCode
```
nb-ohad, line 44: Please do not use a valid `int` value as an indication of an error (even if it is negative) 
Go support multi return value paradigm so you can return a second return value fo type `error`

I also would suggest renaming to a verb, maybe `sampleRediness` or `getRediness`.
Another option is to log and `panic` in that case

readinessProbe/readiness/suite_test.go
```diff
@@ -0,0 +1,104 @@
...
+	// Client to be use by the test code, using a non cached client
+	k8sClient, err = client.New(cfg, options)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(k8sClient).ToNot(BeNil())
+
+	go RunServer(k8sClient, types.NamespacedName{
+		Name:      ManagedOCSName,
+		Namespace: TestNamespace,
+	},
+		ctrl.Log.WithName("readiness"))
```
nb-ohad, line 95: Something with the spacing is wrong here. please correct

readinessProbe/readiness/server_test.go
```diff
@@ -0,0 +1,74 @@
...
+				},
+			}
+
+			Expect(k8sClient.Create(ctx, managedOCS)).Should(Succeed())
+
+			utils.WaitForResource(k8sClient, ctx, managedOCS, timeout, interval)
+		})
+
+		It("should cause the readiness logic to report \"not ready\"", func() {
+			Expect(utils.ReadinessProbe()).To(Equal(http.StatusServiceUnavailable))
```
nb-ohad, line 47: In relation to the comment on the `testutils`, this following suggestion will also add coverage for the error case
```suggestion
                         status, err := utils.ReadinessProbe()
                         Expect(err).ToNot(HaveOccurred())
			Expect(status).To(Equal(http.StatusServiceUnavailable))
```

nb-ohad (changes requested):

readinessProbe/readiness/server_test.go
nb-ohad: ack

readinessProbe/main.go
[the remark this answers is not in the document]
nb-ohad: ack

vendor/modules.txt
nb-ohad: Yes you can resolve

testutils/utils.go
renan-campos: called it ```ProbeReadiness```

nb-ohad:
@JohnStrunk Can you please merge

PR #22, opened 2021-03-09
Support clean uninstallation of OCS

kesavanvt:
When on OCS addon uninstall, the OCM creates a config map with deletion label in `openshift-storage` namespace. The OCS converged addon controller will watch for this deletion label and proceed on removing the OCS resources. 


controllers/managedocs_controller.go
```diff
@@ -110,53 +142,89 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+			r.managedOCS.DeletionTimestamp.IsZero() &&
+			r.isUninstallValid() {
+			ok, err := r.isPvcRemoved()
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			if !ok {
+				return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
+			}
+			r.Log.Info("Uninstallation of OCS started")
```
nb-ohad, line 208: To match all other messages, we should use the same tense
```suggestion
			r.Log.Info("Starting OCS uninstallation")
```

controllers/managedocs_controller.go
```diff
@@ -110,53 +142,89 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+
+			r.Log.Info("Deleting managedOCS")
+			if _, err = r.deleteResource(r.Client, &v1.ManagedOCS{
+				ObjectMeta: metav1.ObjectMeta{Name: r.managedOCS.Name, Namespace: r.namespace},
+			}, r.managedOCS.Name); err != nil {
```
nb-ohad, line 213: Why do we need another ManagedOCS structure? Can we just use `r.managedOCS`? It is already contain the correct name and namespace

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
 
 	return components
 }
+
+func (r *ManagedOCSReconciler) isUninstallValid() bool {
```
nb-ohad, line 327: I think that `isUninstallValid` is a misleading name is misleading that does not describe the action taken by the function. 

This function does not validate anything it just checks if a condition is present. 
To me `isUninstallValid` suggest that we are validating an uninstallation process. 

I would suggest renaming it to something like `shoukdUninstall` or `checkUninstallCondiotion`

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+				return result, nil
+			}
+		}
+	}
+	r.Log.Info("CSV removed successfully")
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ManagedOCSReconciler) isComponentReady() bool {
```
nb-ohad, line 381: `Status.Components` is designed to reflect the status of many components and this check is specific for uninstall. As such I will suggest renaming to a more descriptive name

```suggestion
func (r *ManagedOCSReconciler) areComponentsReadyForUninstall() bool {
```

main.go
```diff
@@ -110,6 +118,21 @@ func main() {
 	}
 }
 
+// getUnrestrictedClient creates a client required for listing PVCs from all namespaces.
+func getUnrestrictedClient() client.Client {
+	var options client.Options
+
+	options.Scheme = runtime.NewScheme()
+	utilruntime.Must(clientgoscheme.AddToScheme(options.Scheme))
+	utilruntime.Must(v1.AddToScheme(options.Scheme))
```
nb-ohad, line 127: You are redefining the same add-to scheme that is already defined in `init()` (and you are even missing some). 
I will suggest refactoring `init()` to call a function `addAllSchemes` that adds everything to a given scheme then use it both there and here. 

For the init case it will be ``addAllSchemes(sceme)`` and here it will be `addAllSchemes(options.Scheme)`

main.go
```diff
@@ -89,10 +93,14 @@ func main() {
...
-		Scheme:               mgr.GetScheme(),
-		AddonParamSecretName: fmt.Sprintf("addon-%v-parameters", envVars[addonNameEnvVarName]),
+		Client:                  mgr.GetClient(),
+		UnrestrictedClient:      getUnrestrictedClient(),
+		Log:                     ctrl.Log.WithName("controllers").WithName("ManagedOCS"),
+		Scheme:                  mgr.GetScheme(),
+		AddonParamSecretName:    fmt.Sprintf("addon-%v-parameters", envVars[addonNameEnvVarName]),
+		DeleteConfigMapName:     envVars[addonNameEnvVarName],
+		DeleteConfigMapLabelKey: fmt.Sprintf("api.openshift.com/addon-%v-delete", envVars[addonNameEnvVarName]),
+		AddonSubscriptionName:   fmt.Sprintf("addon-%v", envVars[addonNameEnvVarName]),
```
nb-ohad, line 103: We have 4 instances of `envVars[addonNameEnvVarName]` in this code section. Can we please refactor into local var ```namespace := envVars[addonNameEnvVarName]``` just before line 95, and use it instead?

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+	err = cl.Get(context.Background(), types.NamespacedName{Name: resourceName,
+		Namespace: r.namespace}, resource)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+	r.Log.Info("Uninstall: Waiting for resource to be deleted", "resource", resourceName)
+	return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
```
nb-ohad, line 425: Why do we need to re-queue manually, if it is an important resource we should watch it

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+		if scName == storageClassCephFsName || scName == storageClassRbdName {
+			r.Log.Info("Found Consumer PVCs using OCS storageclasses, cannot proceed on uninstalltion")
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (r *ManagedOCSReconciler) deleteResource(cl client.Client, resource runtime.Object, resourceName string) (reconcile.Result, error) {
+	err := cl.Delete(context.Background(), resource, client.PropagationPolicy(metav1.DeletePropagationForeground))
```
nb-ohad, line 408:
```suggestion
	err := cl.Delete(r.ctx, resource, client.PropagationPolicy(metav1.DeletePropagationForeground))
```

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+		return ctrl.Result{}, err
+	}
+	err = cl.Get(context.Background(), types.NamespacedName{Name: resourceName,
+		Namespace: r.namespace}, resource)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
```
nb-ohad, line 423: Why do we need this? If sync between deletes is needed the check should be a condition for the next delete not a condition for completion on the current delete.

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+
+	if subComponent.StorageCluster.State != v1.ComponentReady {
+		return false
+	}
+	return true
+}
+
+func (r *ManagedOCSReconciler) isPvcRemoved() (bool, error) {
+	pvcList := &corev1.PersistentVolumeClaimList{}
+	err := r.UnrestrictedClient.List(context.Background(), pvcList)
```
nb-ohad, line 392:
```suggestion
	err := r.UnrestrictedClient.List(r.ctx, pvcList)
```

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+	if err != nil {
+		r.Log.Error(err, "Unable to list PVCs")
+		return false, err
```
nb-ohad, line 395: Please return a proper err and not the general one, for example:
```
return false, fmt.ErrorF(Unable to list PVCs: %v, err)
```

If you do that I am not even sure we will need the log.Error any more

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+	}
+	for _, pvc := range pvcList.Items {
+		scName := *pvc.Spec.StorageClassName
+		if scName == storageClassCephFsName || scName == storageClassRbdName {
+			r.Log.Info("Found Consumer PVCs using OCS storageclasses, cannot proceed on uninstalltion")
+			return false, nil
```
nb-ohad, line 401: same as last comment

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+	if err != nil {
+		r.Log.Error(err, "Unable to list CSVs")
+		return ctrl.Result{}, err
+	}
+	for _, csv := range csvList.Items {
+		if strings.HasPrefix(csv.Name, deployerCsvPrefix) {
+			deployerCsvName := csv.Name
+			if result, err := r.deleteResource(r.Client, &operators.ClusterServiceVersion{
+				ObjectMeta: metav1.ObjectMeta{Name: deployerCsvName, Namespace: r.namespace},
+			}, deployerCsvName); err != nil {
```
nb-ohad, line 369: As a matter of readability, let's not create and inline structs inside if conditions. 

```suggestion
			csv := &operators.ClusterServiceVersion{
				ObjectMeta: metav1.ObjectMeta{
					Name: deployerCsvName, 
					Namespace: r.namespace,
				},
			}
			if result, err := r.deleteResource(r.Client, csv, deployerCsvName); err != nil {
```

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+	}
+	return false
+}
+
+func (r *ManagedOCSReconciler) removeOlmComponents() (reconcile.Result, error) {
+
+	r.Log.Info("Deleting subscription")
+	if result, err := r.deleteResource(r.Client, &operators.Subscription{
+		ObjectMeta: metav1.ObjectMeta{Name: r.AddonSubscriptionName, Namespace: r.namespace},
+	}, r.AddonSubscriptionName); err != nil {
```
nb-ohad, line 350: As a matter of readability, let's not create and inline structs inside if conditions.

```suggestion
	subscription :=  &operators.Subscription{
		ObjectMeta: metav1.ObjectMeta{
			Name: r.AddonSubscriptionName, 
			Namespace: r.namespace,
		},
	}
	if result, err := r.deleteResource(r.Client, subscription, r.AddonSubscriptionName); err != nil {
```

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+		return ctrl.Result{}, err
+	} else if !result.IsZero() {
+		return result, nil
+	}
```
nb-ohad, line 354: why do we need to wait for the subscription to delete before deleting the CSV?

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -90,17 +102,37 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			},
 		),
 	}
+	deleteLabelWatchHandler := handler.EnqueueRequestsFromMapFunc{
+		ToRequests: handler.ToRequestsFunc(
+			func(obj handler.MapObject) []reconcile.Request {
+				if obj.Meta.GetName() == r.DeleteConfigMapName {
+					if _, ok := obj.Meta.GetLabels()[r.DeleteConfigMapLabelKey]; ok {
```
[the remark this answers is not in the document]
kesavanvt: `obj.Meta.GetLabels()` could throw empty map in case of no labels and the current code already handles it

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+	cmNamespaceName := types.NamespacedName{
+		Name:      r.DeleteConfigMapName,
+		Namespace: r.namespace,
+	}
+
+	configmap := &corev1.ConfigMap{}
+	err := r.Get(r.ctx, cmNamespaceName, configmap)
+	if err == nil {
+		labels := configmap.Labels
+		if _, ok := labels[r.DeleteConfigMapLabelKey]; !ok {
```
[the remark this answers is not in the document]
kesavanvt: `configmap.labels`  provides empty map in case of no labels and the current code already handles it

controllers/managedocs_controller.go
```diff
@@ -255,3 +323,104 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+			return false
+		}
+		return true
+	}
+	return false
```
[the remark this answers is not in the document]
kesavanvt: Done, checking nil on configmap wasn't required.

controllers/managedocs_controller.go
kesavanvt: Used `checkUninstallCondition`

controllers/managedocs_controller.go
kesavanvt: Here, i don't want to return an error so keeping as such

controllers/managedocs_controller.go
kesavanvt: There is no specific constraint that we need to wait for the subscription to delete before deleting the CSV. So made the changes accordingly.

controllers/managedocs_controller.go
kesavanvt: As there is no strong requirement for waiting on the olm component deletion, removing the requeue logic

controllers/managedocs_controller.go
kesavanvt: removing it based on the other changes, as it not required in our case .

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: I believe nil access in this case is a panic, we do not want that

controllers/managedocs_controller.go
```diff
@@ -110,53 +142,89 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+		}
+		if _, err := ctrl.CreateOrUpdate(r.ctx, r, storageCluster, func() error {
+			return r.generateStorageCluster(reconcileStrategy, storageCluster)
+		}); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		if r.isComponentReady() &&
+			r.managedOCS.DeletionTimestamp.IsZero() &&
+			r.isUninstallValid() {
```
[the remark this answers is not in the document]
kesavanvt: This alignment was the result of gofmt, the other approach would be sticking to a single line conditions

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: will it provide an empty map even in the case where the resource does not have a `metadata` or `metadata.labels` section? my experience says not, but I might be mistaken

controllers/managedocs_controller.go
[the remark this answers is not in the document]
kesavanvt: When there is no label defined in the resource, obj.Meta.GetLabels() returns `map[]` and not nil. Thus here `ok` will hold either true or false based on the availability of label key.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
kesavanvt: When there is no `metadata.labels` defined in the resource, `configmap.Labels` returns map[] and not nil. Thus here `ok` will hold either true or false based on the availability of label key.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: Ack

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: ack

controllers/managedocs_controller.go
```diff
@@ -255,3 +344,138 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      deployerCsvName,
+					Namespace: r.namespace,
+				},
+			}
+			if err := r.deleteResource(r.Client, csv, deployerCsvName); err != nil {
+				return err
+			}
+		}
+	}
```
nb-ohad, line 402: This loop will not short circuit even after finding the correct CSV. I will suggest separating the finding of the CSV and the action we need to take on that CSV (deleting it)

```suggestion
        var csv operators.ClusterServiceVersion = nil
	for _, candidate := range csvList.Items {
		if strings.HasPrefix(csv.Name, deployerCsvPrefix) {
			csv = candidate
			break
		}
	}
	if err := r.deleteResource(r.Client, csv); err != nil {
		return err
	}
```

controllers/managedocs_controller.go
```diff
@@ -255,3 +344,138 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+		return false
+	}
+	return true
+}
+
+func (r *ManagedOCSReconciler) deleteComponents() error {
+	r.Log.Info("Deleting storageCluster")
+	if err := r.deleteResource(r.Client, &ocsv1.StorageCluster{
+		ObjectMeta: metav1.ObjectMeta{Name: storageClusterName, Namespace: r.namespace},
+	}, storageClusterName); err != nil {
```
nb-ohad, line 421: Let's not inline object creation inside a function call, it makes the code unreadable. 
Also why do we need to create a `storagecluster` at all, we already create an load one in `reconcilePhases`

controllers/managedocs_controller.go
```diff
@@ -255,3 +344,138 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+		return err
+	}
+	return nil
+}
+
+func (r *ManagedOCSReconciler) areComponentsDeleted() bool {
```
nb-ohad, line 427: The function name does not match the implementation, not finding a resource does not mean it was deleted. 
Either  move the deletion timestamp here (to indicate that the function is talking about deleting, or rename to something like `verifyComponentsDoNotExists`

controllers/managedocs_controller.go
```diff
@@ -255,3 +344,138 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+	err := cl.Delete(r.ctx, resource)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil
+		}
+		r.Log.Error(err, "Uninstall: unable to delete resource", "resource", resourceName)
+		return err
+	}
+	return nil
+}
```
nb-ohad, line 462: Two observations:
1. you do not need to pass `resourceName` as it can be extracted from the passed resource.
2. I do not see a point for this function as we can just replace it (on the call site) with a single liner 
```
if err := r.client.delete(r.ctx, resource); err != nil && !errors.IsNotFound(err) {
```

controllers/managedocs_controller.go
```diff
@@ -255,3 +344,138 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+
+func contains(slice []string, s string) bool {
+	for _, item := range slice {
+		if item == s {
+			return true
+		}
+	}
+	return false
+}
```
nb-ohad, line 471: Please move this to our utils file

controllers/managedocs_controller.go
```diff
@@ -255,3 +344,138 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+
+func remove(slice []string, s string) (result []string) {
+	for _, item := range slice {
+		if item == s {
+			continue
+		}
+		result = append(result, item)
+	}
+	return
+}
```
nb-ohad, line 481: Please move this to our utils file

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -110,53 +140,112 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+		}
+
+		componentsDeleted := r.areComponentsDeleted()
+
+		if componentsDeleted {
+			r.Log.Info("Removing finalizer on managedOCS")
+			r.managedOCS.ObjectMeta.Finalizers = remove(r.managedOCS.ObjectMeta.Finalizers, managedOcsFinalizer)
+			if err := r.Client.Update(context.Background(), r.managedOCS); err != nil {
+				r.Log.Info("Update Error", "MetaUpdateErr", "Failed to remove finalizer from managedOcs")
+				return ctrl.Result{}, err
```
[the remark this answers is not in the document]
kesavanvt: Thanks, also returning original error may also be required. So modified like below
`return ctrl.Result{}, fmt.Errorf("Failed to remove finalizer from managedOCS: %v", err)`

controllers/managedocs_controller.go
```diff
@@ -110,53 +140,112 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+			}
+			r.Log.Info("managedOCS removed successfully")
+		}
+	} else if r.managedOCS.DeletionTimestamp.IsZero() && r.managedOCS.UID != "" {
+		
+		if !contains(r.managedOCS.GetFinalizers(), managedOcsFinalizer) {
+			r.Log.Info("Finalizer not found for managedOCS. Adding finalizer")
```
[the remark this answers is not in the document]
kesavanvt: I believe it should not be logged as warning, The finalizers will be added to managedOCS once it is created and it is expected.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: That is fine, we can resolve this one

controllers/managedocs_controller.go
kesavanvt: The storagecluster object is created in other block under reconcilePhases and thus it will be unavailable for executing the above statement. Have removed the inline object creation.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: As a rule of thumb, if the condition does not end the reconcile loop it is a warning and not an error. So I still think it should be logged as a warning as it does not stop us from reconciling farther

controllers/managedocs_controller.go
kesavanvt: Single liner is precise and thus removing the usage of a function to perform deletion.

controllers/managedocs_controller.go
nb-ohad: Can we just pass it to this function or save it on the reconciler?

kesavanvt: ack

controllers/managedocs_controller.go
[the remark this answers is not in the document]
kesavanvt: Have added changes to add finalizer at the time of managedocs creation and modified the above to log as warning.

controllers/managedocs_controller.go
```diff
@@ -255,3 +340,108 @@ func (r *ManagedOCSReconciler) updateComponents() v1.ComponentStatusMap {
...
+		return err
+	}
+	var csv operators.ClusterServiceVersion
+	for _, candidate := range csvList.Items {
+		if strings.HasPrefix(candidate.Name, deployerCsvPrefix) {
+			csv = candidate
+			break
+		}
+	}
+	if csv.UID != "" {
```
nb-ohad, line 391: This will create an in-memory copy of every CSV it goes over.
A better way will be to use a pointer:
```suggestion
        var csv &operators.ClusterServiceVersion = nil
	for index := range csvList.Items {
	        candidate := &csvList.Items[index]
		if strings.HasPrefix(candidate.Name, deployerCsvPrefix) {
			csv = candidate
			break
		}
	}
	if csv != nil {
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -110,53 +140,108 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	// Load the managed ocs resource
 	r.managedOCS = &v1.ManagedOCS{}
 	if err := r.Get(r.ctx, req.NamespacedName, r.managedOCS); err != nil {
-		return ctrl.Result{}, err
+		if errors.IsNotFound(err) {
+			r.Log.Info("ManagedOCS resource not found")
```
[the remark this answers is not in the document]
kesavanvt: Yes, make sense. Modified

kesavanvt:
@nb-ohad have Updated the tests

controllers/managedocs_controller_test.go
```diff
@@ -322,5 +325,213 @@ var _ = Describe("ManagedOCS controller", func() {
...
+							corev1.ResourceStorage: resource.MustParse("1Gi"),
+						},
+					},
+					StorageClassName: &storageClassRbdName,
+				},
+			}
+			pvc2 := &corev1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-pvc2",
+					Namespace: TestNamespace,
```
renan-campos, line 366: There needs to be a test that covers when the PVC is outside of the namespace that the operator is in.

controllers/managedocs_controller_test.go
```diff
@@ -322,5 +325,213 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				},
+				Spec: operators.ClusterServiceVersionSpec{
+					InstallStrategy: operators.NamedInstallStrategy{
+						StrategyName: "test-strategy",
+						StrategySpec: operators.StrategyDetailsDeployment{
+							DeploymentSpecs: []operators.StrategyDeploymentSpec{},
+						},
+					},
+				},
+			}
```
renan-campos, line 398: [suggestion] I like that there are nested whens for the different cases. Since these objects are being used only within the nested when's scope, I think it would be better for readibility to move initialize these objects there.

controllers/managedocs_controller.go
```diff
@@ -110,53 +140,108 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+			return ctrl.Result{}, err
+		}
+
+		if r.checkUninstallCondition() && r.areComponentsReadyForUninstall() {
+			found, err := r.findOcsPvcs()
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			if found {
+				return ctrl.Result{Requeue: true, RequeueAfter: 10 * time.Second}, nil
```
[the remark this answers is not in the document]
kesavanvt: Thanks, done

controllers/managedocs_controller_test.go
kesavanvt: Moved sub and csv objects to respective when block. As pvc objects are used in more than when conditions, i prefer to keep them here.

controllers/managedocs_controller_test.go
kesavanvt: Modified existing tests to create one of the PVC outside the operator namespace

renan-campos:
LGTM

@JohnStrunk , Ohad asked me to contact you to merge after I was done looking through it. 
Please don't forget to squash the commits.

controllers/managedocs_controller_test.go
renan-campos: ty

kesavanvt:
@JohnStrunk Now its ready to get merged

PR #27, opened 2021-03-25
Added logic to prevent downscaling

bindrad:

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -171,33 +171,31 @@ func (r *ManagedOCSReconciler) generateStorageCluster(
 		return err
 	}
 
+	// Get sds count of storage cluster
+	currSdsCount := 0
+	for _, item := range sc.Spec.StorageDeviceSets {
+		if item.Name == deviceSetName {
+			currSdsCount = item.Count
+			break
+		}
+	}
+
 	// Handle strict mode reconciliation
 	if reconcileStrategy == v1.ReconcileStrategyStrict {
 		// Get an instance of the desired state
 		desired := utils.ObjectFromTemplate(templates.StorageClusterTemplate, r.Scheme).(*ocsv1.StorageCluster)
-		if err := r.setStorageClusterParamsFromSecret(desired); err != nil {
+		if err := r.setStorageClusterParamsFromSecret(desired, currSdsCount); err != nil {
 			return err
 		}
+
 		// Override storage cluster spec with desired spec from the template.
 		// We do not replace meta or status on purpose
-
-		if sc.Spec.StorageDeviceSets == nil {
-			sc.Spec = desired.Spec
-		} else {
-			if sc.Spec.StorageDeviceSets[0].Count <= desired.Spec.StorageDeviceSets[0].Count {
-				sc.Spec = desired.Spec
-				r.Log.Info("Actual storage cluster parameter", "actualCount", sc.Spec.StorageDeviceSets[0].Count)
-			} else {
-				err := fmt.Errorf("cannot set storage device set count: actualCount=%v", sc.Spec.StorageDeviceSets[0].Count)
-				r.Log.Error(err, "Reducing storage device set count is not allowed")
-			}
-		}
+		sc.Spec = desired.Spec
 	}
-
 	return nil
 }
 
-func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.StorageCluster) error {
+func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.StorageCluster, currSdsCount int) error {
 
 	// The addon param secret will contain the capacity of the cluster in Ti
 	// size = 1,  creates a cluster of 1 Ti capacity
@@ -210,22 +208,22 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
 		&addonParamSecret,
 	); err != nil {
 		// Do not create the StorageCluster if the we fail to get the addon param secret
-		return fmt.Errorf("failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
+		return fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
 	}
 	addonParams := addonParamSecret.Data
 
 	sizeAsString := string(addonParams[storageClassSizeKey])
+	r.Log.Info("Requested add-on settings", storageClassSizeKey, sizeAsString)
 	sdsCount, err := strconv.Atoi(sizeAsString)
 	if err != nil {
-		return fmt.Errorf("invalid storage cluster size value: %v", sizeAsString)
+		return fmt.Errorf("Invalid storage cluster size value: %v", sizeAsString)
 	}
 
-	r.Log.Info("Storage cluster parameters", "count", sdsCount)
-
 	var ds *ocsv1.StorageDeviceSet = nil
-	for _, item := range sc.Spec.StorageDeviceSets {
+	for index := range sc.Spec.StorageDeviceSets {
+		item := &sc.Spec.StorageDeviceSets[index]
 		if item.Name == deviceSetName {
-			ds = &item
+			ds = item
 			break
 		}
 	}
@@ -233,7 +231,15 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
 		return fmt.Errorf("cloud not find default device set on stroage cluster")
 	}
 
-	sc.Spec.StorageDeviceSets[0].Count = sdsCount
+	// Prevent downscaling by comparing count from secret and count from storage cluster
+	r.Log.Info("Setting storage device set count", "Current", currSdsCount, "New", sdsCount)
+	if currSdsCount <= sdsCount {
+		ds.Count = sdsCount
+	} else {
+		r.Log.V(-1).Info("Requested storage device set count will result in downscaling, which is not supported. Skipping")
+		ds.Count = currSdsCount
+	}
+
 	return nil
 }
 
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -175,14 +175,23 @@ func (r *ManagedOCSReconciler) generateStorageCluster(
 	if reconcileStrategy == v1.ReconcileStrategyStrict {
 		// Get an instance of the desired state
 		desired := utils.ObjectFromTemplate(templates.StorageClusterTemplate, r.Scheme).(*ocsv1.StorageCluster)
-
 		if err := r.setStorageClusterParamsFromSecret(desired); err != nil {
 			return err
 		}
 		// Override storage cluster spec with desired spec from the template.
 		// We do not replace meta or status on purpose
 
-		sc.Spec = desired.Spec
+		if sc.Spec.StorageDeviceSets == nil {
+			sc.Spec = desired.Spec
+		} else {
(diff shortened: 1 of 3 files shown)


controllers/managedocs_controller.go
```diff
@@ -175,14 +175,23 @@ func (r *ManagedOCSReconciler) generateStorageCluster(
...
 			return err
 		}
 		// Override storage cluster spec with desired spec from the template.
 		// We do not replace meta or status on purpose
 
-		sc.Spec = desired.Spec
+		if sc.Spec.StorageDeviceSets == nil {
+			sc.Spec = desired.Spec
+		} else {
+			if sc.Spec.StorageDeviceSets[0].Count <= desired.Spec.StorageDeviceSets[0].Count {
```
nb-ohad, line 187: It is a bad idea to select a device set by index, as it might change in future changes to the code. Always search an item by name.

controllers/managedocs_controller.go
```diff
@@ -175,14 +175,23 @@ func (r *ManagedOCSReconciler) generateStorageCluster(
...
+				sc.Spec = desired.Spec
+				r.Log.Info("Actual storage cluster parameter", "actualCount", sc.Spec.StorageDeviceSets[0].Count)
+			} else {
```
nb-ohad, line 190: In this case, you are just ignoring all other reconcile changes by not applying `desired.Spec` `to sc.Spec`. But this is not the desired behavior, other changes (not related to the sds count) should still be reflected

controllers/managedocs_controller.go
```diff
@@ -175,14 +175,23 @@ func (r *ManagedOCSReconciler) generateStorageCluster(
...
+				err := fmt.Errorf("cannot set storage device set count: actualCount=%v", sc.Spec.StorageDeviceSets[0].Count)
+				r.Log.Error(err, "Reducing storage device set count is not allowed")
+			}
+		}
```
nb-ohad, line 194: Is there a reason this entire block should reside here and not where the actual setting of the count is?
By doing it this way you split the logic of setting the count into 2 distinct places and, which is harder to reason about, as will force you to reimplement some of the logic that is already implemented inside the `setStorageClusterParamsFromSecret` func (like finding the device set by name)

controllers/managedocs_controller_test.go
```diff
@@ -283,6 +284,94 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				scKey := utils.GetResourceKey(sc)
+				Expect(k8sClient.Get(ctx, scKey, sc)).Should(Succeed())
+
+				secret := &corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      TestAddOnParamsSecretName,
+						Namespace: TestNamespace,
+					},
+					Data: map[string][]byte{
+						"size": []byte("2"),
```
nb-ohad, line 315: We do not support a value of 2, only 1 or 4. I suggest using 1 and upscaling to 4

controllers/managedocs_controller_test.go
```diff
@@ -283,6 +284,94 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+				sdsCount, _ := strconv.Atoi(string(secret.Data["size"]))
+
+				scGen := sc.ObjectMeta.Generation
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, scKey, sc)
+					return err == nil && sc.ObjectMeta.Generation > scGen
+				}, timeout, interval).Should(BeTrue())
+
+				Expect(sdsCount == sc.Spec.StorageDeviceSets[0].Count).Should(BeTrue())
```
nb-ohad, line 327: Addressing a specific device set by index is a bad idea as its location in the array is not guaranteed. Please change the code to search by name.

controllers/managedocs_controller_test.go
```diff
@@ -283,6 +284,94 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					},
+				}
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+				sdsCount, _ := strconv.Atoi(string(secret.Data["size"]))
+
+				scGen := sc.ObjectMeta.Generation
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, scKey, sc)
+					return err == nil && sc.ObjectMeta.Generation > scGen
+				}, timeout, interval).Should(BeTrue())
```
nb-ohad, line 325: This looks like a perfect case for a utils function (similar to `utils.WaitForResource`), something like `utils.WaitForResourceUpdate`

controllers/managedocs_controller_test.go
```diff
@@ -283,6 +284,94 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      TestAddOnParamsSecretName,
+						Namespace: TestNamespace,
+					},
+					Data: map[string][]byte{
+						"size": []byte("2"),
+					},
+				}
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+				sdsCount, _ := strconv.Atoi(string(secret.Data["size"]))
```
nb-ohad, line 319: Why do we need to read it, when you have hard-coded it to 2?

controllers/managedocs_controller_test.go
```diff
@@ -283,6 +284,94 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      TestAddOnParamsSecretName,
+						Namespace: TestNamespace,
+					},
+					Data: map[string][]byte{
+						"size": []byte("1"),
+					},
+				}
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+				sdsCount, _ := strconv.Atoi(string(secret.Data["size"]))
```
nb-ohad, line 363: Same comment as in the last test case

controllers/managedocs_controller_test.go
```diff
@@ -283,6 +284,94 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+				scGen := sc.ObjectMeta.Generation
+				Eventually(func() bool {
+					err := k8sClient.Get(ctx, scKey, sc)
+					return err == nil && sc.ObjectMeta.Generation == scGen
+				}, timeout, interval).Should(BeTrue())
+
+				Expect(sdsCount < sc.Spec.StorageDeviceSets[0].Count).Should(BeTrue())
```
nb-ohad, line 371: Same comment as in the last test case

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -210,21 +217,29 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
...
 
 	var ds *ocsv1.StorageDeviceSet = nil
-	for _, item := range sc.Spec.StorageDeviceSets {
+	for index, item := range sc.Spec.StorageDeviceSets {
 		if item.Name == deviceSetName {
-			ds = &item
+			ds = &sc.Spec.StorageDeviceSets[index]
 			break
 		}
 	}
```
nb-ohad, line 228: Using an item will create an in-memory clone of each item in the array, which in this case is a `DeviceSet` object. 
Because you already need the index I will suggest the following to prevent in-memory cloning:

```
	for index := range sc.Spec.StorageDeviceSets {
                 item := &sc.Spec.StorageDeviceSets[index]
		if item.Name == deviceSetName {
			ds = item
			break
		}
	}
```

controllers/managedocs_controller.go
```diff
@@ -210,21 +217,29 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
...
 
-	sc.Spec.StorageDeviceSets[0].Count = sdsCount
+	// Prevent downscaling by comparing count from secret and count from storage cluster
+	if currSdsCount <= sdsCount {
+		ds.Count = sdsCount
+		r.Log.Info("Actual storage cluster parameter", "actualCount", ds.Count)
+	} else {
+		ds.Count = currSdsCount
+		err := fmt.Errorf("Cannot set storage device set count: actualCount=%v", ds.Count)
+		r.Log.Error(err, "Reducing storage device set count is not allowed")
```
nb-ohad, line 240: This condition should not result in aborting the reconcile loop, as such we should log it as a warning and not an error. In this case, there isn't any real reason to create the `error` object at all.

controllers/managedocs_controller_test.go
```diff
@@ -272,17 +272,105 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+				utils.WaitForResourceSpecToUpdate(k8sClient, ctx, sc, timeout, interval)
+
+				var ds *ocsv1.StorageDeviceSet = nil
+				for index, item := range sc.Spec.StorageDeviceSets {
+					if item.Name == deviceSetName {
+						ds = &sc.Spec.StorageDeviceSets[index]
+						break
+					}
+				}
```
nb-ohad, line 323: Same comment as for the similar loop in the controller code

controllers/managedocs_controller_test.go
```diff
@@ -272,17 +272,105 @@ var _ = Describe("ManagedOCS controller", func() {
...
+						Namespace: TestNamespace,
+					},
+					Data: map[string][]byte{
+						"size": []byte("1"),
+					},
+				}
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+
+				time.Sleep(3 * time.Second)
+
```
nb-ohad, line 362: In order to get the new state of the `storagecluster`, you need to read it again (`client.get`) after the sleep
The way it is written now will just look at stale `storageclsuter` info

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -210,21 +217,29 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
...
 	}
 
-	sc.Spec.StorageDeviceSets[0].Count = sdsCount
+	// Prevent downscaling by comparing count from secret and count from storage cluster
+	if currSdsCount <= sdsCount {
+		ds.Count = sdsCount
+		r.Log.Info("Actual storage cluster parameter", "actualCount", ds.Count)
+	} else {
+		ds.Count = currSdsCount
+		r.Log.V(-1).Info("Cannot set storage device set count", "actualCount", ds.Count)
```
nb-ohad, line 240: This log warning does not convey the problem to the reader, please change to something like:

```suggestion
		r.Log.V(-1).Info("Requested storage device set count will result in a downscale, which is not supported. Skipping", "Reqested", sdsCount, "Current", ds.Count)
```

testutils/utils.go
```diff
@@ -43,3 +44,15 @@ func ProbeReadiness() (int, error) {
...
 	return resp.StatusCode, nil
 }
+
+func WaitForResourceSpecToUpdate(k8sClient client.Client, ctx context.Context, obj runtime.Object, timeout time.Duration, interval time.Duration) {
+	accessor, err := meta.Accessor(obj)
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+
+	key := client.ObjectKey{Namespace: accessor.GetNamespace(), Name: accessor.GetName()}
+	scGen := accessor.GetGeneration()
+	Eventually(func() bool {
```
nb-ohad, line 54: Please use `EventuallyWithOffset(1, ...` instead of `Eventually(...` so we will have proper test output

nb-ohad (changes requested):

nb-ohad (approved):

nb-ohad:
@JohnStrunk please merge this one

PR #28, opened 2021-03-29
Deploy and configure a monitoring stack as part of the reconcile loop

nb-ohad:
Deploymemnt is done using a dependencies.yaml file, to instruct OLM to bring and deploy both ocs 4.6 and prometheus-operator bundles as part of ocs-ocd-deployer bundle installation.
While the reconcile loop is responible to create the proper monitoring resources to deploy the monitoring stack


config/rbac/role.yaml
```diff
@@ -15,6 +15,54 @@ rules:
...
+  - list
+  - watch
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - servicemonitors
+  verbs:
+  - get
+  - list
+  - watch
```
kesavanvt, line 65: Can we combine these to one api group?

nb-ohad: This is auto-generated by running `make manifests` and we should not edit that manually

kesavanvt: we can modify the kubebuilder marker to something like `// +kubebuilder:rbac:groups="monitoring.coreos.com",namespace=system,resources={alertmanagerconfigs,alertmanagers,prometheuses,podmonitors,servicemonitors,prometheuserules},verbs=get;list;watch` to get it generated into a single group

nb-ohad: Thanks, Will try that

renan-campos (commented):
Should the monitoring stack affect the operator's readiness?

bundle.Dockerfile
```diff
@@ -5,7 +5,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=ocs-osd-deployer
 LABEL operators.operatorframework.io.bundle.channels.v1=alpha
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.3.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.2.0
```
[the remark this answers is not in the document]
nb-ohad: I am aware, but this downgrade is on purpose, as I encountered some edge cases we are hitting that I was not aware of when we moved to 1.3.0

controllers/managedocs_controller.go
```diff
@@ -131,65 +192,226 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
+	// 	return nil
+	// })
+	// if err != nil {
+	// 	return err
+	// }
+
+	return nil
+}
+
+func (r *ManagedOCSReconciler) reconcileMonitoringResources() error {
```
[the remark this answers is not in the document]
nb-ohad: Prometheus will only watch and react to monitoring resources that match the label selector set on the Prometheus resource. I will add a label to make that clear

controllers/managedocs_controller.go
```diff
@@ -44,71 +45,131 @@ import (
...
 	AddonParamSecretName string
 
-	ctx        context.Context
-	managedOCS *v1.ManagedOCS
-	namespace  string
+	ctx            context.Context
+	managedOCS     *v1.ManagedOCS
+	storageCluster *ocsv1.StorageCluster
+	prometheus     *promv1.Prometheus
+	alertmanager   *promv1.Alertmanager
```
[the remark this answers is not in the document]
nb-ohad: This is not the actual case, as we need all of them also in `getComponentStatus` and later on in the deletion code. 
The reconciler object is acting as a context object for the entire operation and should hold to everything that is not immediately local

controllers/managedocs_controller.go
```diff
@@ -44,71 +45,131 @@ import (
...
 }
 
 // Reconcile TODO
 func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithValues("req.Namespace", req.Namespace, "req.Name", req.Name)
-	log.Info("Reconciling ManagedOCS")
+	log.Info("Starting reconcile for ManagedOCS")
 
 	r.ctx = context.Background()
 	r.namespace = req.Namespace
```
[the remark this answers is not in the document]
nb-ohad: Good catch, I remember removing them.
 It might be the case that these lines probably were re-introduced in a rebase.
 I will remove them

controllers/managedocs_controller.go
```diff
@@ -44,71 +45,131 @@ import (
...
 
-	// Load the managed ocs resource
-	r.managedOCS = &v1.ManagedOCS{}
+	// Setup the reconciler from the request
+	r.reconileSetup(req)
+
+	// Load the managed ocs resource (input)
 	if err := r.Get(r.ctx, req.NamespacedName, r.managedOCS); err != nil {
 		return ctrl.Result{}, err
 	}
```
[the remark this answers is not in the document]
nb-ohad: I disagree.

`ReconcileSetup` is meant as a place to do in-memory setup of the reconciler's memory state (member variables)  before the start of the actual reconcile execution. It should not reach the outside world nor run stuff that has the possibility of failure. This is also the reason why the function should not, and is not, returning an error object (or any result for that matter)

controllers/managedocs_controller.go
```diff
@@ -228,30 +450,10 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
...
-		components.StorageCluster.State = v1.ComponentUnknown
+func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
+	// Ensure managedOCS ownership on a resource
+	if err := ctrl.SetControllerReference(r.managedOCS, resource, r.Scheme); err != nil {
+		return err
 	}
-
-	return components
+	return nil
 }
```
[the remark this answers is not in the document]
nb-ohad: We have more than one call site for `SetControllerReference` (I think 3 or 4) and in every one of them, all the parameters are the same, excluding the resource we would like to own. In addition, all of these parameters are coming from the reconciler itself. This alone is a good reason to have it as a util method.

templates/alertmanager.go
```diff
@@ -0,0 +1,28 @@
...
+*/
+
+package templates
+
+// AlertmanagerTemplate is the template that serves as the base for the Alert Manager deployed by the operator
+const AlertmanagerTemplate = `
+apiVersion: monitoring.coreos.com/v1
+kind: Alertmanager
+spec:
+  replicas: 3
```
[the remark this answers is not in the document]
nb-ohad: Best practice, according to examples, and other engineers with more knowledge on the matter

tools/colors.sh
```diff
@@ -0,0 +1,47 @@
...
+    _colored_echo "${1}" ${CYAN}
+}
+
+function white {
+    _colored_echo "${1}" ${WHITE}
+}
+
+function _colored_echo {
+    echo -e "${2}${1}${NC}"
+}
```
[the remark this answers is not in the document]
nb-ohad: Thanks

controllers/managedocs_controller.go
```diff
@@ -131,65 +192,226 @@ func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
 	return nil
 }
 
-// Set the desired stats for the storage cluster resource
-func (r *ManagedOCSReconciler) generateStorageCluster(
-	reconcileStrategy v1.ReconcileStrategy,
-	sc *ocsv1.StorageCluster,
-) error {
-	r.Log.Info("Reconciling storagecluster", "ReconcileStrategy", reconcileStrategy)
+func (r *ManagedOCSReconciler) getComponentStatus() v1.ComponentStatusMap {
```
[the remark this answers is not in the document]
nb-ohad: Yes, and it is missing

bundle.Dockerfile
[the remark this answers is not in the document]
renan-campos: ok

controllers/managedocs_controller.go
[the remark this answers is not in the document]
renan-campos: True, I didn't think about the deletion code. I think it would be better for clarity to move the context objects into its own structure within the ManagedOCSReconciler object, but I'm fine if you disagree.

controllers/managedocs_controller.go
```diff
@@ -44,72 +45,125 @@ import (
...
 // Reconcile TODO
 func (r *ManagedOCSReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	log := r.Log.WithValues("req.Namespace", req.Namespace, "req.Name", req.Name)
-	log.Info("Reconciling ManagedOCS")
+	log.Info("Starting reconcile for ManagedOCS")
 
-	r.ctx = context.Background()
-	r.namespace = req.Namespace
+	// Setup the reconciler from the request
+	r.reconileSetup(req)
```
renan-campos, line 163:
```misspelling
	r.reconcileSetup(req)
```

controllers/managedocs_controller.go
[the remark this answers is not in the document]
renan-campos: thanks

templates/alertmanager.go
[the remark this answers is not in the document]
renan-campos: Thanks for clarifying.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
renan-campos: Why is it an issue `reconcileSetup` for reach the outside world or return failure?

controllers/managedocs_controller.go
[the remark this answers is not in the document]
renan-campos: ok

controllers/managedocs_controller.go
```diff
@@ -248,32 +241,51 @@ func (r *ManagedOCSReconciler) reconcilePhases() error {
...
-			components.StorageCluster.State = v1.ComponentPending
+			promStatus.State = v1.ComponentReady
 		}
 	} else if errors.IsNotFound(err) {
-		components.StorageCluster.State = v1.ComponentNotFound
+		promStatus.State = v1.ComponentNotFound
 	} else {
-		r.Log.Error(err, "error getting StorageCluster, setting compoment status to Unknown")
-		components.StorageCluster.State = v1.ComponentUnknown
+		r.Log.V(-1).Info("error getting Prometheus, setting compoment status to Unknown")
```
renan-campos, line 271: Is this a way to do warnings? It would be good to have a helper (i.e. r.warning)

controllers/managedocs_controller.go
```diff
@@ -367,15 +378,19 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 	return nil
 }
 
+// reconcileMonitoringResources labels all monitorinf resources (ServiceMonitors, PodeMonitors, and PrometheusRules)
```
renan-campos, line 381: slight typo `monitoring`

renan-campos (changes requested):

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: So what will stay inside the reconciler, as I can argue that everything there is context. include the logger and the client.
Saying that I am resolving it as is

controllers/managedocs_controller.go
nb-ohad: I see the misspell, but in any case, I decided to rename it to `initReconciler`, which make more sense and also address the prev comment regarding the job of this function

controllers/managedocs_controller.go
nb-ohad: Yes, Maybe an abstraction above all logging is needed (still not sure about it), but in any case it should not be part of this PR

controllers/managedocs_controller.go
nb-ohad: fixed

nb-ohad:
> Should the monitoring stack affect the operator's readiness?
Yes, as the monitoring stack is an integral part of the solution

controllers/managedocs_controller.go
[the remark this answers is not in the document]
renan-campos: ok, I didn't see the distinction before but I do now.

controllers/managedocs_controller.go
renan-campos: ok

sabose:
@apahim FYI..if you could review as well

controllers/managedocs_controller.go
```diff
@@ -46,100 +47,155 @@ import (
...
+			&enqueueManangedOCSRequest,
+			monResourcesPredicates,
+		).
+		Watches(
+			&source.Kind{Type: &promv1.PrometheusRule{}},
+			&enqueueManangedOCSRequest,
+			monResourcesPredicates,
+		).
+
+		// Create the controler
```
renan-campos, line 185:
```suggestion
		// Create the controller
```

controllers/managedocs_controller.go
```diff
@@ -314,150 +458,228 @@ func (r *ManagedOCSReconciler) setStorageClusterParamsFromSecret(sc *ocsv1.Stora
...
+	// }
+
+	return nil
 }
 
-func (r *ManagedOCSReconciler) checkUninstallCondition() bool {
-	cmNamespacedName := types.NamespacedName{
-		Name:      r.DeleteConfigMapName,
-		Namespace: r.namespace,
+// reconcileMonitoringResources labels all monitoring resources (ServiceMonitors, PodeMonitors, and PrometheusRules)
```
renan-campos, line 537:
```suggestion
// reconcileMonitoringResources labels all monitoring resources (ServiceMonitors, PodMonitors, and PrometheusRules)
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+
+	setupUninstallConditions := func(
+		shouldAddonConfigMapExist bool,
+		addonConfigMapDeleteLabel string,
+		shouldStorageClusterBeReady bool,
+		shouldPrometheusBeReady bool,
+		shouldAlertmanagerBeReady bool,
+		shouldPVC1Exist bool,
+		shouldPVC2Exist bool,
+	) {
```
renan-campos, line 150: [suggestion] I like this function. I would like to suggest encapsulating the parameters into a struct so that they are explicitly listed when the function is called. I had to keep hopping back to the definition to check which boolean value went to which parameter. Plus if the list where to ever grow, having it in a struct would mean you wouldn't have to modify every call to this function.

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
 				}
-				Expect(k8sClient.Get(ctx, utils.GetResourceKey(managedOCS), managedOCS)).Should(Succeed())
+				utils.EnsureNoResources(k8sClient, ctx, resList, timeout, interval)
+
+				// Remove the secret for future cases
+				Expect(k8sClient.Delete(ctx, secret)).Should(Succeed())
+			})
+		})
+		When("there is an invalid size value in the add-on parameters secret", func() {
+			It("should not create a reconciled resources", func() {
```
renan-campos, line 267:
```suggestion
			It("should not create reconciled resources", func() {
```
or
```suggestion
			It("should not create a reconciled resource", func() {
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
-				sc := &ocsv1.StorageCluster{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      storageClusterName,
-						Namespace: TestNamespace,
-					},
-				}
-				utils.WaitForResource(k8sClient, ctx, sc, timeout, interval)
 			})
 		})
+		When("size is decreased in the add-on parameters secret", func() {
```
renan-campos, line 329:
```suggestion
		When("size is decreased in the add-on parameter's secret", func() {
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				// Updating the Status of the StorageCluster should trigger a reconcile for managed-ocs
 				sc.Status.Phase = "Pending"
-				Expect(k8sClient.Status().Update(ctx, &sc)).Should(Succeed())
+				Expect(k8sClient.Status().Update(ctx, sc)).Should(Succeed())
 			})
-
-			It("should update its installation status", func() {
-				By("reflecting the sc status in the managed-ocs cr")
+			It("should reflect that in the ManagedOCS resource status", func() {
+				By("by seeting Status.Components.StorageCluster.State to Pending")
```
renan-campos, line 366:
```suggestion
				By("by setting Status.Components.StorageCluster.State to Pending")
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				// Updating the Status of the storagecluster resource should trigger a reconcile for managedocs
 				sc.Status.Phase = "Ready"
-				Expect(k8sClient.Status().Update(ctx, &sc)).Should(Succeed())
+				Expect(k8sClient.Status().Update(ctx, sc)).Should(Succeed())
 			})
-
-			It("should update its installation status", func() {
-				By("reflecting the sc status in the managed-ocs cr")
+			It("should reflect that in the ManagedOCS resource status", func() {
+				By("by seeting Status.Components.StorageCluster.State to Ready")
```
renan-campos, line 386:
```suggestion
				By("by setting Status.Components.StorageCluster.State to Ready")
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
-				Expect(k8sClient.Delete(ctx, sc)).Should(Succeed())
-				// Race condition: this needs to occur before reconciliation loop runs.
-				Expect(k8sClient.Get(ctx, utils.GetResourceKey(sc), sc)).Should(
-					WithTransform(errors.IsNotFound, BeTrue()))
-
-				// Wait for the storage cluster to be re created
-				utils.WaitForResource(k8sClient, ctx, sc, timeout, interval)
+		When("prometheus has unavailable replicas", func() {
+			It("should reflect that in the ManagedOCS resource status", func() {
+				By("by seeting Status.Components.Prometheus.State to Pending")
```
renan-campos, line 396:
```suggestion
				By("by setting Status.Components.Prometheus.State to Pending")
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
-				// Verify strict mode
-				managedOCS := &v1.ManagedOCS{
-					ObjectMeta: metav1.ObjectMeta{
-						Name:      ManagedOCSName,
-						Namespace: TestNamespace,
-					},
-				}
+		When("all prometheus replicas are available", func() {
+			It("should reflect that in the ManagedOCS resource status", func() {
+				By("by seeting Status.Components.Prometheus.State to Ready")
```
renan-campos, line 415:
```suggestion
				By("by setting Status.Components.Prometheus.State to Ready")
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				key := utils.GetResourceKey(managedOCS)
+				Eventually(func() v1.ComponentState {
+					Expect(k8sClient.Get(ctx, key, managedOCS)).Should(Succeed())
+					return managedOCS.Status.Components.Prometheus.State
+				}, timeout, interval).Should(Equal(v1.ComponentReady))
+			})
+		})
+		When("alertmanager has unavailable replicas", func() {
+			It("should reflect that in the ManagedOCS resource status", func() {
+				By("by seeting Status.Components.Alertmanager.State to Pending")
```
renan-campos, line 434:
```suggestion
				By("by setting Status.Components.Alertmanager.State to Pending")
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				key := utils.GetResourceKey(managedOCS)
+				Eventually(func() v1.ComponentState {
+					Expect(k8sClient.Get(ctx, key, managedOCS)).Should(Succeed())
+					return managedOCS.Status.Components.Alertmanager.State
+				}, timeout, interval).Should(Equal(v1.ComponentPending))
+			})
+		})
+		When("all alertmanager replicas are available", func() {
+			It("should reflect that in the ManagedOCS resource status", func() {
+				By("by seeting Status.Components.Alertmanager.State to Ready")
```
renan-campos, line 453:
```suggestion
				By("by setting Status.Components.Alertmanager.State to Ready")
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				// Updating the status of the prometheus resource should trigger a reconcile for managedocs
+				am.Status.UnavailableReplicas = 0
+				Expect(k8sClient.Update(ctx, am)).Should(Succeed())
+
+				managedOCS := managedOCSTemplate.DeepCopy()
+				key := utils.GetResourceKey(managedOCS)
+				Eventually(func() v1.ComponentState {
+					Expect(k8sClient.Get(ctx, key, managedOCS)).Should(Succeed())
+					return managedOCS.Status.Components.Alertmanager.State
+				}, timeout, interval).Should(Equal(v1.ComponentPending))
```
renan-campos, line 467:
```suggestion
				}, timeout, interval).Should(Equal(v1.ComponentReady))
```
This should be checking for ready.

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
-						Namespace: TestNamespace,
-					},
-				}
+				// Get an updated storagecluster
+				sc := scTemplate.DeepCopy()
 				scKey := utils.GetResourceKey(sc)
 				Expect(k8sClient.Get(ctx, scKey, sc)).Should(Succeed())
 
-				// Modify the storagecluster spec
+				// Update to empty sepc
```
renan-campos, line 483:
```suggestion
				// Update to empty spec
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
 				spec := sc.Spec.DeepCopy()
 				sc.Spec = ocsv1.StorageClusterSpec{}
 				Expect(k8sClient.Update(ctx, sc)).Should(Succeed())
 
-				// Wait for the storage cluster to be modfied again to reflect that it was reconciled
-				utils.WaitForResourceSpecToUpdate(k8sClient, ctx, sc, timeout, interval)
-
-				// Verify that the storage cluster was reverted to its original state
-				Expect(reflect.DeepEqual(sc.Spec, *spec)).Should(BeTrue())
+				// Wait for the spec changes to be reverted
```
renan-campos, line 488:
```suggestion
				// Reverting secret to reflect the current size.
				secret := addonParamsSecretTemplate.DeepCopy()                                  
                               secret.Data["size"] = []byte("4")
                               Expect(k8sClient.Update(ctx, secret)).Should(Succeed())                                                           

				// Wait for the spec changes to be reverted
```
This test is failing because of the downscaling test above. The secret is set to 1, so when the StorageCluster spec is modified, the deployer reconciles it, but the StorageCluster is made with the Count listed in the secret, since it doesn't see that the current count is higher. This can be fixed by updating the secret.

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
-						Namespace: TestNamespace,
-					},
-					Data: map[string][]byte{
-						"size": []byte("4"),
-					},
-				}
-				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
-
-				utils.WaitForResourceSpecToUpdate(k8sClient, ctx, sc, timeout, interval)
+				// Update to empty sepc
```
renan-campos, line 509:
```suggestion
				// Update to empty spec
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				Consistently(func() *ocsv1.StorageClusterSpec {
+					sc := scTemplate.DeepCopy()
+					Expect(k8sClient.Get(ctx, scKey, sc)).Should(Succeed())
+					return &sc.Spec
+				}, timeout, interval).Should(Equal(&sc.Spec))
+			})
+		})
+		When("the storagecluster resource is deleted", func() {
+			It("should create a new storagecluster in the namespace", func() {
+				// Delete the stroragecluster resource
```
renan-campos, line 549:
```suggestion
				// Delete the storagecluster resource
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				pr := promRuleTemplate.DeepCopy()
+				Expect(k8sClient.Create(ctx, pr)).Should(Succeed())
 
-					utils.EnsureNoResource(k8sClient, ctx, sub, timeout, interval)
+				Eventually(func() bool {
+					return utils.ResourceHasLabel(k8sClient, ctx, pr, monLabelKey, monLabelValue)
+				}, timeout, interval).Should(BeTrue())
+			})
+		})
+		When("the addon config map does not exists while all other uninstall conditions are met", func() {
```
renan-campos, line 644:
```suggestion
		When("the addon config map does not exist while all other uninstall conditions are met", func() {
```

controllers/managedocs_controller_test.go
```diff
@@ -2,639 +2,759 @@ package controllers
...
+				setupUninstallConditions(true, testAddonConfigMapDeleteLabelKey, true, true, false, false, true)
+
+				managedOCS := managedOCSTemplate.DeepCopy()
+				key := utils.GetResourceKey((managedOCS))
+				Consistently(func() error {
+					return k8sClient.Get(ctx, key, managedOCS)
+				}, timeout, interval).Should(Succeed())
+			})
+		})
+		When("All unisntall conditions are met", func() {
```
renan-campos, line 732:
```suggestion
		When("All uninstall conditions are met", func() {
```

renan-campos (commented):
Besides minor typos and the two fixes to tests, looks good to me.

controllers/managedocs_controller_test.go
nb-ohad: Good suggestion, but this is a change that can be introduced in a future PR

controllers/managedocs_controller_test.go
nb-ohad: It is not a single parameter, so it will be parameters` but If you think of add-on parameters as the name of the secret, it can just fit as it is

controllers/managedocs_controller_test.go
nb-ohad: Nice catch

controllers/managedocs_controller_test.go
nb-ohad: I think this actually caught a fundamental problem we have. If we lose the size info on the storage cluster we cannot revert back to its actual state only to the desired state. I can reorder the tests to prevent that from happening, but to solve the underlying problem we will need to think about what does it mean to revert back and what should be the state to revert to.

@renan-campos  I suggest to merge with this single failing test, and let it fail until we have a proper solution

controllers/managedocs_controller_test.go
renan-campos: agreed.

controllers/managedocs_controller_test.go
renan-campos: That works; I haven't thought of it that way.

controllers/managedocs_controller_test.go
renan-campos: Well put. I think an issue should be made here in the repo to capture the context to the problem that you described so it is easily found when it gets addressed.

Do we really want to have a failing test in the codebase? This doesn't need to be immediately addressed, and the suggested change reflects the current workaround: we rely on the secret to keep track of the size. I prefer to have an issue listed in this repo and comments linking it in a passing test over having to remember that there is one test which will fail when I pull and run the tests.

nb-ohad: A failing test describes the result we would like to achieve and it is great as a constant reminder that we need to fix it. Saying that I agree that we should open a github issue to capture the Bug.

nb-ohad: If we do not want to fail the test we need to mark it as a skipped test in the result (not sure how to do it)

nb-ohad: For now, I decided to reorder the tests in a way that they will not fail

renan-campos:
\lgtm

renan-campos (approved):

controllers/managedocs_controller.go
```diff
@@ -46,100 +47,155 @@ import (
...
+	namespace         string
+	reconcileStrategy v1.ReconcileStrategy
 }
 
 // Add necessary rbac permissions for managedocs finalizer in order to set blockOwnerDeletion.
 // +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources={managedocs,managedocs/finalizers},verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=managedocs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=ocs.openshift.io,namespace=system,resources=storageclusters,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="monitoring.coreos.com",namespace=system,resources={alertmanagers,prometheuses},verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="monitoring.coreos.com",namespace=system,resources={podmonitors,servicemonitors,prometheuserules},verbs=get;list;watch;update;patch
```
kesavanvt, line 95: typo, change `prometheuserules` to `prometheusrules`

nb-ohad: Addressed in https://github.com/openshift/ocs-osd-deployer/pull/30

PR #31, opened 2021-04-16
Add changes to integrate pagerduty

kesavanvt:
The Pagerduty service key is fetched from secret created by hive and then loaded into alertmanager config secret. Added rules for secret creation


controllers/managedocs_controller.go
```diff
@@ -52,18 +53,19 @@ const (
...
-	storageClassCephFSName = "ocs-storagecluster-cephfs"
-	deployerCSVPrefix      = "ocs-osd-deployer"
-	monLabelKey            = "app"
-	monLabelValue          = "managed-ocs"
+	managedOCSName               = "managedocs"
+	storageClusterName           = "ocs-storagecluster"
+	prometheusName               = "managed-ocs-prometheus"
+	alertmanagerName             = "managed-ocs-alertmanager"
+	alertmanagerConfigSecretName = "managed-ocs-alertmanager-config-secret"
+	pagerdutySecretName          = "managed-ocs-pagerduty"
```
nb-ohad, line 61: This is a little bit problematic as we have multiple addons backed by the same deployer. 
The way we resolved that for the OCM parameters secret and the addon configmap is by having the secret name provided from the outside when the reconciler is created. And it turns the name of the OCM parameters secret includes the name of the addon.

controllers/managedocs_controller.go
```diff
@@ -551,25 +572,73 @@ func (r *ManagedOCSReconciler) reconcileAlertmanager() error {
...
-	// 	}
+func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
+	r.Log.Info("Reconciling AlertmanagerConfig secret")
+
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.alertmanagerConfigSecret, func() error {
+		if err := r.own(r.alertmanagerConfigSecret); err != nil {
+			return err
+		}
+
+		var pagerdutyServiceKey string
```
nb-ohad, line 583: Please move the variable declaration just above the first use.

controllers/managedocs_controller.go
```diff
@@ -551,25 +572,73 @@ func (r *ManagedOCSReconciler) reconcileAlertmanager() error {
...
+		pagerdutySecretData := r.pagerdutySecret.Data
+		if len(pagerdutySecretData["PAGERDUTY_KEY"]) != 0 {
+			pagerdutyServiceKey = string(pagerdutySecretData["PAGERDUTY_KEY"])
+		}
+		if len(pagerdutySecretData["serviceKey"]) != 0 {
+			pagerdutyServiceKey = string(pagerdutySecretData["serviceKey"])
+		}
+		if pagerdutyServiceKey == "" {
+			return fmt.Errorf("Pagerduty service key is not defined")
+		}
```
nb-ohad, line 598: [suggestion] This code block can be written in a simpler way:
```suggestion
		pagerdutyServiceKey := pagerdutySecretData["PAGERDUTY_KEY"]
		if pagerdutyServiceKey == "" {
                         pagerdutyServiceKey = pagerdutySecretData["serviceKey"]
		}
		if pagerdutyServiceKey == ""  {
			return fmt.Errorf("Pagerduty service key is not defined")
		}
```

In any case, please do use one form of checking that a string is empty, either `s == ""` or `len(s) == 0` but do not switch between the to, as it makes the code more convoluted

nb-ohad: In addition, I think the error message is not specific enough, I will suggest something like: 
"Pagerduty secret does not contain a serviceKey or PAGERDUTY_KEY entries"

controllers/managedocs_controller.go
```diff
@@ -551,25 +572,73 @@ func (r *ManagedOCSReconciler) reconcileAlertmanager() error {
...
+		}{}
+
+		alertmanagerConfig.Route.GroupWait = "30s"
+		alertmanagerConfig.Route.GroupInterval = "5m"
+		alertmanagerConfig.Route.RepeatInterval = "4h"
+		alertmanagerConfig.Route.Receiver = "pagerduty"
+		alertmanagerConfig.Route.Routes[0].Match.Severity = "critical"
+		alertmanagerConfig.Route.Routes[0].Receiver = "pagerduty"
+		alertmanagerConfig.Receivers[0].Name = "pagerduty"
+		alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].ServiceKey = pagerdutyServiceKey
```
nb-ohad, line 628: I would suggest refactoring this into a separate function, something like 
```
func GetAlertManagerConfig(pagerdutyServiceKey secret) runtime.Object {
```

As this code is very verbose and we are going to introduce more configuration into the config for alertmanager (for example DMS configuration)

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -551,25 +572,78 @@ func (r *ManagedOCSReconciler) reconcileAlertmanager() error {
...
+	alertmanagerConfig.Route.Receiver = "pagerduty"
+	alertmanagerConfig.Route.Routes[0].Match.Severity = "critical"
+	alertmanagerConfig.Route.Routes[0].Receiver = "pagerduty"
+	alertmanagerConfig.Receivers[0].Name = "pagerduty"
+	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].ServiceKey = pagerdutyServiceKey
+
+	config, err := yaml.Marshal(alertmanagerConfig)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to parse alertmanager conifg: %v", err)
+	}
```
nb-ohad, line 644: I think this conversion code should be located outside this function, as the original request was to separate the acquisition of the config from the processing and the reconciliation of it.
Also, the cadence of change for the data itself and for the mechanism of processing it. is not the same at all

controllers/managedocs_controller_test.go
```diff
@@ -669,6 +683,65 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				secret := pdSecretTemplate.DeepCopy()
+				Expect(k8sClient.Get(ctx, utils.GetResourceKey(secret), secret)).Should(
+					WithTransform(errors.IsNotFound, BeTrue()),
+				)
+
+				// Ensure, over a period of time, that the resources are not created
+				resList := []runtime.Object{
+					amConfigSecretTemplate.DeepCopy(),
+				}
+				utils.EnsureNoResources(k8sClient, ctx, resList, timeout, interval)
```
nb-ohad, line 698: No need to use the list version of the util

```suggestion
				utils.EnsureNoResource(k8sClient, ctx, amConfigSecretTemplate.DeepCopy(), timeout, interval)
```

controllers/managedocs_controller_test.go
```diff
@@ -669,6 +683,65 @@ var _ = Describe("ManagedOCS controller", func() {
...
+			It("should not create alertmanager config secret", func() {
+				// Create empty pagerduty secret
+				secret := pdSecretTemplate.DeepCopy()
+				Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+
+				// Ensure, over a period of time, that the resources are not created
+				resList := []runtime.Object{
+					amConfigSecretTemplate.DeepCopy(),
+				}
+				utils.EnsureNoResources(k8sClient, ctx, resList, timeout, interval)
```
nb-ohad, line 711: same as last comment

controllers/managedocs_controller_test.go
```diff
@@ -669,6 +683,65 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+				// Remove the secret for future cases
+				Expect(k8sClient.Delete(ctx, secret)).Should(Succeed())
+			})
+		})
```
nb-ohad, line 733: Why do we care about other keys, I see a case for an invalid value (which we cannot test) but an invalid key is not a case.

nb-ohad (changes requested):

nb-ohad (approved):

nb-ohad:
@JohnStrunk Please merge

PR #36, opened 2021-04-20
Replace all text based resource templates with API struct instances

nb-ohad:
The change bring with it the following benefits:
- Type safety, specifically in the cases where the API structure changes upon vendor upgrade.
- Prevent mistakes that are not caught by schema validation (e.g. introduction of fields that are not specified in the API schema).
- A reconcile iteration will execute faster as we do not deserialize yaml strings anymore

### DIFF
--- a/templates/alertmanager.go
+++ b/templates/alertmanager.go
@@ -16,14 +16,16 @@ limitations under the License.
 
 package templates
 
+import (
+	promv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+)
+
 // AlertmanagerTemplate is the template that serves as the base for the Alert Manager deployed by the operator
-const AlertmanagerTemplate = `
-apiVersion: monitoring.coreos.com/v1
-kind: Alertmanager
-spec:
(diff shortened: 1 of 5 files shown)


templates/alertmanager.go
```diff
@@ -16,13 +16,15 @@ limitations under the License.
...
-    matchLabels:
-      app: managed-ocs
-`
+var _3 = int32(3)
+
+var AlertmanagerTemplate = promv1.Alertmanager{
+	Spec: promv1.AlertmanagerSpec{
+		Replicas: &_3,
+	},
+}
```
kesavanvt, line 30: Considering the changes made in alertmanager config secret name
```suggestion
var AlertmanagerTemplate = promv1.Alertmanager{
	Spec: promv1.AlertmanagerSpec{
		Replicas:     &_3,
		ConfigSecret: "managed-ocs-alertmanager-config-secret",
	},
}

```

kesavanvt (approved):

nb-ohad:
@JohnStrunk Please merge this one

PR #37, opened 2021-04-22
DeadMan's snitch integration.

renan-campos:


controllers/managedocs_controller.go
```diff
@@ -125,6 +129,13 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			},
 		),
 	)
+	deadMansSnitchSecretPredicates := builder.WithPredicates(
+		predicate.NewPredicateFuncs(
+			func(meta metav1.Object, _ runtime.Object) bool {
+				return meta.GetName() == r.DeadMansSnitchSecretName
+			},
+		),
+	)
```
nb-ohad, line 138: I believe this is the third predicate that is filtering a secret under a given name, the other 2 are `pagerdutySecretPredicates` and `addonParamsSecretPredicates`. 
Can we merge all of these into a single predicate that will fit a single secret watch? something like `knownSecretsPredicate` which will test for any of the names.

controllers/managedocs_controller.go
```diff
@@ -550,6 +574,20 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+
+		desired := templates.DMSPrometheusRuleTemplate.DeepCopy()
+		r.dmsRule.Spec = desired.Spec
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
```
nb-ohad, line 590: I would suggest to keep a single reconciled entity per reconcile func.
This can be moved to a `reconcileDMSPrometheusRule` func (called from reconcilePhases)

controllers/managedocs_controller.go
```diff
@@ -604,39 +649,87 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
+			&receiver{
+				Name: "deadmanssnitch",
+				WebhookConfigs: []*webhookConfig{
+					&webhookConfig{
+						URL: dmsURL,
+					},
+				},
+			},
+		},
+	}
```
nb-ohad, line 730: What is the reason behind this refactor? I can see a couple of reasons why the change should be rejected:
1. It just creates much more code where it is not needed
2. It creates a list of named types that are used only once, which are not needed because we do not reuse them
3. It introduces pointer logic into the structures 
4. And the most important one: It undoes a conscious decision made by under contributor, just recently, without an explanation.

nb-ohad (changes requested):

controllers/managedocs_controller.go
renan-campos: Please explain the benefit of moving keeping a single reconciled entity per reconcile func.

nb-ohad: The current code division (`reconcilePhases` calling `reconcileXXXX` where XXXX is a specific resource) creates an expectation for a reader/contributor when learning the codebase. 

Creating an exception in this case without a specific reason (unless I am missing something that forces us to put these two `CreateOrUpdate` invocations into the same func), and where the alternative (a `reconcileDMSPrometheusRule` func) has no visible cons (again, unless I am missing something), just introduce noise into the current code structure. 

Now. One can argue that the original code division is not convenient or does not provide value. My answer to that is that without pointing to any specific flaw, any other solution is just a subjective opinion. And as a matter of fact, if we are going to break this structure, let's not do it arbitrarily, which means we should do it for all objects and unfold everything into `reconcilePhases` (I am not really advocating for this, just pointing out that this exception is arbitrary)

controllers/managedocs_controller.go
renan-campos: Please explain why it is necessary for _this_ specific PR to have to do this refactor.

nb-ohad: This PR adds another `.watches()` call on the reconciler, for the `Secret` kind, which is a bad practice as it is not really needed. As such, it looks to me like an opportunity to do this very simple refactoring. 

But, if you prefer to push this change, as a separate PR, immediately after this one merges you are welcome to do so. 

The important part is to stop continuously pushing this kind of additions, which can harm the performance of the reconciler, as now we have 3 watches that trigger 3 separate watch pipelines for every secret change in the namespace

renan-campos: I would be happy to submit a PR focused on refactoring the code once this one has been merged.

nb-ohad: Agreed, consider this one resolved

controllers/managedocs_controller.go
renan-campos: This refactor was necessary for me to accomplish what this PR sets out to do. I struggled to add the needed fields to the unnamed struct that was originally there. I followed your suggestion and wound up with the following yaml:
```
route:                                                                                                                                             
  group_wait: 30s                                                                                                                                  
  group_interval: 5m                                                                                                                               
  repeat_interval: 4h                                                                                                                              
  receiver: pagerduty                                                                                                                              
  routes:                                                                                                                                          
  - receiver: pagerduty                                                                                                                            
    group_wait: 30s                                                                                                                                
    group_interval: 5m                                                                                                                             
    repeat_interval: 4h                                                                                                                            
    match:                                                                                                                                         
      severity: critical                                                                                                                           
  - receiver: deadmanssnitch                                                                                                                       
    group_wait: 5m                                                                                                                                 
    group_interval: 5m                                                                                                                             
    repeat_interval: 5m                                                                                                                            
    match:                                                                                                                                         
      alertname: deadmanssnitch
receivers:
- name: pagerduty
  pagerduty_configs:
  - service_key: FOO
  webhook_configs:
  - {}
- name: deadmanssnitch
  pagerduty_configs:
  - {}
  webhook_configs:
  - url: BAR
 ```
I didn't like that it lists the webhook_configs and pagerduty_configs even when there isn't one there, and couldn't figure out a way to change it. Instead of spending too much time on that I thought it was cleaner and easier to write it in the way that I did. Because of the points you made here, I tried running the operator with the code that generates the above yaml, but it does break the alertmanager with the following error:
```
level=error ts=2021-04-23T20:11:43.662Z caller=coordinator.go:124 component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/config/alertmanager.yaml err="missing URL in webhook config"
``` 
If it needs to be changed to an unnamed struct, I'll need help to write it in a satisfactory way.

To speak to your 4th point (I was confused why it's last on the list if it's the most important), it was not my intention to undo something someone else did. The unnamed struct approach worked when the config was only for pagerduty, I just don't have the abilities to expand it and did what I could to get it to work.

Comment (nb-ohad) [reply controllers/managedocs_controller.go:730]: Ok, I understand the issue and the point takes. Next time, it will be helpful to add this kind of explanation to the PR description. 

Saying that I would prefer the more minimal change that does not introduce pointer logic and sticks more closely to the original contributor's solution (will not introduce empty arrays or structures): 
 ```suggestion
	type PagerDutyConfig struct {
	    ServiceKey string `yaml:"service_key"`
	}
	
	type WebhookConfig struct {
	    Url string `yaml:"url"`
	}

	alertmanagerConfig := struct {
		Route struct {
			Routes         [2]struct {
				GroupWait       string `yaml:"group_wait"`
				GroupInterval  string `yaml:"group_interval"`
				RepeatInterval string `yaml:"repeat_interval"`
				Receiver           string `yaml:"receiver"`
				Match struct {
					Severity string `yaml:"severity"`
				} `yaml:"match"`
			} `yaml:"routes"`
		} `yaml:"route"`
		Receivers [2]struct {
			Name                         string `yaml:"name"`
			PagerdutyConfigs    []PagerDutyConfig `yaml:"pagerduty_configs,omitempty"`
			WebhookConfigs 	 []WebhookConfig   `yaml:"webhooks_configs,omitempty"`
		} `yaml:"receivers"`
	}{}
	
	alertmanagerConfig.Route.Routes[0].GroupWait = "30s"
	alertmanagerConfig.Route.Routes[0].GroupInterval = "5m"
	alertmanagerConfig.Route.Routes[0].RepeatInterval = "4h"
	alertmanagerConfig.Route.Routes[0].Receiver = "pagerduty"
	alertmanagerConfig.Route.Routes[0].Match.Severity = "critical"
	alertmanagerConfig.Route.Routes[0].Receiver = "pagerduty"

	alertmanagerConfig.Route.Routes[1].GroupWait = "5m"
	alertmanagerConfig.Route.Routes[1].GroupInterval = "5m"
	alertmanagerConfig.Route.Routes[1].RepeatInterval = "5m"
	alertmanagerConfig.Route.Routes[1].Receiver = "pagerduty"
	alertmanagerConfig.Route.Routes[1].Match.Severity = "critical"
	alertmanagerConfig.Route.Routes[1].Receiver = "pagerduty"
	alertmanagerConfig.Route.Routes[1].Match.Severity = "critical"
	alertmanagerConfig.Route.Routes[1].Receiver = "deadmansnitch"
	
	alertmanagerConfig.Receivers[0].Name = "pagerduty"
	alertmanagerConfig.Receivers[0].PagerdutyConfigs =  []PagerDutyConfig{{}}
	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].ServiceKey = "pagerdutyServiceKey"
	
	alertmanagerConfig.Receivers[1].Name = "deadmansnitch"
	alertmanagerConfig.Receivers[1].WebhookConfigs = []WebhookConfig{{}}
	alertmanagerConfig.Receivers[1].WebhookConfigs[0].Url = "http://www.deadmansnitch.com"
```

controllers/managedocs_controller.go
```diff
@@ -604,44 +667,71 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
+	alertmanagerConfig.Route.Routes[1].GroupInterval = "5m"
+	alertmanagerConfig.Route.Routes[1].RepeatInterval = "5m"
+	alertmanagerConfig.Route.Routes[1].Receiver = "deadmanssnitch"
+	alertmanagerConfig.Route.Routes[1].Match = make(map[string]string)
+	alertmanagerConfig.Route.Routes[1].Match["alertname"] = "deadmanssnitch"
+
 	alertmanagerConfig.Receivers[0].Name = "pagerduty"
-	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].ServiceKey = pagerdutyServiceKey
+
+	alertmanagerConfig.Receivers[1].Name = "pagerduty"
```
nb-ohad, line 721: This receiver has the same name as the one in position 0 (which has not configuration attached), do we really need 3 receivers? if so, we cannot name them the same name (the name is used by the route to identify a receiver).

controllers/managedocs_controller.go
```diff
@@ -604,44 +667,71 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
+				Receiver       string            `yaml:"receiver,omitempty"`
+				Match          map[string]string `yaml:"match,omitempty"`
 			} `yaml:"routes"`
 		} `yaml:"route"`
-		Receivers [1]struct {
-			Name             string `yaml:"name"`
-			PagerdutyConfigs [1]struct {
-				ServiceKey string `yaml:"service_key"`
-			} `yaml:"pagerduty_configs"`
+		Receivers [3]struct {
```
nb-ohad, line 696: Why 3 receivers? We need pagerduty and dms. What is the 3rd one used for?

controllers/managedocs_controller_test.go
```diff
@@ -278,6 +291,28 @@ var _ = Describe("ManagedOCS controller", func() {
...
+		// alertmanager state set to Pending
+		// deletion ConfigMap created
+		// ------------------------------->         checkUninstallCondition() called, passes
+		// ------------------------------->         areComponentsReadyForUninstall() called, passes.
+		// ------------------------------->         managedOCS resource is deleted.
+		//
+		// Adding this sleep before creating the configmap mitigates this from happening by preventing
+		// all of the changes from occuring in the middle of a single reconcile loop.
+		time.Sleep(time.Millisecond * 25)
+
```
nb-ohad, line 315: The entire setup process (func) is protected by lines 212-214:

```
		// Delete the configmap to ensure that we will not trigger uninstall accidentally
		// via and intermediate state
		configMap := addonConfigMapTemplate.DeepCopy()
		err := k8sClient.Delete(ctx, configMap)
		Expect(err == nil || errors.IsNotFound(err)).Should(BeTrue())
```

This ensures that for the entire duration of the func the config map does not exist which means uninstall will fail on this condition. 
		
Any of the changes in this function cannot be observed before the recreation of the config map (last step in the process) which means that the only reconcile iteration that matters is the one that observed the existence of the configmap. As components status is always updated at the start of the reconcile loop it means that the situation described here cannot happen.

It might be that I am missing something here, if so please reply with an explanation. But in any case, an arbitrary sleep is not a solution as we can never know how long after k8s DB is updated, a watch event will be triggered. And as a general note, a watch trigger timing has no guarantees or bounds, the only guarantee is that it will eventually fire.

nb-ohad (changes requested):

controllers/managedocs_controller.go
renan-campos: This was a mistake on my part, I'll change it back to two.

controllers/managedocs_controller_test.go
renan-campos: > This ensures that for the entire duration of the func the config map does not exist which means uninstall will fail on this condition.

The order of events listed in the comment describes a scenario where this is not the case. This is what was observed to have happened.

nb-ohad:
@renan-campos 
You missed a second invocation of `r.checkUninstallCondition` at line 358. let's fix that and I will approve the PR

nb-ohad (approved):

nb-ohad:
@JohnStrunk Please merge

PR #40, opened 2021-05-05
Readiness probe considers monitoring resource status as well.

renan-campos:
Addresses https://github.com/openshift/ocs-osd-deployer/issues/39

### DIFF
--- a/readinessProbe/readiness/server_test.go
+++ b/readinessProbe/readiness/server_test.go
@@ -3,7 +3,6 @@ package readiness
 import (
 	"context"
 	"net/http"
-	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


readinessProbe/readiness/server_test.go
```diff
@@ -20,59 +20,137 @@ var _ = Describe("ManagedOCS Readiness Probe", func() {
...
+		ctx := context.Background()
+
+		managedOCS := &v1.ManagedOCS{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      ManagedOCSName,
+				Namespace: TestNamespace,
+			},
+		}
+
+		It("should create a managedocs resource", func() {
```
renan-campos, line 33: Will switch to By

readinessProbe/readiness/server_test.go
```diff
@@ -12,67 +11,102 @@ import (
...
+		When("the managedocs resource lists its StorageCluster as not \"ready\"", func() {
+			It("should cause the readiness probe to return StatusServiceUnavailable", func() {
+				Expect(k8sClient.Get(ctx, utils.GetResourceKey(managedOCS), managedOCS)).Should(Succeed())
+
+				managedOCS.Status = v1.ManagedOCSStatus{
 					Components: v1.ComponentStatusMap{
 						StorageCluster: v1.ComponentStatus{
 							State: v1.ComponentPending,
 						},
 					},
```
nb-ohad, line 34: In order to validate that the readiness(or unreadiness)  is affected solely because of the specific component (which is what is tested here), I will also set all other components to Ready.

Saying that I will suggest a util function (declared in this file) that just takes the Tree statues and returns the `managedOCS` object, or alternatively a template mechanism (as we use in the tests for the reconcile loop). 

Without it, we might forget in he future to set some component status when we will introduce new components

readinessProbe/readiness/server_test.go
```diff
@@ -12,67 +11,102 @@ import (
...
+			It("should cause the readiness probe to return StatusServiceUnavailable", func() {
+				Expect(k8sClient.Get(ctx, utils.GetResourceKey(managedOCS), managedOCS)).Should(Succeed())
+
+				managedOCS.Status = v1.ManagedOCSStatus{
+					Components: v1.ComponentStatusMap{
+						Prometheus: v1.ComponentStatus{
+							State: v1.ComponentPending,
+						},
+					},
+				}
```
nb-ohad, line 55: Same as last comment

readinessProbe/readiness/server_test.go
```diff
@@ -12,67 +11,102 @@ import (
...
-			// This test expects the ManagedOCS resource to already have been made.
-			err := k8sClient.Get(ctx, utils.GetResourceKey(managedOCS), managedOCS)
-			Expect(err).ToNot(HaveOccurred())
+				managedOCS.Status = v1.ManagedOCSStatus{
+					Components: v1.ComponentStatusMap{
+						Alertmanager: v1.ComponentStatus{
+							State: v1.ComponentPending,
+						},
+					},
+				}
```
nb-ohad, line 75: same as last comment

readinessProbe/readiness/suite_test.go
```diff
@@ -89,11 +97,31 @@ var _ = BeforeSuite(func(done Done) {
...
+	ctx := context.Background()
+
+	managedOCS := &v1.ManagedOCS{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ManagedOCSName,
+			Namespace: TestNamespace,
+		},
+	}
+	Expect(k8sClient.Create(ctx, managedOCS)).Should(Succeed())
+	utils.WaitForResource(k8sClient, ctx, managedOCS, timeout, interval)
```
nb-ohad, line 109: You just created the`managedOCS` resource and validated the creation., why do we need the WaitForResource on the same resource?

nb-ohad (changes requested):

readinessProbe/readiness/suite_test.go
renan-campos: That is a good point. This additional check was not needed.

readinessProbe/readiness/server_test.go
renan-campos: ack

readinessProbe/readiness/server_test.go
renan-campos: ack

readinessProbe/readiness/server_test.go
```diff
@@ -12,67 +11,101 @@ import (
...
+			prometheusStatus = v1.ComponentPending
+		}
+
+		if alertmanagerReady == true {
+			alertmanagerStatus = v1.ComponentReady
+		} else {
+			alertmanagerStatus = v1.ComponentPending
+		}
+
+		Expect(k8sClient.Get(ctx, utils.GetResourceKey(managedOCS), managedOCS)).Should(Succeed())
```
nb-ohad, line 50: If you decided to assert inside the util (using Expect in this case) please use the WithOffset variants, if you will not use it the report errors will just point to the util and not the use-case.

[Suggestion] 2 other options:  
1. Do the get (and assert it) from the outside, send the `managedOCS` object to the util to be updated then do the update call (and assertion) from the outside

2. [My favourite] do the get and update in the util but do not assert just return the error in case of an error. On hte call site just assert the call to the util 
```
Expect(setupReadinessConditions).Should(Succeed())
```

nb-ohad (changes requested):

nb-ohad (approved):

nb-ohad:
@renan-campos Please squash before we merge

nb-ohad:
@JohnStrunk Please merge

PR #48, opened 2021-05-18
Configured Alert Manager to add SOP links to alerts sent to PagerDuty.

renan-campos:
This was tested by deploying the add-on on an OCM cluster, firing off an alert, then checking PagerDuty for the alert received.

image

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -42,6 +42,7 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         - name: ADDON_NAME
+        - name: SOP_ENDPOINT
       - name: readiness-server
         command:
         - /readinessServer
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -82,6 +82,7 @@ type ManagedOCSReconciler struct {
 	DeployerSubscriptionName     string
 	PagerdutySecretName          string
 	DeadMansSnitchSecretName     string
+	SOPEndpoint                  string
 
 	ctx                      context.Context
 	managedOCS               *v1.ManagedOCS
@@ -657,7 +658,8 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
 // To sanitize the input the yaml is first represented as types in golang.
 func (r *ManagedOCSReconciler) generateAlertmanagerConfig(pagerdutyServiceKey string, dmsURL string) interface{} {
 	type PagerDutyConfig struct {
-		ServiceKey string `yaml:"service_key"`
+		ServiceKey string            `yaml:"service_key"`
+		Details    map[string]string `yaml:"details"`
 	}
 
 	type WebhookConfig struct {
@@ -675,6 +677,7 @@ func (r *ManagedOCSReconciler) generateAlertmanagerConfig(pagerdutyServiceKey st
 				GroupInterval  string            `yaml:"group_interval,omitempty"`
 				RepeatInterval string            `yaml:"repeat_interval,omitempty"`
 				Receiver       string            `yaml:"receiver,omitempty"`
+				GroupBy        []string          `yaml:"group_by,omitempty"`
 				Match          map[string]string `yaml:"match,omitempty"`
 			} `yaml:"routes"`
 		} `yaml:"route"`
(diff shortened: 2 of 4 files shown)


controllers/managedocs_controller.go
```diff
@@ -67,6 +67,7 @@ const (
 	deployerCSVPrefix            = "ocs-osd-deployer"
 	monLabelKey                  = "app"
 	monLabelValue                = "managed-ocs"
+	sopLink                      = "https://red-hat-storage.github.io/ocs-sop/sop/OSD/{{ .GroupLabels.alertname }}.html"
```
nb-ohad, line 70: I would like for this URL to not be hardcoded, lets read it from an env variable that we will set in the CSV (SOP_ENDPOINT).
This way we will be able to develop and test against a mock website and even differentiate between SOPs for the converged, converged-dev (which will be backed by new and unreviewed SOPs as we continue development) and the upcoming Global Service offering (which will have its own set of SOPs).

controllers/managedocs_controller.go
```diff
@@ -704,6 +709,9 @@ func (r *ManagedOCSReconciler) generateAlertmanagerConfig(pagerdutyServiceKey st
 	alertmanagerConfig.Receivers[0].Name = "pagerduty"
 	alertmanagerConfig.Receivers[0].PagerdutyConfigs = []PagerDutyConfig{{}}
 	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].ServiceKey = pagerdutyServiceKey
+	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].SendResolved = false
+	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].Details = make(map[string]string)
+	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].Details["SOP"] = sopLink
```
nb-ohad, line 714:
```suggestion
	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].Details = map[string]string {
	    "SOP": sopLink,
	}
```

controllers/managedocs_controller.go
```diff
@@ -704,6 +709,9 @@ func (r *ManagedOCSReconciler) generateAlertmanagerConfig(pagerdutyServiceKey st
 	alertmanagerConfig.Receivers[0].Name = "pagerduty"
 	alertmanagerConfig.Receivers[0].PagerdutyConfigs = []PagerDutyConfig{{}}
 	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].ServiceKey = pagerdutyServiceKey
+	alertmanagerConfig.Receivers[0].PagerdutyConfigs[0].SendResolved = false
```
nb-ohad, line 712: What does sendResolved accomplish?

nb-ohad (changes requested):

controllers/managedocs_controller.go
renan-campos: I thought it configured if the alert is sent marked as resolved, but  from checking the documentation I see that it actually dictates "Whether or not to notify about resolved alerts." Since it isn't relevant to this PR, I took it out.

config/manager/manager.yaml
```diff
@@ -42,6 +42,8 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         - name: ADDON_NAME
+        - name: SOP_ENDPOINT
+          value: "https://red-hat-storage.github.io/ocs-sop/sop/OSD/{{ .GroupLabels.alertname }}.html"
```
nb-ohad, line 46: This should be set in the CSV in the addon bundle not here
So the var here should be empty (same as ADDON_NAME)

nb-ohad (changes requested):

config/manager/manager.yaml
renan-campos: Why not have a value here? I very often forget to set the ADDON_NAME when deploying from the csv and would prefer not to have another configuration to remember.

nb-ohad: Two reasons:
1. This is an upstream project and as such should not point to a RedHat specific thing
2. I am not sure that the env var will be overridden if it is already set

nb-ohad:
@JohnStrunk Can you please merge this one?

PR #54, opened 2021-05-25
Controlling Rook RBD CSI pod resource limits using the rook-ceph-operator-config ConfigMap. 

renan-campos:


Makefile
```diff
@@ -51,6 +51,7 @@ run: generate fmt vet manifests export_env_vars
 	kubectl create secret generic addon-${ADDON_NAME}-parameters -n ${NAMESPACE} --from-literal size=1 --dry-run=client -oyaml | kubectl apply -f -
 	kubectl create secret generic ${ADDON_NAME}-pagerduty -n ${NAMESPACE} --from-literal PAGERDUTY_KEY="test-key" --dry-run=client -oyaml | kubectl apply -f -
 	kubectl create secret generic ${ADDON_NAME}-deadmanssnitch -n ${NAMESPACE} --from-literal SNITCH_URL="https://test-url" --dry-run=client -oyaml | kubectl apply -f -
+	kubectl create configmap rook-ceph-operator-config -n ${NAMESPACE} --dry-run=client -oyaml | kubectl apply -f -
```
renan-campos, line 54: For the Operator to run properly now, this ConfigMap must needs exist.

controllers/managedocs_controller.go
```diff
@@ -128,13 +130,16 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			},
 		),
 	)
-	addonDeleteConfigMapPredicates := builder.WithPredicates(
+	configMapPredicates := builder.WithPredicates(
```
renan-campos, line 133: Since there are multiple ConfigMaps, I refactored this predicate to watch for both of them.

controllers/managedocs_controller.go
```diff
@@ -895,3 +903,117 @@ func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
...
+		rookConfigMap.Data = make(map[string]string)
+	}
+
+	rbdProvisionerLimits, err := yaml.Marshal([]struct {
+		Name     string                  `yaml:"name"`
+		Resource strResourceRequirements `yaml:"resource"`
+	}{
+		{
+			Name:     "csi-provisioner",
+			Resource: convertResourceRequirementsToString(utils.GetResourceConfig("csi-provisioner")),
```
renan-campos, line 927: Since the `rook-ceph-operator-config configmap` expects the resources requirements to be in yaml format, I use yaml.marshal. However, `GetResourceConfig` returns `corev1.ResourceRequirements` which use Quantity objects to represent their values that do not marshal into the values that are needed. I added the helper `convertResourceRequirementsToString` so the the marshalling occurs properly.

controllers/managedocs_controller_test.go
```diff
@@ -373,13 +402,37 @@ var _ = Describe("ManagedOCS controller", func() {
...
-				By("Creating a prometous resource")
+				By("Creating a prometheus resource")
 				utils.WaitForResource(k8sClient, ctx, promTemplate.DeepCopy(), timeout, interval)
 
 				By("Creating an alertmanager resource")
 				utils.WaitForResource(k8sClient, ctx, amTemplate.DeepCopy(), timeout, interval)
 			})
 		})
+		When("there is no rook-ceph-operator-config ConfigMap", func() {
+			It("should not create reconciled resources", func() {
```
renan-campos, line 413: Me reasoning for enforcing this behavior is as follows: If we can't set the limits on the resources we shouldn't run the resources. Let me know if this is unreasonable and I will change it.

controllers/managedocs_controller.go
```diff
@@ -895,3 +903,116 @@ func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
...
+	if err := r.get(rookConfigMap); err != nil {
+		// Because resource limits will not be set, failure to get the Rook ConfigMap results in failure to reconcile.
+		return fmt.Errorf("Failed to get Rook ConfigMap: %v", err)
+	}
+
+	if rookConfigMap.Data == nil {
+		rookConfigMap.Data = make(map[string]string)
+	}
+
+	rbdProvisionerLimits := utils.MarshalRookResourceRequirements([]utils.RookResourceRequirements{
```
nb-ohad, line 921: The `RookResourceRequirements` is used as a list in more than one place. 
I will suggest to define the following in the util file and use it instead
```
type RookResourceRequirementsList []RookResourceRequirements
```

controllers/managedocs_controller_test.go
```diff
@@ -363,6 +369,29 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+					if _, ok := configMap.Data["CSI_RBD_PROVISIONER_RESOURCE"]; !ok {
+						return false
+					}
+
+					if _, ok := configMap.Data["CSI_RBD_PLUGIN_RESOURCE"]; !ok {
+						return false
+					}
+
+					return true
```
nb-ohad, line 391: 1. I think we should also check that the value of the key is not an empty string.
2. you are not testing the keys for cephfs

```suggestion
		return configMap.Data["CSI_RBD_PROVISIONER_RESOURCE"] != "" &&
		     configMap.Data["CSI_RBD_PLUGIN_RESOURCE"] != "" &&
		     configMap.Data["CSI_CEPHFS_PROVISIONER_RESOURCE"] != "" &&
		     configMap.Data["CSI_CEPHFS_PLUGIN_RESOURCE"] != ""
```

controllers/managedocs_controller_test.go
```diff
@@ -373,13 +402,37 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+				// Ensure there is no rook-ceph-operator-config ConfigMap present
+				configMap := rookConfigMapTemplate.DeepCopy()
+				Expect(k8sClient.Delete(ctx, configMap)).Should(Succeed())
+
+				// Ensure, over a period of time, that the resources are not created
+				utils.EnsureNoResources(k8sClient, ctx, resList, timeout, interval)
+
+				// Recreate the configMap for other tests.
+				Expect(k8sClient.Create(ctx, configMap)).Should(Succeed())
```
nb-ohad, line 433: If you are deleting resources in the test you need to verify and wait for them to be create (by the operator) in the end of the test

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -895,3 +903,117 @@ func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
 	}
 	return nil
 }
+
+func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
+	rookConfigMap := &corev1.ConfigMap{}
+	rookConfigMap.Name = rookConfigMapName
+	rookConfigMap.Namespace = r.namespace
```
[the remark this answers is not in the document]
renan-campos: If its alright, I'd prefer to keep this local to this function since this is the only place that gets used, then move it to the reconciler object if it becomes necessary to do so.

controllers/managedocs_controller.go
```diff
@@ -895,3 +903,117 @@ func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
...
+		rookConfigMap.Data = make(map[string]string)
+	}
+
+	rbdProvisionerLimits, err := yaml.Marshal([]struct {
+		Name     string                  `yaml:"name"`
+		Resource strResourceRequirements `yaml:"resource"`
+	}{
+		{
+			Name:     "csi-provisioner",
+			Resource: convertResourceRequirementsToString(utils.GetResourceRequirements("csi-provisioner")),
```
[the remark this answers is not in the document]
renan-campos: With the updated changes, this call is no longer used.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: 👍

nb-ohad (approved):

nb-ohad (approved):

nb-ohad:
@JohnStrunk  please merge

PR #61, opened 2021-06-10
Added logic to get metrics from in-cluster prometheus

bindrad:
Resources added:
1. Service monitor
2. Role
3. Role binding with prometheuss-k8s
4. Secret-data is fetched from a secret in monitoring namespace

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
@@ -55,21 +56,26 @@ const (
 )
 
 const (
-	managedOCSName               = "managedocs"
-	storageClusterName           = "ocs-storagecluster"
-	prometheusName               = "managed-ocs-prometheus"
-	alertmanagerName             = "managed-ocs-alertmanager"
-	alertmanagerConfigSecretName = "managed-ocs-alertmanager-config-secret"
-	dmsRuleName                  = "dms-monitor-rule"
-	storageClassSizeKey          = "size"
-	deviceSetName                = "default"
-	storageClassRbdName          = "ocs-storagecluster-ceph-rbd"
-	storageClassCephFSName       = "ocs-storagecluster-cephfs"
-	deployerCSVPrefix            = "ocs-osd-deployer"
-	ocsOperatorName              = "ocs-operator"
-	monLabelKey                  = "app"
-	monLabelValue                = "managed-ocs"
-	rookConfigMapName            = "rook-ceph-operator-config"
+	managedOCSName                         = "managedocs"
+	storageClusterName                     = "ocs-storagecluster"
+	prometheusName                         = "managed-ocs-prometheus"
+	alertmanagerName                       = "managed-ocs-alertmanager"
+	alertmanagerConfigSecretName           = "managed-ocs-alertmanager-config-secret"
+	dmsRuleName                            = "dms-monitor-rule"
+	storageClassSizeKey                    = "size"
+	deviceSetName                          = "default"
+	storageClassRbdName                    = "ocs-storagecluster-ceph-rbd"
+	storageClassCephFSName                 = "ocs-storagecluster-cephfs"
+	deployerCSVPrefix                      = "ocs-osd-deployer"
+	ocsOperatorName                        = "ocs-operator"
+	monLabelKey                            = "app"
+	monLabelValue                          = "managed-ocs"
+	rookConfigMapName                      = "rook-ceph-operator-config"
+	k8sMetricsServiceMonitorName           = "k8s-metrics-service-monitor"
+	grafanaDatasourceSecretName            = "grafana-datasources"
+	grafanaDatasourceSecretKey             = "prometheus.yaml"
+	k8sMetricsServiceMonitorAuthSecretName = "k8s-metrics-service-monitor-auth"
+	openshiftMonitoringNamespace           = "openshift-monitoring"
 )
 
 // ManagedOCSReconciler reconciles a ManagedOCS object
@@ -87,17 +93,19 @@ type ManagedOCSReconciler struct {
 	DeadMansSnitchSecretName     string
 	SOPEndpoint                  string
 
-	ctx                      context.Context
-	managedOCS               *v1.ManagedOCS
-	storageCluster           *ocsv1.StorageCluster
-	prometheus               *promv1.Prometheus
-	dmsRule                  *promv1.PrometheusRule
-	alertmanager             *promv1.Alertmanager
-	pagerdutySecret          *corev1.Secret
-	deadMansSnitchSecret     *corev1.Secret
-	alertmanagerConfigSecret *corev1.Secret
-	namespace                string
-	reconcileStrategy        v1.ReconcileStrategy
+	ctx                                context.Context
+	managedOCS                         *v1.ManagedOCS
+	storageCluster                     *ocsv1.StorageCluster
+	prometheus                         *promv1.Prometheus
+	dmsRule                            *promv1.PrometheusRule
+	alertmanager                       *promv1.Alertmanager
+	pagerdutySecret                    *corev1.Secret
+	deadMansSnitchSecret               *corev1.Secret
+	alertmanagerConfigSecret           *corev1.Secret
+	k8sMetricsServiceMonitor           *promv1.ServiceMonitor
+	k8sMetricsServiceMonitorAuthSecret *corev1.Secret
+	namespace                          string
+	reconcileStrategy                  v1.ReconcileStrategy
 }
 
 // Add necessary rbac permissions for managedocs finalizer in order to set blockOwnerDeletion.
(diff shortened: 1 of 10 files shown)


controllers/managedocs_controller.go
```diff
@@ -1085,3 +1136,55 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
+		if r.federateSecret.Data == nil {
+			return fmt.Errorf("no data in federate secret")
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
```
kesavanvt, line 1189:
```suggestion
	return err
```

bindrad: @kesavanvt Done the change

controllers/managedocs_controller.go
```diff
@@ -1085,3 +1136,51 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
+		}
+		if datasources.DataSources[0].BasicAuthUser == "" {
+			return fmt.Errorf("federate secret does not contain a Username entry")
+		}
+		r.federateSecret.Data = map[string][]byte{"Username": []byte(datasources.DataSources[0].BasicAuthUser)}
+		if datasources.DataSources[0].BasicAuthPassword == "" {
+			return fmt.Errorf("federate secret does not contain a Password entry")
+		}
+		r.federateSecret.Data["Password"] = []byte(datasources.DataSources[0].BasicAuthPassword)
+		if r.federateSecret.Data == nil {
```
kesavanvt, line 1180: Why does the null check is done after setting Username and Password

controllers/managedocs_controller.go
```diff
@@ -1085,3 +1136,51 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
+			return err
+		}
+		if datasources.DataSources[0].BasicAuthUser == "" {
+			return fmt.Errorf("federate secret does not contain a Username entry")
+		}
+		r.federateSecret.Data = map[string][]byte{"Username": []byte(datasources.DataSources[0].BasicAuthUser)}
+		if datasources.DataSources[0].BasicAuthPassword == "" {
+			return fmt.Errorf("federate secret does not contain a Password entry")
+		}
+		r.federateSecret.Data["Password"] = []byte(datasources.DataSources[0].BasicAuthPassword)
```
kesavanvt, line 1179: Please use any one way of assigning values to the fields

controllers/managedocs_controller.go
bindrad: To make sure whether federateSecret has the data or not

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1137,51 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+		DataSources []GrafanaDataSource `json:"datasources"`
+	}
+
+	prometheusConfig := secret.Data[grafanaDataSourceSecretKey]
+	datasources := GrafanaDataSourceSecret{}
+
+	err = json.Unmarshal(prometheusConfig, &datasources)
+	if err != nil {
+		return err
+	}
```
nb-ohad, line 1168:
```suggestion
	authInfo := struct {
		DataSources []struct {
			BasicAuthPassword string `json:"basicAuthPassword"`
			BasicAuthUser     string `json:"basicAuthUser"`
		} `json:"datasources"`
	}{}
	if err := json.Unmarshal(prometheusConfig, &datasources); err != nil {
		return err
	}
```

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1137,51 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+			return err
+		}
+		if datasources.DataSources[0].BasicAuthUser == "" {
+			return fmt.Errorf("federate secret does not contain a Username entry")
+		}
+		r.federateSecret.Data = map[string][]byte{"Username": []byte(datasources.DataSources[0].BasicAuthUser)}
+		if datasources.DataSources[0].BasicAuthPassword == "" {
+			return fmt.Errorf("federate secret does not contain a Password entry")
+		}
+		r.federateSecret.Data = map[string][]byte{"Password": []byte(datasources.DataSources[0].BasicAuthPassword)}
```
nb-ohad, line 1180: This will  redefine the data struct and you will lose the username value

controllers/managedocs_controller.go
nb-ohad: This line is guaranteed to succeed because you are defining the data field just above this line

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to create k8s-metrics-service-monitor-auth secret: %v", err)
+	}
+	return nil
+}
+
+func (r *ManagedOCSReconciler) reconcileK8sMetricsServiceMonitor() error {
```
nb-ohad, line 1168: We use uppercase for abbreviations (unless they appear at the start of a variable or function name):

```suggestion
func (r *ManagedOCSReconciler) reconcileK8SMetricsServiceMonitor() error {
```

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
 	}
 	return csv
 }
+
+func (r *ManagedOCSReconciler) reconcileK8sMetricsServiceMonitorAuthSecret() error {
```
nb-ohad, line 1122: We use uppercase for abbreviations (unless they appear at the start of a variable or function name):

```suggestion
func (r *ManagedOCSReconciler) reconcileK8SMetricsServiceMonitorAuthSecret() error {
```

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+		if err := r.own(r.k8sMetricsServiceMonitor); err != nil {
+			return err
+		}
+		desired := templates.K8sMetricsServiceMonitorTemplate.DeepCopy()
+		r.k8sMetricsServiceMonitor.Spec = desired.Spec
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to create k8s-metrics-service-monitor service monitor: %v", err)
+	}
```
nb-ohad, line 1181: Same as the last comment regarding resource names, also it will rarely be a create, most of the time it will be an update. 

```suggestion
	if err != nil {
		return fmt.Errorf("Failed to update k8sMetricsServiceMonitor: %v", err)
	}
```

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+			DataSources []struct {
+				BasicAuthPassword string `json:"basicAuthPassword"`
+				BasicAuthUser     string `json:"basicAuthUser"`
+			} `json:"datasources"`
+		}{}
+
+		grafanaDatasourcesPrometheusConfig := secret.Data[grafanaDatasourceSecretKey]
+
+		if err := json.Unmarshal(grafanaDatasourcesPrometheusConfig, &authInfoStructure); err != nil {
+			return err
```
nb-ohad, line 1143: Please add context to the error, something like:
`fmt.Errorf("Could not unmarshal Grapana datasource data: %v", err)`

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+		}
+		if err := r.own(r.k8sMetricsServiceMonitorAuthSecret); err != nil {
+			return err
+		}
```
nb-ohad, line 1147: As of our coding convention, own should be the first operation inside `CreateOrUpdate`, as it is independent on all other operations.

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+				if value.BasicAuthPassword == "" {
+					return fmt.Errorf("k8s-metrics-service-monitor-auth secret does not contain a Password entry")
+				}
+				r.k8sMetricsServiceMonitorAuthSecret.Data["Password"] = []byte(value.BasicAuthPassword)
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to create k8s-metrics-service-monitor-auth secret: %v", err)
```
nb-ohad, line 1163: Same as the last comment regarding resource names, also it will rarely be a create, most of the time it will be an update.

```suggestion
		return fmt.Errorf("Failed to update k8sMetricsServiceMonitorAuth: %v", err)
```

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+			return fmt.Errorf("Failed to get grafana-datasources secret from openshift-monitoring namespace: %v", err)
+		}
+
+		authInfoStructure := struct {
+			DataSources []struct {
+				BasicAuthPassword string `json:"basicAuthPassword"`
+				BasicAuthUser     string `json:"basicAuthUser"`
+			} `json:"datasources"`
+		}{}
+
```
nb-ohad, line 1139: Remove unneeded space

controllers/managedocs_controller.go
```diff
@@ -1086,3 +1118,66 @@ func getCSVByPrefix(csvList opv1a1.ClusterServiceVersionList, name string) *opv1
...
+		grafanaDatasourcesPrometheusConfig := secret.Data[grafanaDatasourceSecretKey]
+
+		if err := json.Unmarshal(grafanaDatasourcesPrometheusConfig, &authInfoStructure); err != nil {
```
nb-ohad, line 1142: Why create a temp variable for a single access? 

```suggestion
		if err := json.Unmarshal(secret.Data[grafanaDatasourceSecretKey], &authInfoStructure); err != nil {
```

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
```diff
@@ -839,6 +860,60 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				}{}
+				authInfoStructure.DataSources = make([]struct {
+					BasicAuthPassword string "json:\"basicAuthPassword\""
+					BasicAuthUser     string "json:\"basicAuthUser\""
+				}, 1)
+				authInfoStructure.DataSources[0].BasicAuthPassword = "password"
+				authInfoStructure.DataSources[0].BasicAuthUser = "internal"
+				grafanaSecret := grafanaDatasourceSecretTemplate.DeepCopy()
+				grafanaSecret.Data = make(map[string][]byte)
+				grafanaSecret.Data["prometheus.yaml"], _ = json.Marshal(authInfoStructure)
```
nb-ohad, line 880: All of this code should go into the `grafanaDatasourceSecretTemplate` template, the idea behind templates is that you should not need to refill static info every time manually. 

```suggestion
				grafanaSecret := grafanaDatasourceSecretTemplate.DeepCopy()
```

controllers/managedocs_controller_test.go
```diff
@@ -839,6 +860,60 @@ var _ = Describe("ManagedOCS controller", func() {
...
+		When("there is k8s-metrics-service-monitor-auth secret in primary namespace", func() {
+			It("should create k8s-metrics-service-monitor service monitor", func() {
+				secret := testK8sMetricsServiceMonitorAuthSecretTemplate.DeepCopy()
+				secretKey := utils.GetResourceKey(secret)
+				Expect(k8sClient.Get(ctx, secretKey, secret)).Should(Succeed())
+				sm := testK8sMetricsServiceMonitorAuthSecretTemplate.DeepCopy()
+				smKey := utils.GetResourceKey(sm)
+				Expect(k8sClient.Get(ctx, smKey, sm)).Should(Succeed())
+			})
+		})
```
nb-ohad, line 895: This case is irrelevant, the service monitor should be created in any case. 
They are non-coupled (but ordered) operations.

controllers/managedocs_controller_test.go
```diff
@@ -839,6 +860,60 @@ var _ = Describe("ManagedOCS controller", func() {
 				utils.WaitForResource(k8sClient, ctx, amConfigSecretTemplate.DeepCopy(), timeout, interval)
 			})
 		})
+		When("there is grafana-datasources secret in openshift-monitoring namespace", func() {
```
nb-ohad, line 863:
```suggestion
		When("a Grafana datasources secret exists in the openshift-monitoring namespace", func() {
```

nb-ohad (changes requested):

nb-ohad (approved):

PR #62, opened 2021-06-14
create alertmanagerconfig and update the controller accordingly.

crombus:
Signed-off-by: crombus <pkundra@redhat.com>


crombus:
Need a little help in writing controller code. The template passes with no errors.

controllers/managedocs_controller.go
```diff
@@ -319,9 +320,9 @@ func (r *ManagedOCSReconciler) initReconciler(req ctrl.Request) {
 	r.deadMansSnitchSecret.Name = r.DeadMansSnitchSecretName
 	r.deadMansSnitchSecret.Namespace = r.namespace
 
-	r.alertmanagerConfigSecret = &corev1.Secret{}
-	r.alertmanagerConfigSecret.Name = alertmanagerConfigSecretName
-	r.alertmanagerConfigSecret.Namespace = r.namespace
+	r.alertmanagerConfig = &promv1a1.AlertmanagerConfig{}
+	r.alertmanagerConfig.Name = alertmanagerConfigSecretName
```
nb-ohad, line 324: Why do use the secret name? The secret name ends with the word "secret" which does not fit here. 
I will suggest using a name that follows our convention: `managed-ocs-alertmanager-config` in a constant named `alertmanagerConfigName`

controllers/managedocs_controller.go
```diff
@@ -693,14 +694,9 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
-		alertmanagerConfig := r.generateAlertmanagerConfig(pagerdutyServiceKey, dmsURL)
-		config, err := yaml.Marshal(alertmanagerConfig)
-		if err != nil {
-			return fmt.Errorf("Unable to encode alertmanager conifg: %v", err)
-		}
-		r.alertmanagerConfigSecret.Data = map[string][]byte{
-			"alertmanager.yaml": config,
-		}
+		desired := templates.AlertmanagerConfigTemplate.DeepCopy()
+		r.updateAlertmanagerConfig(desired, pagerdutyServiceKey, dmsURL)
```
nb-ohad, line 698: Why do we need to call into a separate func where the entire body of that func is 3 or 4 lines of code?
It just complicates reading this func because it makes the reader jump between functions with no real reason.

Can we just embed the needed changes to the desired state here?

controllers/managedocs_controller.go
```diff
@@ -710,67 +706,10 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
-
-	alertmanagerConfig.Receivers[1].Name = "DeadMansSnitch"
-	alertmanagerConfig.Receivers[1].WebhookConfigs = []WebhookConfig{{}}
-	alertmanagerConfig.Receivers[1].WebhookConfigs[0].Url = dmsURL
-
-	return &alertmanagerConfig
+func (r *ManagedOCSReconciler) updateAlertmanagerConfig(ac *promv1a1.AlertmanagerConfig, pagerdutyServiceKey string, dmsURL string) {
+	ac.Spec.Receivers[0].PagerDutyConfigs[0].ServiceKey = &corev1.SecretKeySelector{Key: pagerdutyServiceKey}
+	ac.Spec.Receivers[0].PagerDutyConfigs[0].Details = []promv1a1.KeyValue{{Key: "SOP"}, {Value: r.SOPEndpoint}}
+	ac.Spec.Receivers[1].WebhookConfigs[0].URL = &dmsURL
```
nb-ohad, line 712: Assuming that a specific receiver is located at a specific index is problematic, the code will break if we add another receiver in the future as the first reciver. 

To mitigate you can just search for the correct receiver, as an example:

```suggestion
for i := range ac.Spec.Receivers {
    receiver := &range ac.Spec.Receivers[i]
    switch receiver.name {
    case "pagerduty":
        receiver.PagerDutyConfigs[0].ServiceKey.Key = pagerdutyServiceKey
        receiver.PagerDutyConfigs[0].Details[0].Key = "SOP"
        receiver.PagerDutyConfigs[0].Details[0].Value = r.SOPEndpoint
    case "DeadManSnitch":
        receiver.WebhookConfigs[0].URL = &dmsURL
    }
}
```

Please note the following (In my suggestion above):
1. The structure itself (arrays, selector, etc.) should move to the template. No reason to create it in code.
2. It is ok to directly access via an index a config for a specific receiver as (e.g. `PagerDutyConfigs[0]`) as this is internal structure for that specific receiver

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -668,11 +668,11 @@ func (r *ManagedOCSReconciler) reconcileAlertmanager() error {
...
 }
 
-func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
+func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 	r.Log.Info("Reconciling AlertmanagerConfig secret")
 
-	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.alertmanagerConfigSecret, func() error {
-		if err := r.own(r.alertmanagerConfigSecret); err != nil {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.alertmanagerConfig, func() error {
+		if err := r.own(r.alertmanagerConfig); err != nil {
```
kesavanvt, line 675: Add necessary rbac for creating/updating AlertmanagerConfig resource.

controllers/managedocs_controller.go
```diff
@@ -693,86 +693,25 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
+		desired := templates.AlertmanagerConfigTemplate.DeepCopy()
+		for i := range desired.Spec.Receivers {
+			receiver := &desired.Spec.Receivers[i]
+			switch receiver.Name {
+			case "pagerduty":
+				receiver.PagerDutyConfigs[0].ServiceKey.Key = pagerdutyServiceKey
+				receiver.PagerDutyConfigs[0].Details[0].Key = "SOP"
+				receiver.PagerDutyConfigs[0].Details[0].Value = r.SOPEndpoint
+			case "DeadManSnitch":
+				receiver.WebhookConfigs[0].URL = &dmsURL
```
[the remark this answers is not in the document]
nb-ohad: @renan-campos I believe the template will:
https://github.com/openshift/ocs-osd-deployer/pull/62/files#diff-58a22c1a0aa1ce7cccfb52548cea9e8198ac45e8fd16bc20d5bcbf34b5aa5c55R66-R68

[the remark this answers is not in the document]
renan-campos: I wrote a small go program to test the way WebhookConfigs object is initialized in the template, and found that an out-of-range error will occur: https://play.golang.org/p/Eox3wLVNZDc

templates/alertmanagerconfig.go
```diff
@@ -0,0 +1,70 @@
...
+		},
+		Receivers: []promv1a1.Receiver{{
+			Name: "pagerduty",
+			PagerDutyConfigs: []promv1a1.PagerDutyConfig{
+				{ServiceKey: &corev1.SecretKeySelector{Key: ""}},
+				{Details: []promv1a1.KeyValue{{Key: ""}, {Value: ""}}},
+			},
+		}, {
+			Name:           "DeadMansSnitch",
+			WebhookConfigs: []promv1a1.WebhookConfig{}}},
```
renan-campos, line 68:
```suggestion
			WebhookConfigs: []promv1a1.WebhookConfig{{}}}},
```

To potentially fix the out-of-range error discussed above.

controllers/managedocs_controller.go
```diff
@@ -690,86 +690,25 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
-			"alertmanager.yaml": config,
+		desired := templates.AlertmanagerConfigTemplate.DeepCopy()
+		for i := range desired.Spec.Receivers {
+			receiver := &desired.Spec.Receivers[i]
+			switch receiver.Name {
+			case "pagerduty":
+				receiver.PagerDutyConfigs[0].ServiceKey.Key = pagerdutyServiceKey
+				receiver.PagerDutyConfigs[0].Details[0].Key = "SOP"
+				receiver.PagerDutyConfigs[0].Details[0].Value = r.SOPEndpoint
+			case "DeadManSnitch":
```
renan-campos, line 701:
```suggestion
			case "DeadMansSnitch":
```

nb-ohad (approved):

controllers/managedocs_controller.go
```diff
@@ -48,28 +47,29 @@ import (
 	"github.com/openshift/ocs-osd-deployer/templates"
 	"github.com/openshift/ocs-osd-deployer/utils"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	promv1a1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
```
kesavanvt, line 50: Its required to register `promv1a1` in the scheme

`utilruntime.Must(promv1a1.AddToScheme(scheme))` (main.go)

templates/alertmanagerconfig.go
```diff
@@ -0,0 +1,70 @@
...
+var AlertmanagerConfigTemplate = promv1a1.AlertmanagerConfig{
+	Spec: promv1a1.AlertmanagerConfigSpec{
+		Route: &promv1a1.Route{
+			Routes: []apiextensionsv1.JSON{
+				convertToApiExtV1JSON(promv1a1.Route{
+					GroupBy:        []string{"alertname"},
+					GroupWait:      "30s",
+					GroupInterval:  "5m",
+					RepeatInterval: "12h",
+					Matchers:       []promv1a1.Matcher{{Name: "severity"}, {Value: "critical"}},
```
kesavanvt, line 45: These are not two separate matchers
```suggestion
					Matchers:       []promv1a1.Matcher{{Name: "severity", Value: "critical"}},
```

templates/alertmanagerconfig.go
```diff
@@ -0,0 +1,70 @@
...
+					Receiver:       "pagerduty",
+				},
+				),
+				convertToApiExtV1JSON(promv1a1.Route{
+					GroupBy:        []string{"alertname"},
+					GroupWait:      "30s",
+					GroupInterval:  "5m",
+					RepeatInterval: "5m",
+					Matchers:       []promv1a1.Matcher{{Name: "alertname"}, {Value: "DeadMansSnitch"}},
```
kesavanvt, line 54: Same here
```suggestion
					Matchers:       []promv1a1.Matcher{{Name: "alertname", Value: "DeadMansSnitch"}},,
```

templates/alertmanager.go
```diff
@@ -29,5 +30,10 @@ var AlertmanagerTemplate = promv1.Alertmanager{
 		Replicas:     &_3,
 		ConfigSecret: "managed-ocs-alertmanager-config-secret",
 		Resources:    utils.GetResourceRequirements("alertmanager"),
+		AlertmanagerConfigSelector: &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"dhruv": "test",
```
nb-ohad, line 35: @bindrad is that here by mistake? if so what should be the actual matcher?

controllers/managedocs_controller.go
```diff
@@ -690,86 +690,27 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
+				receiver.PagerDutyConfigs[0].ServiceKey.Key = "PAGERDUTY_KEY"
+				receiver.PagerDutyConfigs[0].ServiceKey.LocalObjectReference.Name = r.PagerdutySecretName
+				receiver.PagerDutyConfigs[0].Details[0].Key = "SOP"
+				receiver.PagerDutyConfigs[0].Details[0].Value = r.SOPEndpoint
+			case "DeadMansSnitch":
+				receiver.WebhookConfigs[0].URL = &dmsURL
+			}
 		}
+		r.alertmanagerConfig.Spec = desired.Spec
+		r.alertmanagerConfig.ObjectMeta.Labels = map[string]string{"dhruv": "test"}
```
nb-ohad, line 707: @bindrad  is that here by mistake? if so what should be the actual matcher? 
Also, if it is static it should go into the template

nb-ohad (changes requested):

controllers/managedocs_controller.go
bindrad: It was a mistake, I and Pulkit were trying to contribute to each other's PR.

templates/alertmanager.go
bindrad: It was a mistake, I and Pulkit were trying to contribute to each other's PR.

controllers/managedocs_controller.go
```diff
@@ -678,8 +678,13 @@ func (r *ManagedOCSReconciler) reconcileAlertmanager() error {
...
 
 		desired := templates.AlertmanagerTemplate.DeepCopy()
-		r.alertmanager.ObjectMeta.Labels = map[string]string{monLabelKey: monLabelValue}
+		desired.Spec.AlertmanagerConfigSelector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				monLabelKey: monLabelValue,
+			},
+		}
 		r.alertmanager.Spec = desired.Spec
+		r.alertmanager.ObjectMeta.Labels = map[string]string{monLabelKey: monLabelValue}
```
nb-ohad, line 687: This will override all labels on the alertmanager which means that all other labels will be lost. 
Please add the label to an existing `ObjectMeta.labels` instead

controllers/managedocs_controller.go
```diff
@@ -714,86 +719,27 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfigSecret() error {
...
+				receiver.PagerDutyConfigs[0].ServiceKey.Key = "PAGERDUTY_KEY"
+				receiver.PagerDutyConfigs[0].ServiceKey.LocalObjectReference.Name = r.PagerdutySecretName
+				receiver.PagerDutyConfigs[0].Details[0].Key = "SOP"
+				receiver.PagerDutyConfigs[0].Details[0].Value = r.SOPEndpoint
+			case "DeadMansSnitch":
+				receiver.WebhookConfigs[0].URL = &dmsURL
+			}
 		}
+		r.alertmanagerConfig.Spec = desired.Spec
+		r.alertmanagerConfig.ObjectMeta.Labels = map[string]string{monLabelKey: monLabelValue}
```
nb-ohad, line 736: This will override all labels on the alertmanagerconfig which means that all other labels will be lost.
Please add the label to an existing ObjectMeta.labels instead

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -678,8 +678,13 @@ func (r *ManagedOCSReconciler) reconcileAlertmanager() error {
 		}
 
 		desired := templates.AlertmanagerTemplate.DeepCopy()
-		r.alertmanager.ObjectMeta.Labels = map[string]string{monLabelKey: monLabelValue}
+		desired.Spec.AlertmanagerConfigSelector = &metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				monLabelKey: monLabelValue,
+			},
+		}
```
kesavanvt, line 685: I think this can be moved to the alertmanager.go template. Since we are adding the selectors for other resources in the templates. 
`
AlertmanagerConfigSelector: &resourceSelector`

Also , please remove the ConfigSecret from the alertmanager.go template, as we are not using it anymore

nb-ohad: @kesavanvt  I actually disagree here, we have constants that hold the values of the label key and label value. We should use them and not repeat the string values somewhere else (even if it is a template). I will suggest doing one of the following:
Either allow the templates to have access to the const and use them in the template,
or remove the selector from all templates and configure everything in the reconciler where we have access to the consts

nb-ohad (approved):

PR #65, opened 2021-06-18
Fix ocs dependency to use the latest z stream version

kesavanvt:
When a new z stream version is released, previous version is removed from the registry. This fix will pick the latest available z stream version

### DIFF
--- a/config/metadata/dependencies.yaml
+++ b/config/metadata/dependencies.yaml
@@ -6,4 +6,4 @@ dependencies:
 - type: olm.package
   value:
     packageName: ocs-operator
-    version: "4.7.0"
+    version: ">4.7.0 <4.8.0"


config/metadata/dependencies.yaml
```diff
@@ -6,4 +6,4 @@ dependencies:
 - type: olm.package
   value:
     packageName: ocs-operator
-    version: "4.7.0"
+    version: ">4.7.0 < 4.8.0"
```
nb-ohad, line 9: Not sure, but it might be that the extra space between the "<" and the "4.8.0" might cause problems.

nb-ohad (changes requested):

config/metadata/dependencies.yaml
kesavanvt: Done

nb-ohad (approved):

PR #67, opened 2021-06-28
Updating functionality of deploy-with-olm.

renan-campos:


tools/deploy-with-olm.sh
```diff
@@ -47,24 +58,42 @@ exit_on_err make docker-push IMG=${DEPLOYER_IMAGE}
...
 
 # Deploy the bundle
 blue "Deploy/Run the ${BUNDLE_NAME}:${BUNDLE_VERSION} in the cluster (at ${TARGET_NAMESPACE})"
-${K8S_CLIENT} get namespace olm >/dev/null 2>&1
-if [ $? -eq 0 ]; then
-    exit_on_err ${K8S_CLIENT} create namespace ${TARGET_NAMESPACE}
-fi
-exit_on_err ${K8S_CLIENT} apply -f ./tools/ocs-catalogsource.yaml
+
+exit_on_err ${K8S_CLIENT} create namespace ${NAMESPACE} --dry-run=client -o yaml | kubectl apply -f -
```
ezio-auditore, line 84: ${K8S_CLIENT} instead of kubectl as well?

renan-campos: Yes, that's defined in line 19 and comes from the original script. The default value is kubectl.

nb-ohad: We use an env var so you could specify `oc` instead of `kubectl` when you run

config/manager/manager.yaml
```diff
@@ -42,7 +42,9 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         - name: ADDON_NAME
+          value: sed_replace_addon_name
         - name: SOP_ENDPOINT
+          value: sed_replace_sop_endpoint
       - name: readiness-server
```
nb-ohad, line 48: This file is not part of the deploy script or tool, it is part of the general config bundle, we should not touch it just to fix something in the deploy script

tools/deploy-with-olm.sh
```diff
@@ -47,24 +58,42 @@ exit_on_err make docker-push IMG=${DEPLOYER_IMAGE}
 # Generate the olm bundle image
 blue "Generate and push the olm bundle image: ${BUNDLE_IMAGE}"
 exit_on_err make bundle IMG=${DEPLOYER_IMAGE}
+exit_on_err sed -i "s|sed_replace_addon_name|${ADDON_NAME}|" ${BUNDLE_FILE}
+exit_on_err sed -i "s|sed_replace_sop_endpoint|${SOP_ENDPOINT}|" ${BUNDLE_FILE}
```
nb-ohad, line 62: Can we use `yq`, which actually understands the YAML format,  instead of `sed`. 
The benefit will also be that we will not need to introduce keys for the find/replace logic

tools/deploy-with-olm.sh
```diff
@@ -47,24 +58,42 @@ exit_on_err make docker-push IMG=${DEPLOYER_IMAGE}
...
+
+exit_on_err sed "s|sed_replace_addon_name|${ADDON_NAME}|" ${SNITCH_YAML} | \
+sed "s|sed_replace_namespace|${TARGET_NAMESPACE}|" | \
+sed "s|sed_replace_snitch_url|${SNITCH_URL}|" | \
+${K8S_CLIENT} apply -f -
+
+exit_on_err sed -i "s|sed_replace_addon_name|${ADDON_NAME}|" ${PD_YAML} | \
+sed -i "s|sed_replace_namespace|${TARGET_NAMESPACE}|" | \
+sed -i "s|sed_replace_pagerduty_key|${PD_KEY}|" | \
+${K8S_CLIENT} apply -f -
```
nb-ohad, line 97: Why are we trying to configure PD and DMS as part of the deployment?
We can just create the resources and update them, if needed, in another operation after deployment. 
 
 Doing it here just complicated the deployment and add more steps that might fail

tools/yamls/pd-service-key-secret.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: ocs-converged-pagerduty
+  namespace: sed_replace_namespace
+data:
+  PAGERDUTY_KEY: sed_replace_pagerduty_key
```
nb-ohad, line 8: This entire file can be replaced with a `kubectl create secret ...` command

tools/yamls/snitch-url-secret.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: sed_replace_addon_name-deadmanssnitch
+  namespace: sed_replace_namespace
+data:
+  SNITCH_URL: sed_replace_snitch_url
```
nb-ohad, line 8: This entire file can be replaced with a `kubectl create secret ...` command

nb-ohad (changes requested):

tools/deploy-with-olm.sh
renan-campos: Oh, I used `kubectl` instead of `${K8S_CLIENT}` at the end of this line by accident. Good catch I'll change that.

tools/deploy-with-olm.sh
renan-campos: I added it here because I prefer to have the operator start healthy in an environment where all of its expected prerequisites are set up, which are:
- namespace 
- addon parameter secret
- dms secret
- pd secret

The operator will start in an error loop if PD and DMS are not configured. I don't see why anyone would want that...

If needed I can move these to a install_prerquisites.sh script or something, I just like having to run one script to get what I need done.

tools/deploy-with-olm.sh
renan-campos: I checked the documentation for `yq`, but couldn't find any functionality that does field replaement as is being done here. Do you have any examples of this?

config/manager/manager.yaml
renan-campos: Good point, I'll have the deploy script add the values to the fields.

tools/deploy-with-olm.sh
nb-ohad: Ok, I was not clear enough. 
My question is not why are you creating the resources, it meant to be why are you trying to set meaningful values to the fields inside the resources.  

Namespace can be specified as part fo the creation op, while the values of the other fields are irrelevant

renan-campos: The script has the _ability_ to set meaningful values to those resources if specified, but the default behavior is a default value. If pd and dms configuration isn't relevant to the run, the user of the script can just omit that configuration; the script will deploy all that is needed.

nb-ohad: So let's make the default values default, as in empty

tools/deploy-with-olm.sh
nb-ohad: See the following SO answer https://stackoverflow.com/questions/66446495/how-to-patch-multi-document-yaml-file-on-condition-using-yq

You can do it via the `|=` op or use kustomize which we already have in our repo

.gitignore
```diff
@@ -10,6 +10,8 @@ ocs-osd-deployer
 
 # Generated bundle and manifests
 bundle
+bundle.Dockerfile
```
renan-campos, line 13: Like `bundle/`, operator-sdk generates `bundle.Dockerfile`. I therefore .gitignore it.

config/metadata/dependencies.yaml
```diff
@@ -6,4 +6,4 @@ dependencies:
 - type: olm.package
   value:
     packageName: ocs-operator
-    version: ">4.7.0 <4.8.0"
+    version: ">=4.7.0 <4.8.0"
```
renan-campos, line 9: The catalogsource that deploy-with-olm.sh deploys only has ocs-operator 4.7.0. I added this version to dependencies.yaml so that the operator will install with this catalogsource.

Makefile
```diff
@@ -10,7 +10,10 @@ endif
 ifneq ($(origin DEFAULT_CHANNEL), undefined)
 BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
 endif
-BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
+ifneq ($(origin bundle_dir), undefined)
+BUNDLE_DIR = --output-dir=$(bundle_dir)
+endif
+BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL) $(BUNDLE_DIR)
```
nb-ohad, line 16: Please use CAPITAL_LETTERS for all vars names, also touching the BUNDLE_METADATA_OPTS as it is a generated code and as such might change between operator-sdk's versions

DIverging from the original generated code might introduce complexity when we will have to follow the operator-sdk upgrade path

```suggestion
BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL) 

ifneq ($(origin OUTPUT_DIR), undefined)
BUNDLE_OUTPUT_DIR = --output-dir=$(OUTPUT_DIR)
endif
BUNDLE_FLAGS ?= $(BUNDLE_OUTPUT_DIR)
```

In the above suggestion, I have introduced BUNDLE_FLAGS which you will have to append to the end of the proper operator sdk command immediately after BUNDLE_METADATA_OPTS

tools/deploy-with-olm.sh
```diff
@@ -39,21 +39,47 @@ DEPLOYER_IMAGE=${DEPLOYER_IMAGE:-${IMAGE_REPO}/${DEPLOYER_NAME}:${DEPLOYER_VERSI
...
+
 # In install olm if needed
 blue "Checking for existing olm"
-${K8S_CLIENT} get namespace olm >/dev/null 2>&1
-if [ $? -eq 0 ]; then
+if ${K8S_CLIENT} get namespace olm >/dev/null 2>&1
+then
+    blue "olm found"
+elif ${K8S_CLIENT} get namespace openshift-operator-lifecycle-manager &> /dev/null
+then
```
nb-ohad, line 82: Please install OLM before deploying anything on the cluster. 
The cluster shouldn't be touched if OLM has failed to be deployed

nb-ohad (changes requested):

.gitignore
nb-ohad: ack

config/metadata/dependencies.yaml
nb-ohad: As of now, this is ok, but we will need to decouple the catalog source from our deploy script as it currently hard coded against my private repo

tools/deploy-with-olm.sh
```diff
@@ -39,32 +39,52 @@ DEPLOYER_IMAGE=${DEPLOYER_IMAGE:-${IMAGE_REPO}/${DEPLOYER_NAME}:${DEPLOYER_VERSI
 # Deploy target
 TARGET_NAMESPACE=${TARGET_NAMESPACE:-openshift-storage}
 
+# CSV environment variables
+ADDON_NAME=${ADDON_NAME:-ocs-converged}
+SOP_ENDPOINT=${SOP_ENDPOINT:-""}
```
nb-ohad, line 44: This is not needed, undefined SOP_ENDPOINT will resolve to an empty string

tools/deploy-with-olm.sh
```diff
@@ -39,32 +39,52 @@ DEPLOYER_IMAGE=${DEPLOYER_IMAGE:-${IMAGE_REPO}/${DEPLOYER_NAME}:${DEPLOYER_VERSI
...
+BUNDLE_FILE=bundle/manifests/ocs-osd-deployer.clusterserviceversion.yaml
+
```
nb-ohad, line 46: Remove the whitespace

nb-ohad (changes requested):

nb-ohad (approved):

PR #68, opened 2021-07-07
Enable MCG in the OSD deployer

ezio-auditore:
MCG will only be enabled if the addon params contain enable-mcg(bool) true
Signed-off-by: Kaustav Majumder kmajumder@redhat.com

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -63,6 +63,7 @@ const (
 	alertmanagerConfigName                 = "managed-ocs-alertmanager-config"
 	dmsRuleName                            = "dms-monitor-rule"
 	storageClassSizeKey                    = "size"
+	enableMCGKey                           = "enable-mcg"
 	deviceSetName                          = "default"
 	storageClassRbdName                    = "ocs-storagecluster-ceph-rbd"
 	storageClassCephFSName                 = "ocs-storagecluster-cephfs"
@@ -565,9 +566,14 @@ func (r *ManagedOCSReconciler) updateStorageClusterFromAddonParamsSecret(sc *ocs
 		return fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
 	}
 	addonParams := addonParamSecret.Data
-
 	sizeAsString := string(addonParams[storageClassSizeKey])
-	r.Log.Info("Requested add-on settings", storageClassSizeKey, sizeAsString)
+	enableMCGRaw, exists := addonParams[enableMCGKey]
+	// Setting hardcoded value here to force no MCG deployment
+	enableMCGAsString := "false"
+	if exists {
+		enableMCGAsString = string(enableMCGRaw)
+	}
+	r.Log.Info("Requested add-on settings", storageClassSizeKey, sizeAsString, enableMCGKey, enableMCGAsString)
 	desiredDeviceSetCount, err := strconv.Atoi(sizeAsString)
 	if err != nil {
 		return fmt.Errorf("Invalid storage cluster size value: %v", sizeAsString)
@@ -602,6 +608,17 @@ func (r *ManagedOCSReconciler) updateStorageClusterFromAddonParamsSecret(sc *ocs
 	} else {
 		r.Log.V(-1).Info("Requested storage device set count will result in downscaling, which is not supported. Skipping")
 		ds.Count = currDeviceSetCount
+	}
+	// Check and enable MCG in Storage Cluster spec
+	mcgEnable, err := strconv.ParseBool(enableMCGAsString)
+	if err != nil {
+		return fmt.Errorf("Invalid Enable MCG value: %v", enableMCGAsString)
+	}
+	if mcgEnable {
+		r.Log.Info("Enabling Multi Cloud Gateway")
+		sc.Spec.MultiCloudGateway.ReconcileStrategy = "manage"
+	} else if sc.Spec.MultiCloudGateway.ReconcileStrategy == "manage" {
+		r.Log.V(-1).Info("Trying to disable Multi Cloud Gateway, Invalid operation")
 	}
 
 	return nil
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -434,11 +434,95 @@ var _ = Describe("ManagedOCS controller", func() {
 				Expect(configMap.Data["test-key"]).Should(Equal("test-value"))
 			})
 		})
+		When("there is no enable-mcg field in the add-on parameters secret", func() {
+			It("should not create storagecluster", func() {
+				Skip("Skipping this test till Nooba MCG is integrated in deployer")
+				secret := addonParamsSecretTemplate.DeepCopy()
+				secret.Data["size"] = []byte("1")
+				Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+				// Ensure, over a period of time, that the resource is not created
+				utils.EnsureNoResource(k8sClient, ctx, scTemplate.DeepCopy(), timeout, interval)
+			})
+		})
+		When("there is an invalid enable-mcg value in the add-on parameters secret", func() {
+			It("should not create storagecluster", func() {
+				// Create a invalid enable-mcg parameter value
+				secret := addonParamsSecretTemplate.DeepCopy()
+				secret.Data["size"] = []byte("1")
+				secret.Data["enable-mcg"] = []byte("NO")
+				Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+
+				// Ensure, over a period of time, that the resource is not created
+				utils.EnsureNoResource(k8sClient, ctx, scTemplate.DeepCopy(), timeout, interval)
+			})
+		})
+		When("there is a valid enable-mcg value in the addon-on parameters secret which is false", func() {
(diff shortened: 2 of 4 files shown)


controllers/managedocs_controller.go
```diff
@@ -567,11 +568,22 @@ func (r *ManagedOCSReconciler) updateStorageClusterFromAddonParamsSecret(sc *ocs
...
+	if enableMcgAsString != "" {
+		desiredMcgEnable, err := strconv.ParseBool(enableMcgAsString)
+		if err != nil {
+			return fmt.Errorf("Invalid Enable MCG value: %v", enableMcgAsString)
+		}
+		if desiredMcgEnable {
+			r.Log.Info("Enabling MCG")
+			sc.Spec.MultiCloudGateway.ReconcileStrategy = "manage"
+		}
+	}
```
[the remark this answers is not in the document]
ezio-auditore: Moving this section to end of the function causes 18 unit tests to fail since none of them are  expecting the new parameter
`enable-mcg`. I feel we can do two things
1. fix all the failing tests to include a dummy value for the new parameter 
2. Move this to a new reconcile function and make it execute just after` updateStorageClusterFromAddonParamsSecret()` is called, so the new addon won't disrupt existing tests.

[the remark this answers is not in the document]
nb-ohad: I believe it happens not because of order, but because the parameter is a non optional one (as opposed to the original code where you have introduced it as optional)

And yes, when we introduce non-optional parameters we need to make changes to both unit-tests and makefile (to make sure `make run` is working correctly) because failing on a missing parameter is the desired and expected behavior

controllers/managedocs_controller.go
```diff
@@ -567,7 +568,9 @@ func (r *ManagedOCSReconciler) updateStorageClusterFromAddonParamsSecret(sc *ocs
 	addonParams := addonParamSecret.Data
 
 	sizeAsString := string(addonParams[storageClassSizeKey])
-	r.Log.Info("Requested add-on settings", storageClassSizeKey, sizeAsString)
+	// Setting hardcoded value here to force no MCG deployment
+	enableMCGAsString := "false"
```
nb-ohad, line 572:
```suggestion
	// Setting hardcoded value here to force no MCG deployment
	// enableMCGAsString := string(addonParams[enableMCGKey])
	enableMCGAsString := "false"
```

controllers/managedocs_controller_test.go
```diff
@@ -389,6 +389,25 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					scTemplate.DeepCopy(),
+					promTemplate.DeepCopy(),
+					amTemplate.DeepCopy(),
+				}
+				utils.EnsureNoResources(k8sClient, ctx, resList, timeout, interval)
+
+				// Remove the secret for future cases
+				Expect(k8sClient.Delete(ctx, secret)).Should(Succeed())
+			})
+		})
```
nb-ohad, line 410: After hardcoding the `enable-mcg` param to false, this test should fail.

controllers/managedocs_controller_test.go
```diff
@@ -408,6 +427,26 @@ var _ = Describe("ManagedOCS controller", func() {
...
```
nb-ohad, line 449: Same comment as the last one

nb-ohad (changes requested):
A couple of general comments: 
- Missing a test case for validation of enabling of mcg (verifying changes to `storagecluster`) 
- Missing a test case for the prevention of disabling MCG after it has been enabled

renan-campos:
LGTM

controllers/managedocs_controller.go
ezio-auditore: Applied recommended changes

controllers/managedocs_controller_test.go
ezio-auditore: modified the test in the next patch

controllers/managedocs_controller_test.go
ezio-auditore: modified the test in the next patch

ezio-auditore:
> A couple of general comments:
> 
> * Missing a test case for validation of enabling of mcg (verifying changes to `storagecluster`)
added in next patch
> * Missing a test case for the prevention of disabling MCG after it has been enabled
Unable to add any relevant test since MCG cannot be enabled at all since the value is hardcoded. The new tests I have added confirms that. Once we remove the hardcoded `enableMCGAsString := "false"` writing this test should work.

nb-ohad:
@ezio-auditore Can we write the test, verify it, then hide it behind an hardcoded `if` or a comment?

ezio-auditore:
> @ezio-auditore Can we write the test, verify it, then hide it behind an hardcoded `if` or a comment?

I have added `Skip` to this particular test till MCG is properly integrated into Deployer.

controllers/managedocs_controller_test.go
```diff
@@ -434,11 +434,96 @@ var _ = Describe("ManagedOCS controller", func() {
 				Expect(configMap.Data["test-key"]).Should(Equal("test-value"))
 			})
 		})
+		When("there is no enable-mcg field in the add-on parameters secret", func() {
+			It("should not create reconciled resources", func() {
```
nb-ohad, line 438: As you only check for storagecluster, it is best to explicitly specify it:
```suggestion
			It("should not create storagecluster", func() {
```

controllers/managedocs_controller_test.go
```diff
@@ -434,11 +434,96 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				Skip("Skipping this test till Nooba MCG is integrated in deployer")
+				secret := addonParamsSecretTemplate.DeepCopy()
+				secret.Data["size"] = []byte("1")
+				Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+				// Ensure, over a period of time, that the resource is not created
+				utils.EnsureNoResource(k8sClient, ctx, scTemplate.DeepCopy(), timeout, interval)
+			})
+		})
+		When("there is an invalid enable-mcg value in the add-on parameters secret", func() {
+			It("should not create reconciled resources", func() {
```
nb-ohad, line 448: Same as last comment

controllers/managedocs_controller_test.go
```diff
@@ -434,11 +434,96 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				secret.Data["size"] = []byte("1")
+				secret.Data["enable-mcg"] = []byte("NO")
+				Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+
+				// Ensure, over a period of time, that the resource is not created
+				utils.EnsureNoResource(k8sClient, ctx, scTemplate.DeepCopy(), timeout, interval)
+			})
+		})
+		When("there is a valid enable-mcg value in the addon-on parameters secret which is false", func() {
+			It("should not change the storage cluster spec", func() {
```
nb-ohad, line 460: Again this is not specific enough
```suggestion
			It("Should change storagecluster's MCG reconcile strategy to manage ", func() {
```

controllers/managedocs_controller_test.go
```diff
@@ -434,11 +434,96 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					err := k8sClient.Get(ctx, utils.GetResourceKey(sc), sc)
+					if err != nil {
+						return true
+					}
+					return sc.Spec.MultiCloudGateway.ReconcileStrategy == "manage"
+				}, timeout, interval).Should(BeFalse())
+			})
+		})
+		When("there is a valid enable-mcg value in the addon-on parameters secret which is true", func() {
+			It("should change the storage cluster spec to enable MCG", func() {
```
nb-ohad, line 477: See last comment

controllers/managedocs_controller_test.go
```diff
@@ -434,11 +434,96 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					err := k8sClient.Get(ctx, utils.GetResourceKey(sc), sc)
+					if err != nil {
+						return false
+					}
+					return sc.Spec.MultiCloudGateway.ReconcileStrategy == "manage"
+				}, timeout, interval).Should(BeTrue())
+			})
+		})
+		When("MCG is already enabled and enable-mcg value in addon-on parameter secret is false", func() {
+			It("should not disable MCG in storage cluster spec", func() {
```
nb-ohad, line 494: see last comment

controllers/managedocs_controller_test.go
```diff
@@ -434,11 +434,96 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+				// MCG spec should not get updated
+				Eventually(func() bool {
+					sc := scTemplate.DeepCopy()
+					err := k8sClient.Get(ctx, utils.GetResourceKey(sc), sc)
+					if err != nil {
+						return false
+					}
+					return sc.Spec.MultiCloudGateway.ReconcileStrategy == "manage" ||
+						sc.Spec.MultiCloudGateway.ReconcileStrategy == ""
```
nb-ohad, line 506: `""` is not an acceptable value in this case, we need to look specifically for "manage"

nb-ohad (changes requested):

nb-ohad (approved):

PR #70, opened 2021-07-21
Integrate SendGrid to support customer email notification

kesavanvt:
The customer email ids are fecthed during the addon install and relevant storage alerts are sent to the target address via alertmanager routes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ export_env_vars:
 export NAMESPACE = openshift-storage
 export ADDON_NAME = ocs-converged
 export SOP_ENDPOINT = https://red-hat-storage.github.io/ocs-sop/sop/OSD/{{ .GroupLabels.alertname }}.html
+export ALERT_SMTP_FROM_ADDR = noreply-test@test.com
 
 # Run tests
 ENVTEST_ASSETS_DIR = $(shell pwd)/testbin
@@ -56,9 +57,11 @@ readinessServer: fmt vet
 # Run against the configured Kubernetes cluster in ~/.kube/config
 run: generate fmt vet manifests export_env_vars
 	kubectl create namespace ${NAMESPACE} --dry-run=client -o yaml | kubectl apply -f -
-	kubectl create secret generic addon-${ADDON_NAME}-parameters -n ${NAMESPACE} --from-literal=size=1 --from-literal=enable-mcg=false --dry-run=client -oyaml | kubectl apply -f -
+	kubectl create secret generic addon-${ADDON_NAME}-parameters -n ${NAMESPACE} --from-literal size=1 --from-literal enable-mcg=false --dry-run=client -oyaml | kubectl apply -f -
 	kubectl create secret generic ${ADDON_NAME}-pagerduty -n ${NAMESPACE} --from-literal PAGERDUTY_KEY="test-key" --dry-run=client -oyaml | kubectl apply -f -
 	kubectl create secret generic ${ADDON_NAME}-deadmanssnitch -n ${NAMESPACE} --from-literal SNITCH_URL="https://test-url" --dry-run=client -oyaml | kubectl apply -f -
+	kubectl create secret generic ${ADDON_NAME}-smtp -n ${NAMESPACE} --from-literal host="smtp.sendgrid.net" --from-literal password="test-key" --from-literal port="587" \
+	--from-literal username="apikey" --dry-run=client -oyaml | kubectl apply -f -
 	kubectl create configmap rook-ceph-operator-config -n ${NAMESPACE} --dry-run=client -oyaml | kubectl apply -f -
 	echo -e "apiVersion: operators.coreos.com/v1alpha1" \
 	      "\nkind: ClusterServiceVersion" \
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -43,6 +43,7 @@ spec:
               fieldPath: metadata.namespace
         - name: ADDON_NAME
         - name: SOP_ENDPOINT
+        - name: ALERT_SMTP_FROM_ADDR
       - name: readiness-server
         command:
         - /readinessServer
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -64,6 +64,7 @@ const (
 	dmsRuleName                            = "dms-monitor-rule"
 	storageClassSizeKey                    = "size"
 	enableMCGKey                           = "enable-mcg"
+	notificationEmailKeyPrefix             = "notification-email"
 	deviceSetName                          = "default"
 	storageClassRbdName                    = "ocs-storagecluster-ceph-rbd"
 	storageClassCephFSName                 = "ocs-storagecluster-cephfs"
@@ -92,7 +93,9 @@ type ManagedOCSReconciler struct {
 	DeployerSubscriptionName     string
 	PagerdutySecretName          string
 	DeadMansSnitchSecretName     string
+	SMTPSecretName               string
 	SOPEndpoint                  string
+	AlertSMTPFrom                string
 
 	ctx                                context.Context
 	managedOCS                         *v1.ManagedOCS
@@ -100,8 +103,10 @@ type ManagedOCSReconciler struct {
 	prometheus                         *promv1.Prometheus
 	dmsRule                            *promv1.PrometheusRule
 	alertmanager                       *promv1.Alertmanager
+	addonParamSecret                   *corev1.Secret
 	pagerdutySecret                    *corev1.Secret
 	deadMansSnitchSecret               *corev1.Secret
+	smtpSecret                         *corev1.Secret
 	alertmanagerConfig                 *promv1a1.AlertmanagerConfig
 	k8sMetricsServiceMonitor           *promv1.ServiceMonitor
 	k8sMetricsServiceMonitorAuthSecret *corev1.Secret
@@ -138,7 +143,8 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				name := meta.GetName()
 				return name == r.AddonParamSecretName ||
 					name == r.PagerdutySecretName ||
-					name == r.DeadMansSnitchSecretName
+					name == r.DeadMansSnitchSecretName ||
+					name == r.SMTPSecretName
 			},
 		),
 	)
@@ -322,9 +328,17 @@ func (r *ManagedOCSReconciler) initReconciler(req ctrl.Request) {
 	r.alertmanager.Name = alertmanagerName
 	r.alertmanager.Namespace = r.namespace
 
+	r.addonParamSecret = &corev1.Secret{}
+	r.addonParamSecret.Name = r.AddonParamSecretName
+	r.addonParamSecret.Namespace = r.namespace
+
 	r.pagerdutySecret = &corev1.Secret{}
 	r.pagerdutySecret.Name = r.PagerdutySecretName
 	r.pagerdutySecret.Namespace = r.namespace
+
+	r.smtpSecret = &corev1.Secret{}
+	r.smtpSecret.Name = r.SMTPSecretName
+	r.smtpSecret.Namespace = r.namespace
 
 	r.deadMansSnitchSecret = &corev1.Secret{}
 	r.deadMansSnitchSecret.Name = r.DeadMansSnitchSecretName
@@ -383,6 +397,10 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 		r.reconcileStrategy = v1.ReconcileStrategyStrict
 		if strings.EqualFold(string(r.managedOCS.Spec.ReconcileStrategy), string(v1.ReconcileStrategyNone)) {
 			r.reconcileStrategy = v1.ReconcileStrategyNone
+		}
+
+		if err := r.get(r.addonParamSecret); err != nil {
+			return ctrl.Result{}, fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
 		}
 
 		// Reconcile the different resources
@@ -554,18 +572,8 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
 }
 
 func (r *ManagedOCSReconciler) updateStorageClusterFromAddonParamsSecret(sc *ocsv1.StorageCluster) error {
-	// The addon param secret will contain the capacity of the cluster in Ti
-	// size = 1,  creates a cluster of 1 Ti capacity
-	// size = 2,  creates a cluster of 2 Ti capacity etc
+	addonParams := r.addonParamSecret.Data
 
-	addonParamSecret := &corev1.Secret{}
-	addonParamSecret.Name = r.AddonParamSecretName
-	addonParamSecret.Namespace = r.namespace
-	if err := r.get(addonParamSecret); err != nil {
-		// Do not create the StorageCluster if the we fail to get the addon param secret
-		return fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
-	}
-	addonParams := addonParamSecret.Data
 	sizeAsString := string(addonParams[storageClassSizeKey])
 	enableMCGRaw, exists := addonParams[enableMCGKey]
 	// Setting hardcoded value here to force no MCG deployment
@@ -747,6 +755,40 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 			return fmt.Errorf("DeadMan's Snitch secret does not contain a SNITCH_URL entry")
 		}
 
+		alertingAddressList := []string{}
+		i := 0
+		for {
+			alertingAddress, found := r.addonParamSecret.Data[notificationEmailKeyPrefix+fmt.Sprintf("-%v", i)]
+			i++
+			if found {
+				alertingAddressList = append(alertingAddressList, string(alertingAddress))
+			} else {
+				break
+			}
+		}
+
+		smtpSecretData := map[string][]byte{}
+		if err := r.get(r.smtpSecret); err != nil {
+			return fmt.Errorf("Unable to get SMTP secret: %v", err)
+		}
+		smtpSecretData = r.smtpSecret.Data
+		smtpHost := string(smtpSecretData["host"])
+		if smtpHost == "" {
+			return fmt.Errorf("smtp secret does not contain a host entry")
+		}
+		smtpPort := string(smtpSecretData["port"])
+		if smtpPort == "" {
+			return fmt.Errorf("smtp secret does not contain a port entry")
+		}
+		smtpUsername := string(smtpSecretData["username"])
+		if smtpUsername == "" {
+			return fmt.Errorf("smtp secret does not contain a username entry")
+		}
+		smtpPassword := string(smtpSecretData["password"])
+		if smtpPassword == "" {
+			return fmt.Errorf("smtp secret does not contain a password entry")
+		}
+
 		desired := templates.AlertmanagerConfigTemplate.DeepCopy()
 		for i := range desired.Spec.Receivers {
 			receiver := &desired.Spec.Receivers[i]
@@ -758,6 +800,18 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 				receiver.PagerDutyConfigs[0].Details[0].Value = r.SOPEndpoint
 			case "DeadMansSnitch":
 				receiver.WebhookConfigs[0].URL = &dmsURL
+			case "SendGrid":
+				if len(alertingAddressList) > 0 {
+					receiver.EmailConfigs[0].Smarthost = fmt.Sprintf("%s:%s", smtpHost, smtpPort)
+					receiver.EmailConfigs[0].AuthUsername = smtpUsername
+					receiver.EmailConfigs[0].AuthPassword.LocalObjectReference.Name = r.SMTPSecretName
+					receiver.EmailConfigs[0].AuthPassword.Key = "password"
+					receiver.EmailConfigs[0].From = r.AlertSMTPFrom
+					receiver.EmailConfigs[0].To = strings.Join(alertingAddressList, ", ")
+				} else {
+					r.Log.V(-1).Info("Customer Email for alert notification is not provided")
+					receiver.EmailConfigs = []promv1a1.EmailConfig{}
+				}
 			}
 		}
 		r.alertmanagerConfig.Spec = desired.Spec
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -137,6 +137,13 @@ var _ = Describe("ManagedOCS controller", func() {
 		},
 		Data: map[string][]byte{},
 	}
+	smtpSecretTemplate := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      testSMTPSecretName,
+			Namespace: testPrimaryNamespace,
+		},
+		Data: map[string][]byte{},
+	}
 	amConfigTemplate := promv1a1.AlertmanagerConfig{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      alertmanagerConfigName,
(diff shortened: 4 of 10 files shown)


renan-campos:
Hey Kesavan, when you get a chance could you add code to create these new secrets in tools/deploy-with-olm.sh?

ezio-auditore:
> Hey Kesavan, when you get a chance could you add code to create these new secrets in tools/deploy-with-olm.sh?
**Also the Makefile under the `run` rule**

controllers/managedocs_controller.go
```diff
@@ -730,6 +758,15 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
...
 			case "DeadMansSnitch":
 				receiver.WebhookConfigs[0].URL = &dmsURL
+			case "SendGrid":
+				if notificationEmailValue != "" {
+					receiver.EmailConfigs[0].Smarthost = fmt.Sprintf("%s:%s", smtpSecretData["host"], smtpSecretData["port"])
+					receiver.EmailConfigs[0].AuthUsername = string(smtpSecretData["username"])
+					receiver.EmailConfigs[0].AuthPassword.LocalObjectReference.Name = r.SMTPSecretName
+					receiver.EmailConfigs[0].AuthPassword.Key = "password"
+					receiver.EmailConfigs[0].From = r.AlertSMTPFrom
+				}
```
nb-ohad, line 768: The above code assumes that you can opt in but not opt-out. 
I think it will be best to always configure the secret and only empty/fill the email list to accomplish both opt-in and opt-out functionality.

main.go
```diff
@@ -174,6 +177,12 @@ func readEnvVars() (map[string]string, error) {
 	}
 	envVars[sopEndpointEnvVarName] = val
 
+	val, found = os.LookupEnv(alertSMTPFromEnvVarName)
+	if !found {
+		return nil, fmt.Errorf("%s environment variable must be set", sopEndpointEnvVarName)
+	}
+	envVars[alertSMTPFromEnvVarName] = val
+
```
nb-ohad, line 185: As we see that a pattern emerges here, I would suggest a refactor as following:
```
	envVars := map[string]string{
		namespaceEnvVarName: "",
		addonNameEnvVarName: "",
		sopEndpointEnvVarName: "",
		alertSMTPFromEnvVarName: "",
	}
	for envVarName := range envVars {
		val, found := os.LookupEnv(envVarName)
		if !found {
			return nil, fmt.Errorf("%s environment variable must be set", envVarName)
		}
		envVars[envVarName] = val
	}
```

Also, the error message is using the wrong env name, `sopEndpointEnvVarName` and not `alertSMTPFromEnvVarName`

config/manager/manager.yaml
```diff
@@ -43,6 +43,7 @@ spec:
               fieldPath: metadata.namespace
         - name: ADDON_NAME
         - name: SOP_ENDPOINT
+        - name: ALERT_SMTP_FROM
```
nb-ohad, line 46: Can we change from  `ALERT_SMTP_FROM` to `ALERT_SMTP_FROM_ADDR`

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -719,6 +730,23 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
...
+		var customerAlertingEmailAddress string
+		notificationEmailValue := string(addonParams[notificationEmailKey])
+		if notificationEmailValue != "" {
+			emailAddresses := strings.Split(strings.TrimSpace(notificationEmailValue), " ")
+			customerAlertingEmailAddress = strings.Join(emailAddresses, ", ")
+			if err := r.get(r.smtpSecret); err != nil {
+				return fmt.Errorf("Unable to get SMTP secret: %v", err)
+			}
+			smtpSecretData = r.smtpSecret.Data
+		}
```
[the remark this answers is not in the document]
kesavanvt: Assuming the addon paramam secret key would be as follows
```
notification-email1
notification-email2
notification-email3
...
```

templates/alertmanagerconfig.go
```diff
@@ -31,17 +31,43 @@ func convertToApiExtV1JSON(val interface{}) apiextensionsv1.JSON {
...
+
+var smtpRegxRuleMatchers = "CephClusterNearFull" +
+	"|CephClusterCriticallyFull" +
+	"|CephClusterReadOnly" +
+	"|CephOSDCriticallyFull" +
+	"|CephOSDNearFull" +
+	"|PersistentVolumeUsageNearFull" +
+	"|PersistentVolumeUsageCritical"
+
+var pagerdutyRegxRuleMatchers = "critical|warning"
```
nb-ohad, line 44: Why aren't we using exact alerts names here?

templates/alertmanagerconfig.go
```diff
@@ -31,17 +31,43 @@ func convertToApiExtV1JSON(val interface{}) apiextensionsv1.JSON {
...
+					Receiver: "null",
+				},
+				),
 				convertToApiExtV1JSON(promv1a1.Route{
 					GroupBy:        []string{"alertname"},
 					GroupWait:      "30s",
 					GroupInterval:  "5m",
 					RepeatInterval: "12h",
-					Matchers:       []promv1a1.Matcher{{Name: "severity", Value: "critical"}},
+					Matchers:       []promv1a1.Matcher{{Name: "alertname", Value: smtpRegxRuleMatchers, Regex: true}},
```
nb-ohad, line 61: As we are using full alerts names can we use a list of matchers and not a single matcher with a regexp?

templates/alertmanagerconfig.go
```diff
@@ -31,17 +31,43 @@ func convertToApiExtV1JSON(val interface{}) apiextensionsv1.JSON {
...
+					Receiver:       "SendGrid",
+				},
+				),
+				convertToApiExtV1JSON(promv1a1.Route{
+					GroupBy:        []string{"alertname"},
+					GroupWait:      "30s",
+					GroupInterval:  "5m",
+					RepeatInterval: "12h",
+					Matchers:       []promv1a1.Matcher{{Name: "severity", Value: pagerdutyRegxRuleMatchers, Regex: true}},
```
nb-ohad, line 70: This should be a name based matcher, not a severity based one as we have an exact white list of alerts for this receiver

tools/deploy-with-olm.sh
```diff
@@ -42,9 +42,15 @@ TARGET_NAMESPACE=${TARGET_NAMESPACE:-openshift-storage}
 
 # CSV environment variables
 ADDON_NAME=${ADDON_NAME:-ocs-converged}
+ALERT_SMTP_FROM_ADDR=${ALERT_SMTP_FROM_ADDR:-noreply-test@rhmw.io}
```
nb-ohad, line 45: Lets not introduce specific  red hat email addressed into this public repo

tools/deploy-with-olm.sh
```diff
@@ -79,9 +85,10 @@ blue "Creating catalogsource containing needed dependencies on cluster"
 exit_on_err ${K8S_CLIENT} -n ${TARGET_NAMESPACE} apply -f ./tools/yamls/ocs-catalogsource.yaml
 
 blue "Creating secrets needed for the operator to run without error"
-exit_on_err ${K8S_CLIENT} -n ${TARGET_NAMESPACE} create secret generic addon-${ADDON_NAME}-parameters --from-literal=size=${CLUSTER_SIZE} --dry-run=client -o yaml | ${K8S_CLIENT} -n ${TARGET_NAMESPACE} apply -f -
+exit_on_err ${K8S_CLIENT} -n ${TARGET_NAMESPACE} create secret generic addon-${ADDON_NAME}-parameters --from-literal=size=${CLUSTER_SIZE} --from-literal=notification-email-0=${ALERT_EMAIL} --dry-run=client -o yaml | ${K8S_CLIENT} -n ${TARGET_NAMESPACE} apply -f -
```
nb-ohad, line 88: Where does the ALERT_EMAIL value come from? I couldn't find a definition for it. 
Are we assuming empty string if the person running the script does not provide an email in an env var?

controllers/managedocs_controller_test.go
```diff
@@ -436,8 +442,11 @@ var _ = Describe("ManagedOCS controller", func() {
 		})
 		When("there is a valid size in the add-on parameter secret", func() {
 			It("should create reconciled resources", func() {
-				// Create a valid add-on parameters secret
+				// Ensure there is no add-on parameters secret present
 				secret := addonParamsSecretTemplate.DeepCopy()
+				Expect(k8sClient.Delete(ctx, secret)).Should(Succeed())
```
nb-ohad, line 447: This is clean up code for the prev test, so let's move it to the prev test

controllers/managedocs_controller_test.go
```diff
@@ -852,7 +861,41 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+				// Create pagerduty secret with valid key
+				pdSecret := pdSecretTemplate.DeepCopy()
+				pdSecret.Data["PAGERDUTY_KEY"] = []byte("test-key")
+
+				// Create deadman's snitch secret with a valid key
+				dmsSecret := dmsSecretTemplate.DeepCopy()
+				dmsSecret.Data["SNITCH_URL"] = []byte("test-key")
+
+				Expect(k8sClient.Create(ctx, pdSecret)).Should(Succeed())
```
nb-ohad, line 882: move each create after the code that set it up, in this case move this to line 877

controllers/managedocs_controller_test.go
```diff
@@ -852,7 +861,41 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				Expect(k8sClient.Delete(ctx, pdSecret)).Should(Succeed())
+				Expect(k8sClient.Delete(ctx, dmsSecret)).Should(Succeed())
+
+				// Remove notification email for future cases
+				delete(secret.Data, "notification-email-0")
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+
+			})
+		})
+		When("there is a value for PAGERDUTY_KEY in the pagerduty secret, a value for SNITCH_URL in the deadmanssnitch secret and a SMTP secret", func() {
```
nb-ohad, line 898: Each test should isolate one check, this one seems to me like an aggregation of many checks. 
If this is check is supposed to test a situation where all alert manager conditions are met, let's describe it as such. 

Think. about the case where we will introduce many more conditions, we will not want to track all of them in the description for this case. 

```suggestion
		When("All conditions for creating an alertmanager config are met ", func() {
```

controllers/managedocs_controller_test.go
```diff
@@ -862,9 +905,17 @@ var _ = Describe("ManagedOCS controller", func() {
 				dmsSecret := dmsSecretTemplate.DeepCopy()
 				dmsSecret.Data["SNITCH_URL"] = []byte("test-key")
 
+				// Create smtp secret with valid keys
+				smtpSecret := smtpSecretTemplate.DeepCopy()
+				smtpSecret.Data["host"] = []byte("test-host")
+				smtpSecret.Data["port"] = []byte("20")
+				smtpSecret.Data["username"] = []byte("test-key")
+				smtpSecret.Data["password"] = []byte("test-password")
```
nb-ohad, line 913: This and the prev 3-4 tests are about the creation of the alertmanager config under certain circumstances. 
This means that we are setting up a state for all alertmanager config pre-condition on each one of the test cases. 
This includes: 
1. pager duty secret
2. dms secret 
3. smtpSecret secret.

We already encountered a similar situation with uninstall conditions. In that case, we introduced `setupUninstallConditions` to help us manage the setup of uninstallation conditions in a central place. This allowed us to introduce new conditions without the need to adjust each test case for uninstalltion. 

My suggestion is to repeat that pattern here. We can introduce another helper:
```
func setupAlertmanagerConfigConditions(pdSecret bool, dmsSecret bool, smtpSecret bool)
```
and use it in all test cases that test for the existence of an alertmanager config

controllers/managedocs_controller_test.go
```diff
@@ -1021,6 +1072,65 @@ var _ = Describe("ManagedOCS controller", func() {
...
 			})
 		})
+		When("there is a notification email address in the add-on parameter secret and smtp secret", func() {
+			It("should update alertmanager email config with smtp details", func() {
+				secret := addonParamsSecretTemplate.DeepCopy()
+				secretKey := utils.GetResourceKey(secret)
+				Expect(k8sClient.Get(ctx, secretKey, secret)).Should(Succeed())
+
+				// Add notification email to the addon param secret
+				secret.Data["notification-email-0"] = []byte("test@email.com")
```
nb-ohad, line 1082: This is another candidate for param to the `setupAlertmanagerConfigConditions` util I suggested in the last comment. You can even use an array to set up multiple addresses at once.

controllers/managedocs_controller_test.go
```diff
@@ -1021,6 +1072,65 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				Expect(k8sClient.Update(ctx, secret)).Should(Succeed())
+
+				// Wait for alertmanager to get updated with smtp details
+				amconfig := amConfigTemplate.DeepCopy()
+				amconfigKey := utils.GetResourceKey(amconfig)
+				Eventually(func() bool {
+					Expect(k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+					return len(amconfig.Spec.Receivers[3].EmailConfigs) > 0
```
nb-ohad, line 1090: Do not assume receivers location, this can break very easly

controllers/managedocs_controller_test.go
```diff
@@ -1021,6 +1072,65 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				}, timeout, interval).Should(Equal(true))
+
+				Expect(amconfig.Spec.Receivers[3].EmailConfigs[0].To).Should(Equal("test@email.com"))
```
nb-ohad, line 1093: same as last comment

nb-ohad (changes requested):

nb-ohad:
@kesavanvt another general note, 
The current test only covers a single email address. To verify that our code handles multiple email addresses correctly I suggest to add two new test cases:
1. Test case that adds a secondary email address 
2. Test case that remove the second email address

tools/deploy-with-olm.sh
kesavanvt: Updated ALERT_EMAIL to have a default value

templates/alertmanagerconfig.go
kesavanvt: matchers are ANDed and with regex are ORed. Thus alertnames are provided as regex.

templates/alertmanagerconfig.go
kesavanvt: matchers are ANDed and with regex are ORed. Thus alertnames are provided as regex.

templates/alertmanagerconfig.go
kesavanvt: The Alerts that are not relevant to PagerDuty are routed to corresponding services and its not required to specify the full list of alerts that needs to routed to pagerduty. Thus picking the remaining alerts based on severity for pagerduty.

controllers/managedocs_controller_test.go
```diff
@@ -1114,6 +1187,115 @@ var _ = Describe("ManagedOCS controller", func() {
 				}, timeout, interval).Should(Equal(ctrlutils.GetResourceRequirements("ocs-operator")))
 			})
 		})
+		When("there is a notification email address in the add-on parameter secret and smtp secret", func() {
+			It("should update alertmanager email config with smtp details", func() {
+				setupAlertmanagerConfigConditions(true, "test-key", true, "test-url", true, []string{"test@email.com"})
```
nb-ohad, line 1192: As we do not actually care about the email address can we just specify it as a number?
Something like `number of notification email addressed`

controllers/managedocs_controller_test.go
```diff
@@ -1114,6 +1187,115 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				var resIndex int
+				Eventually(func() bool {
+					Expect(k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+					for i, reciever := range amconfig.Spec.Receivers {
+						if reciever.Name == "SendGrid" {
+							resIndex = i
+						}
+					}
+					return len(amconfig.Spec.Receivers[resIndex].EmailConfigs) > 0
+				}, timeout, interval).Should(Equal(true))
```
nb-ohad, line 1206: A couple of comments:
1. This for loop will create an in-memory copy of the receiver.
2. You do not need to wait for the loop to end after finding the proper receiver. 
3. Why are we checking the length of the `EmialConfigs`? this test case is not about the email config it is about the existence of SMTP receiver. The tests below already cover the email configs.

The suggestion below will fix both these issues:
```suggestion
				Eventually(func() bool {
					for i := range amconfig.Spec.Receivers {
						reciever := &amconfig.Spec.Receivers[i]
						if reciever.Name == "SendGrid" {
							return true;
						}
					}
				}, timeout, interval).Should(Equal(true))
```

On top of that this `Eventually` construct is repeating on multiple test cases which makes it a candidate for a test util (in the `testutils/utils.go` file). Something like: 
```func waitForAlertManagerSendGridReceiver(emailAddresses string[]) ```

File: controllers/managedocs_controller_test.go [hunk 19]
```diff
@@ -1114,6 +1187,115 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					amconfig := amConfigTemplate.DeepCopy()
+					amconfigKey := utils.GetResourceKey(amconfig)
+					Expect(k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+					for i, reciever := range amconfig.Spec.Receivers {
+						if reciever.Name == "SendGrid" {
+							resIndex = i
+						}
+					}
+					return amconfig.Spec.Receivers[resIndex].EmailConfigs[0].To
+				}, timeout, interval).Should(Equal("test2@email.com"))
```
nb-ohad, line 1227: See comment above

controllers/managedocs_controller_test.go
```diff
@@ -1114,6 +1187,115 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					amconfig := amConfigTemplate.DeepCopy()
+					amconfigKey := utils.GetResourceKey(amconfig)
+					Expect(k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+					for i, reciever := range amconfig.Spec.Receivers {
+						if reciever.Name == "SendGrid" {
+							resIndex = i
+						}
+					}
+					return amconfig.Spec.Receivers[resIndex].EmailConfigs[0].To
+				}, timeout, interval).Should(Equal("test2@email.com, test3@email.com"))
```
nb-ohad, line 1246: See comment above

controllers/managedocs_controller_test.go
```diff
@@ -1114,6 +1187,115 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					amconfig := amConfigTemplate.DeepCopy()
+					amconfigKey := utils.GetResourceKey(amconfig)
+					Expect(k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+					for i, reciever := range amconfig.Spec.Receivers {
+						if reciever.Name == "SendGrid" {
+							resIndex = i
+						}
+					}
+					return amconfig.Spec.Receivers[resIndex].EmailConfigs[0].To
+				}, timeout, interval).Should(Equal("test2@email.com"))
```
nb-ohad, line 1271: See comment above

controllers/managedocs_controller_test.go
```diff
@@ -1114,6 +1187,115 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					amconfig := amConfigTemplate.DeepCopy()
+					amconfigKey := utils.GetResourceKey(amconfig)
+					Expect(k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+					for i, reciever := range amconfig.Spec.Receivers {
+						if reciever.Name == "SendGrid" {
+							resIndex = i
+						}
+					}
+					return len(amconfig.Spec.Receivers[resIndex].EmailConfigs) == 0
+				}, timeout, interval).Should(Equal(true))
```
nb-ohad, line 1296: See comment above

controllers/managedocs_controller_test.go
```diff
@@ -1114,6 +1187,115 @@ var _ = Describe("ManagedOCS controller", func() {
...
+				Eventually(func() bool {
+					amconfig := amConfigTemplate.DeepCopy()
+					amconfigKey := utils.GetResourceKey(amconfig)
+					Expect(k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+					for i, reciever := range amconfig.Spec.Receivers {
+						if reciever.Name == "SendGrid" {
+							resIndex = i
+						}
+					}
+					return len(amconfig.Spec.Receivers[resIndex].EmailConfigs) == 0
```
nb-ohad, line 1295: I was under the impression that the entire receiver will be removed if the last email is removed.

nb-ohad (changes requested):

tools/deploy-with-olm.sh
```diff
@@ -42,10 +42,17 @@ TARGET_NAMESPACE=${TARGET_NAMESPACE:-openshift-storage}
 
 # CSV environment variables
 ADDON_NAME=${ADDON_NAME:-ocs-converged}
+ALERT_SMTP_FROM_ADDR=${ALERT_SMTP_FROM_ADDR:-noreply-test@email.com}
 BUNDLE_FILE=${OUTPUT_DIR}/manifests/ocs-osd-deployer.clusterserviceversion.yaml
 CLUSTER_SIZE=${CLUSTER_SIZE:-1}
 ENABLE_MCG_FLAG=${ENABLE_MCG_FLAG:-false}
 
+#SMTP config variables
+[ -z "$NOTIFICATION_EMAIL" ] || NOTIFICATION_EMAIL_OPT="--from-literal=notification-email-0=${NOTIFICATION_EMAIL}"
```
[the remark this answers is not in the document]
kesavanvt: No, This is one way of passing the parameters. Keeping as it is since this format is followed in the above file

controllers/managedocs_controller_test.go
```diff
@@ -404,12 +489,11 @@ var _ = Describe("ManagedOCS controller", func() {
 				}
 				utils.EnsureNoResources(k8sClient, ctx, resList, timeout, interval)
 
-				// Remove the secret for future cases
-				Expect(k8sClient.Delete(ctx, secret)).Should(Succeed())
 			})
 		})
 		When("there is a rook-ceph-operator-config ConfigMap", func() {
 			It("should ensure there are RBD CSI resource limits", func() {
+				// This test assumes a addon param secret is already created
```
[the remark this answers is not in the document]
kesavanvt: According to the order of reconciliation,  rook-ceph-operator-config reconciliation is executed before the storagecluster reconciliation, this doesn't allow to move the above test after the addon param success test.  Thus moved the above test before to addon params tests. Also, The above test expects to have addon param.

tools/deploy-with-olm.sh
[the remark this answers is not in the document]
nb-ohad: It does not make sense as all the other `--from-literal` arguments do not include the "=". Let stick to one format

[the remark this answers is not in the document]
kesavanvt: Done, modified to maintain uniformity

controllers/managedocs_controller_test.go
```diff
@@ -353,7 +360,138 @@ var _ = Describe("ManagedOCS controller", func() {
...
+			} else {
+				Expect(k8sClient.Create(ctx, pdSecret)).Should(Succeed())
+			}
+		} else if pdSecretExists {
+			Expect(k8sClient.Delete(ctx, pdSecret)).Should(Succeed())
+		}
+	}
+
 	Context("reconcile()", func() {
+		When("there is a rook-ceph-operator-config ConfigMap", func() {
```
nb-ohad, line 461: Let's move this test after the addon params test, this way we could assume that the addon params are already in a correct state.

In order to allow the addon params to test creation of resources, we will need to make sure that a `rook-ceph-operator-config` config map exists before the first test

nb-ohad (changes requested):

Makefile
```diff
@@ -37,7 +37,7 @@ export_env_vars:
 export NAMESPACE = openshift-storage
 export ADDON_NAME = ocs-converged
 export SOP_ENDPOINT = https://red-hat-storage.github.io/ocs-sop/sop/OSD/{{ .GroupLabels.alertname }}.html
-
+export ALERT_SMTP_FROM_ADDR = noreply-test@test.com
```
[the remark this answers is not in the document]
kesavanvt: This is for setting the environment variable `ALERT_SMTP_FROM_ADDR` and its required for `make run`. Keeping as it is.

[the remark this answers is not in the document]
nb-ohad: Haa, the coin have dropped! This is an env that the operator need defined in order to execute.

testutils/utils.go
```diff
@@ -79,3 +81,29 @@ func ToJsonOrDie(value interface{}) []byte {
...
+			if reciever.Name == "SendGrid" {
+				resIndex = i
+				break
+			}
+		}
+		if len(amconfig.Spec.Receivers[resIndex].EmailConfigs) > 0 {
+			return amconfig.Spec.Receivers[resIndex].EmailConfigs[0].To
+		}
+		return ""
+	}, timeout, interval).Should(Equal(strings.Join(emailadresses, ", ")))
```
nb-ohad, line 108: You have a couple of issues with this util function:
1. You do not protect against a receiver list that does not contain the desired reciver.  
2. You find the receiver ref but then you just throw it and dave the index just to reindex into the Receivers array.


```suggestion
	EventuallyWithOffset(1, func() string {
		amconfig := &promv1a1.AlertmanagerConfig{}
		ExpectWithOffset(2, k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
		var reciever *RecieverType = nullptr;
		for i := range amconfig.Spec.Receivers {
			r := &amconfig.Spec.Receivers[i]
			if r.Name == "SendGrid" {
				reciever = r
				break
			}
		}

		if reciver != nullptr && len(amconfig.Spec.Receivers[resIndex].EmailConfigs) > 0 {
			return amconfig.Spec.Receivers[resIndex].EmailConfigs[0].To
		}
		return ""
	}, timeout, interval).Should(Equal(strings.Join(emailadresses, ", ")))
```

or even better and much simpler:
```suggestion
	EventuallyWithOffset(1, func() string {
		amconfig := &promv1a1.AlertmanagerConfig{}
		ExpectWithOffset(2, k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
		for i := range amconfig.Spec.Receivers {
			reciever := &amconfig.Spec.Receivers[i]
			if reciever.Name == "SendGrid" {
				if len(amconfig.Spec.Receivers[resIndex].EmailConfigs) > 0 {
					return amconfig.Spec.Receivers[resIndex].EmailConfigs[0].To
				} else {
				        return ""
				}
			}
		}
		return ""
	}, timeout, interval).Should(Equal(strings.Join(emailadresses, ", ")))
```

nb-ohad (changes requested):

testutils/utils.go
```diff
@@ -79,3 +81,29 @@ func ToJsonOrDie(value interface{}) []byte {
 		panic(err)
 	}
 }
+
+func WaitForAlertManagerSendGridReceiver(
```
nb-ohad, line 85: The name does not reflect what the util does. I will suggest naming it as: `WaitForAlertMangerSMTPReciverEmailConfigToUpdate`

testutils/utils.go
```diff
@@ -79,3 +81,29 @@ func ToJsonOrDie(value interface{}) []byte {
...
+	timeout time.Duration,
+	interval time.Duration,
+) {
+	var resIndex int
+	EventuallyWithOffset(1, func() string {
+		amconfig := &promv1a1.AlertmanagerConfig{}
+		ExpectWithOffset(2, k8sClient.Get(ctx, amconfigKey, amconfig)).Should(Succeed())
+		for i := range amconfig.Spec.Receivers {
+			reciever := &amconfig.Spec.Receivers[i]
+			if reciever.Name == "SendGrid" {
```
nb-ohad, line 99: Can we pass the receiver's name from the outside. this is util and as such should not assume a specific name.
This goes hand in hand with the comment above about the util name

nb-ohad (changes requested):

nb-ohad (approved):

PR #71, opened 2021-07-22
Updating prometheus rule for DMS.

renan-campos:
With the recent changes to Alertmanager's configuration, the DeadMan's
Snitch alert was filtered out and never sent to to the DMS endpoint.
This PR adds the labels needed for alertmanager to forward the message
and enhances an existing test to ensure the expected labels are
configured.

Signed-off-by: Renan Campos <rcampos@redhat.com>

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -638,6 +638,17 @@ func (r *ManagedOCSReconciler) reconcileDMSPrometheusRule() error {
 		}
 
 		desired := templates.DMSPrometheusRuleTemplate.DeepCopy()
+
+		for _, group := range desired.Spec.Groups {
+			if group.Name == "snitch-alert" {
+				for _, rule := range group.Rules {
+					if rule.Alert == "DeadMansSnitch" {
+						rule.Labels["namespace"] = r.namespace
+					}
+				}
+			}
+		}
+
 		r.dmsRule.Spec = desired.Spec
 
 		return nil
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -897,7 +897,7 @@ var _ = Describe("ManagedOCS controller", func() {
 			})
 		})
 		When("the dms prometheus rule resource is deleted", func() {
-			It("should create a new dms prometheus rule in the namespace", func() {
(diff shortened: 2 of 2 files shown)


controllers/managedocs_controller.go
```diff
@@ -638,6 +638,7 @@ func (r *ManagedOCSReconciler) reconcileDMSPrometheusRule() error {
 		}
 
 		desired := templates.DMSPrometheusRuleTemplate.DeepCopy()
+		desired.Spec.Groups[0].Rules[0].Labels["namespace"] = r.namespace
```
nb-ohad, line 641: Accessing groups and labels by an index is a bad practice as it makes the code very fregile on futrue changes.
Can we use the group name `snitch-alert` and rule alert name `DeadMansSnitch` to find the proper item in the lists:

```suggestion
		for i range desired.Spec.Groups {
			group := &desired.Spec.Groups[i]
			if group.Name == `snitch-alert` {
				for j range group.Rules {
					rule := &group.Rules[j]
					if rule.Alert == `DeadMansSnitch` {
						rule.Labels["namespace"] = r.namespace
					}
				}
			}
		}
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
renan-campos: I have made the changes to make the code more flexible as you described but I'd like to point out a trade-off: Having to be flexible with the location of this group and rule now introduces logic into the test code. 

The test in the original commit enforced having the location of the 'snitch-alert' group and the 'DeadMansSnitch' alert rule at index 0. Since these are static lists whose changes would be caught by the test I thought that would be ok to index directly. 

I have it written both ways so whichever you choose we'll go with.

nb-ohad: @renan-campos I share your point as I didn't ask to change the test code, my ask was on the controller code. 

It is ok to have additional assumptions in test code, but then the responsibility, as you mentioned, is to have them tested. I would even say that I prefer the test code to take the location assumptions. 

So let's revert the test code back to the original contribution, but leave the new changes in the controller

nb-ohad (approved):

nb-ohad:
@renan-campos Please squash the commits and I will merge

PR #72, opened 2021-08-02
Added network policies to enhance security

kesavanvt:
Ingress and egress network polices are applied 

Ingress NetworkPolicy - Allow pod to pod communication within namespace(openshift-storage), Blocks traffic from outside openshift-storage. As this is applied to all the pods in the Namespace, default deny ingress networkpolicy is not required

EgressNetworkPolicy - Blocks external traffic and allowing only the required external domains


kesavanvt:
@nb-ohad Updated and ready for review

templates/networkPolicy.go
```diff
@@ -0,0 +1,72 @@
...
+			},
+			{
+				To: openshiftv1.EgressNetworkPolicyPeer{
+					CIDRSelector: "0.0.0.0/0",
+				},
+				Type: openshiftv1.EgressNetworkPolicyRuleDeny,
+			},
+		},
+	},
+}
```
nb-ohad, line 72: Let's keep a single resource template in a file, so we will have
`templates/networkpolicy.go` and `templates/egressnetworkpolicy.go` (matching the names of the resource kind)

While doing the change can you please also rename all other template files to lower case, the different operating system has different file systems that might or might not be case sensitive.
The offending files are: 
`dmsPrometheusRule.go` -> `dmsprometheusrule.go`
`k8sMetricsServiceMonitor.go` -> `k8smetricsservicemonitor.go`

controllers/managedocs_controller.go
```diff
@@ -1050,6 +1075,42 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
+		if err := r.own(r.egressNetworkPolicy); err != nil {
+			return err
+		}
+		desired := templates.EgressNetworkPolicyTemplate.DeepCopy()
+		dmsURL, err := url.Parse(string(r.deadMansSnitchSecret.Data["SNITCH_URL"]))
+		if err != nil {
+			return fmt.Errorf("Unable to parse DMS url: %v", err)
+		}
+		desired.Spec.Egress[0].To.DNSName = dmsURL.Hostname()
+		desired.Spec.Egress[1].To.DNSName = string(r.smtpSecret.Data["host"])
```
nb-ohad, line 1089: You cannot use a secret without verifying that it was loaded (as load it if it wasn't). If you do so it might be that you are just looking into an empty object. 

You should also not assume that this will run after or before any other reconcileXXXX function.

```suggestion
		if r.deadMansSnitchSecret.UID == "" {
			if err := r.get(r.deadMansSnitchSecret); err != nil {
				return fmt.Errorf("Unable to get DeadMan's Snitch secret: %v", err)
			}
		}
		dmsURLAsString := string(r.deadMansSnitchSecret.Data["SNITCH_URL"])
		if dmsURL == "" {
			return fmt.Errorf("DeadMan's Snitch secret does not contain a SNITCH_URL entry")
		}
		if dmsURL, err := url.Parse(dmsURLAsString); err == nil {
			desired.Spec.Egress[0].To.DNSName = dmsURL.Hostname()
			desired.Spec.Egress[1].To.DNSName = string(r.smtpSecret.Data["host"])
		} else {
			return fmt.Errorf("Unable to parse DMS url: %v", err)
		}
```

In order to prevent re-get of the secret in case of order change, I will also suggest adding the guard 
```if r.deadMansSnitchSecret.UID == "" {``` 
To the secret get on the `reconcileAlertmanagerConfig` function

**Review by nb-ohad — CHANGES_REQUESTED** [review]
As a general comment, I think we should add a test case that verifies that the egress policy holds the desired spec, with the DMS host. and that changes to the DMS host are reflected in the egress policy as well.

File: templates/networkpolicy.go [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+package templates
+
+import (
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var IngressNetworkPolicyTemplate = netv1.NetworkPolicy{
```
nb-ohad, line 24: There is no such thing as an Ingress network policy (in k8s), it just network policy, Can we rename the template accordengly?

controllers/managedocs_controller.go
```diff
@@ -1050,6 +1079,63 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
+				return fmt.Errorf("Unable to get SMTP secret: %v", err)
+			}
+		}
+		smtpHost := string(r.smtpSecret.Data["host"])
+		if smtpHost == "" {
+			return fmt.Errorf("smtp secret does not contain a host entry")
+		}
+
+		desired.Spec.Egress[0].To.DNSName = snitchURL.Hostname()
+		desired.Spec.Egress[1].To.DNSName = smtpHost
```
nb-ohad, line 1114: Addressing Egress rules by an index is very fregile. Especially on template changes.  
As these are not named objects my suggestion will be to have all static egress rules in the template and just append the dynamic ones.

So instead of doing:
```
desired.Spec.Egress[0].To.DNSName = snitchURL.Hostname()
```
you do
```
desired.Spec.Egress = append(desired.Spec.Egress, <new egress object with the proper hostname set>)

nb-ohad (changes requested):

nb-ohad (approved):

Issue #74, opened 2021-08-05
Empty string value from notification email are added to the alertingAddressList when user opts out for sendgrid service


kesavanvt:
Below are the key, values in add param secret when user opts for sendgrid service with 3 notification emails
```
  notification-email-0: "test0@email.com"
  notification-email-1: "test1@email.com"
  notification-email-2: "test2@email.com"
```

Empty string values are stored in the corresponding keys when user opts out
  ```
notification-email-0: ""
  notification-email-1: ""
  notification-email-2: ""
```
  
 Deployer needs to handle the empty string to avoid adding them to the `alertingAddressList`. This will prevent misconfiguration of sendgrid in alertmanagerconfig

PR #75, opened 2021-08-05
Fix to handle empty string from notification email when opt out

kesavanvt:
Fixes : #74 

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -761,7 +761,10 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 			alertingAddress, found := r.addonParamSecret.Data[notificationEmailKeyPrefix+fmt.Sprintf("-%v", i)]
 			i++
 			if found {
-				alertingAddressList = append(alertingAddressList, string(alertingAddress))
+				alertingAddressAsString := string(alertingAddress)
+				if alertingAddressAsString != "" {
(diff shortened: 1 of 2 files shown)


controllers/managedocs_controller.go
```diff
@@ -760,9 +760,10 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 		for {
 			alertingAddress, found := r.addonParamSecret.Data[notificationEmailKeyPrefix+fmt.Sprintf("-%v", i)]
 			i++
-			if found {
-				alertingAddressList = append(alertingAddressList, string(alertingAddress))
-			} else {
+			alertingAddressAsString := string(alertingAddress)
+			if found && alertingAddressAsString != "" {
+				alertingAddressList = append(alertingAddressList, alertingAddressAsString)
+			} else if !found {
```
nb-ohad, line 766: Why check found twice? 
```suggestion
			if found  {
				alertingAddressAsString := string(alertingAddress)
				if alertingAddressAsString != nil {
					alertingAddressList = append(alertingAddressList, alertingAddressAsString)
				}
			} else {
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
kesavanvt: Done

nb-ohad (approved):

PR #76, opened 2021-08-09
[Bug 1987271] Requeuing ManagedOCS CR upon conflict while uninstall

ezio-auditore:
While updating the status for the ManagedOCS cr the request is stale and it was causing a panic. This PR attempts to solve the issue by requeuing the request after 10 sec.


nb-ohad:
@ezio-auditore I don't get the fix. Returning an error will ensure a queue (and a log entry) so why do you propose to change that and how does it solve the panic?

PR #77, opened 2021-08-11
[WIP] Added metric and alert for Uninstallation Stuck Due To PVC

bindrad:

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	"github.com/prometheus/client_golang/prometheus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -47,6 +48,7 @@ import (
 	openshiftv1 "github.com/openshift/api/network/v1"
 	ocsv1 "github.com/openshift/ocs-operator/pkg/apis/ocs/v1"
 	v1 "github.com/openshift/ocs-osd-deployer/api/v1alpha1"
+	metrics "github.com/openshift/ocs-osd-deployer/metrics"
 	"github.com/openshift/ocs-osd-deployer/templates"
 	"github.com/openshift/ocs-osd-deployer/utils"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
@@ -83,6 +85,7 @@ const (
 	grafanaDatasourceSecretKey             = "prometheus.yaml"
 	k8sMetricsServiceMonitorAuthSecretName = "k8s-metrics-service-monitor-auth"
 	openshiftMonitoringNamespace           = "openshift-monitoring"
+	additionalRule                         = "additional-prometheus-rule"
 )
 
 // ManagedOCSReconciler reconciles a ManagedOCS object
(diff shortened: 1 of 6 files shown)


controllers/managedocs_controller.go
```diff
@@ -1234,6 +1246,37 @@ func (r *ManagedOCSReconciler) reconcileOCSCSV() error {
...
+			return err
+		}
+
+		desired := templates.AdditionalPrometheusRule.DeepCopy()
+
+		for _, group := range desired.Spec.Groups {
+			if group.Name == "pvc-uninstall-stuck-alert" {
+				for _, rule := range group.Rules {
+					if rule.Alert == "UninstallStuckDueToPVC" {
+						rule.Labels["namespace"] = r.namespace
```
kesavanvt, line 1263: I see you are adding only the namespace here, any specific reason why we are filtering by group name and alertname?. Can we simply add namespace to all existing rules?

controllers/managedocs_controller.go
```diff
@@ -1234,6 +1246,37 @@ func (r *ManagedOCSReconciler) reconcileOCSCSV() error {
...
+				}
+			}
+		}
+
+		r.additionalRule.Spec = desired.Spec
+
+		return nil
+	})
+	if err != nil {
+		return err
```
kesavanvt, line 1274: return meaningful error message

metrics/metrics.go
```diff
@@ -0,0 +1,20 @@
+package metrics
```
kesavanvt, line 1: Add license information

templates/additionalPrometheusRule.go
```diff
@@ -0,0 +1,49 @@
...
+*/
+
+package templates
+
+import (
+	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// This prometheus rule ensures that a DMS alert occurs during every prometheus scrape.
```
kesavanvt, line 24: DMS alert ?

kesavanvt:
Modify template filename to lowercase 
`additionalPrometheusRule.go`  ->` additionalprometheusrule.go `

controllers/managedocs_controller.go
bindrad: We only have two Prometheus rules in the deployer. First one is DMS and second one is this additional prometheus rule. DMS gets labeled in reconcileDMSPrometheusRule and AdditionalPrometheusRule get labeled in reconcileAdditionalPrometheusRule.

kesavanvt: I agree that both rules are labeled in different methods. My question was why do we check for `group.Name` to be `pvc-uninstall-stuck-alert` and `rule.Alert` to be `UninstallStuckDueToPVC`. Any reason on why we are doing this?. Why can't we just traverse through all the groups and alerts in `AdditionalPrometheusRule` and label with namespace?.

bindrad: Yes, I can do that. Anyways all the additionalPrometheusRule will need the namespace label

controllers/managedocs_controller.go
```diff
@@ -83,6 +85,7 @@ const (
 	grafanaDatasourceSecretKey             = "prometheus.yaml"
 	k8sMetricsServiceMonitorAuthSecretName = "k8s-metrics-service-monitor-auth"
 	openshiftMonitoringNamespace           = "openshift-monitoring"
+	additionalRule                         = "additional-prometheus-rule"
```
kesavanvt, line 88:
```suggestion
	additionalRuleName                         = "additional-prometheus-rule"
```

controllers/managedocs_controller.go
```diff
@@ -1149,6 +1163,29 @@ func (r *ManagedOCSReconciler) reconcileIngressNetworkPolicy() error {
...
 }
 
+func (r *ManagedOCSReconciler) reconcileAdditionalMetrics() error {
+	if r.checkUninstallCondition() {
+		metrics.ODFAddonPhase.With(
+			prometheus.Labels{
+				"status": "Uninstalling",
+			},
+		).Set(1)
+	} else if r.areComponentsReadyForUninstall() {
```
kesavanvt, line 1173: invoking this method is misleading though it satisfies the need, can we modify the methodname to `areComponentsReady`?

bindrad: Yes, I'll change the method name

bindrad:
deemed to be obsolete

PR #80, opened 2021-08-16
Handle SMTP secret when addon uninstall is initiated

kesavanvt:
The SMTP secret required for the email service is removed by OCM on uninstall trigger, this fix avoids setting up SMTP config when uninstall is initiated


kesavanvt:
Closing this one as the fix was made from app sre

PR #81, opened 2021-08-17
[Bug 1987271] Refreshing manangedOCS object after deletion is scheduled

ezio-auditore:
During uninstall when managedOCS object is scheduled for deletion a stale copy of it remains in local. When managedOCS status is updated it causes a conflict. This PR fixes that by refreshing the local copy of the object before status is updated.
Signed-off-by: kmajumder <kmajumder@redhat.com>

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -478,6 +478,14 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			if err := r.delete(r.managedOCS); err != nil {
 				return ctrl.Result{}, fmt.Errorf("unable to delete managedocs: %v", err)
 			}
+			// Refreshing local managedOCS object after deletion is scheduled
+			// to avoid conflict while updating status
+			if err := r.get(r.managedOCS); err != nil {
+				if !errors.IsNotFound(err) {
+					return ctrl.Result{}, err
+				}
+				r.Log.V(-1).Info("Trying to reload ManagedOCS resource after delete failed, ManagedOCS resource not found")
+			}
 		}
 
 	} else if initiateUninstall {


controllers/managedocs_controller.go
```diff
@@ -478,6 +478,11 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			if err := r.delete(r.managedOCS); err != nil {
 				return ctrl.Result{}, fmt.Errorf("unable to delete managedocs: %v", err)
 			}
+			// Refreshing local managedOCS object after deletion is scheduled
+			// to avoid conflict while updating status
+			if err := r.get(r.managedOCS); err != nil {
+				return ctrl.Result{}, nil
```
nb-ohad, line 484: Why are we ignoring all errors?

nb-ohad (changes requested):

controllers/managedocs_controller.go
ezio-auditore: Should I log the error and pass on to calling function?

ezio-auditore: As discussed offline , if resource not found then logging a warning and for other errors returning the error to the calling function.

controllers/managedocs_controller.go
```diff
@@ -478,6 +478,16 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
+			// Refreshing local managedOCS object after deletion is scheduled
+			// to avoid conflict while updating status
+			if err := r.get(r.managedOCS); err != nil {
+				if errors.IsNotFound(err) {
+					r.Log.Info("ManagedOCS resource not found")
+					return ctrl.Result{}, nil
+				} else {
+					return ctrl.Result{}, err
+				}
+			}
```
nb-ohad, line 490: Why the early return if no error? what if we would like to add more logic at the end of this function?
Also the message should be a warning and we need more context in the message itself.

```suggestion
			if err := r.get(r.managedOCS); err != nil &&  {
				if !errors.IsNotFound(err) {
					return ctrl.Result{}, err
				}
				r.Log.V(-1).Info("Trying to reload ManagedOCS resource after delete failed, ManagedOCS resource not found")
			}
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
ezio-auditore: Fixed in next iteration

nb-ohad (approved):

PR #82, opened 2021-08-25
[Bug 1995551] Fix pv mount failure issue

kesavanvt:
Add ingress rule for default namespace to accept traffic from default kubernetes service

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -1155,6 +1155,12 @@ func (r *ManagedOCSReconciler) reconcileIngressNetworkPolicy() error {
 			return err
 		}
 		desired := templates.NetworkPolicyTemplate.DeepCopy()
+		desired.Spec.Ingress[0].From = []netv1.NetworkPolicyPeer{
+			{
+				PodSelector: &metav1.LabelSelector{},
+			},
+		}
+		desired.Spec.PodSelector = metav1.LabelSelector{}
 		r.ingressNetworkPolicy.Spec = desired.Spec
 		return nil
 	})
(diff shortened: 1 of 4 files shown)


sabose:
@renan-campos could you review?

renan-campos:
Could you explain why this fixes the pv mount issue? It would help to have this as a comment in the code to understand why this addition ingress policy is added.

renan-campos:
In the BZ it was observed that traffic between csi pods and mon pods on different nodes is also not going through. This is needed if the client pod is on a different host from any of the mon pods. I found a similar way to get around this and posted it on the BZ. I think policy will also need to be added to this solution as well. Traffic between the mds pod and csi pod will need to be validated as well.

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -1155,6 +1164,45 @@ func (r *ManagedOCSReconciler) reconcileIngressNetworkPolicy() error {
...
+					Values: []string{
+						"rook-ceph-mds",
+						"rook-ceph-osd",
+						"rook-ceph-mgr",
+						"rook-ceph-mon",
+						"rook-ceph-crashcollector",
+					},
+				},
+			},
+		}
```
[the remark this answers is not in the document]
kesavanvt: Done

controllers/managedocs_controller.go
```diff
@@ -1155,6 +1164,45 @@ func (r *ManagedOCSReconciler) reconcileIngressNetworkPolicy() error {
...
+			MatchExpressions: []metav1.LabelSelectorRequirement{
+				{
+					Key:      "app",
+					Operator: metav1.LabelSelectorOpIn,
+					Values: []string{
+						"rook-ceph-mds",
+						"rook-ceph-osd",
+						"rook-ceph-mgr",
+						"rook-ceph-mon",
+						"rook-ceph-crashcollector",
```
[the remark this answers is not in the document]
kesavanvt: rook-ceph-toolbox is a hostnetwork pod and it doesn't obey network policy

controllers/managedocs_controller.go
```diff
@@ -1146,6 +1155,12 @@ func (r *ManagedOCSReconciler) reconcileIngressNetworkPolicy() error {
 			return err
 		}
 		desired := templates.NetworkPolicyTemplate.DeepCopy()
+		desired.Spec.Ingress[0].From = []netv1.NetworkPolicyPeer{
+			{
+				PodSelector: &metav1.LabelSelector{},
+			},
+		}
+		desired.Spec.PodSelector = metav1.LabelSelector{}
```
renan-campos, line 1163: Instead of one NetworkPolicy template that we modify in the operator code to satisfy the two Ingress Network policies, I think it would be better for readibility if there were two separate templates. Then for each NetworkPolicy, there would be a single location to check and modify each of the content of the policy.

nb-ohad: I agree with @renan-campos, It will allow each policy to grow independently in the future if needed, and will make the reconcile code even simpler, as currently, we have no dynamic information in the policies at all.

kesavanvt: Noted. Have made changes accordingly

templates/networkpolicy.go
```diff
@@ -38,3 +38,39 @@ var NetworkPolicyTemplate = netv1.NetworkPolicy{
...
+						"rook-ceph-osd",
+						"rook-ceph-mgr",
+						"rook-ceph-mon",
+						"rook-ceph-crashcollector",
+					},
+				},
+			},
+		},
+	},
+}
```
nb-ohad, line 76: Can we please keep each template in its own file

kesavanvt: Its of same kind, why its really required to have it in a separate file?

kesavanvt: Made changes accordingly

nb-ohad (approved):

PR #83, opened 2021-08-31
Added phase to status of managedocs

bindrad:
Phase of ManagedOCS can have three values, first value is Ready when the deployer is Ready to perform the reconcile operations but is not performing the operations.
The second value is Configuring, when the deployer is performing reconcile operations and making changes to the resources. 
The third value is Uninstalling, when the uninstallation of the addon is initiated the deployer will enter uninstall phase.


bindrad:
deemed to be obsolete

PR #88, opened 2021-09-14
Added changes to remove noobaa operator pod

kesavanvt:
OCS CSV is patched by scaling down the noobaa deployment to remove noobaa operator pod

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -1230,6 +1230,15 @@ func (r *ManagedOCSReconciler) reconcileOCSCSV() error {
 	var isChanged bool
 	deployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
 	for i := range deployments {
+		deployment := &deployments[i]
+		// Disable noobaa operator by scaling down the replica of noobaa deploymnet
+		// in OCS CSV.
+		if deployment.Name == "noobaa-operator" &&
+			(deployment.Spec.Replicas == nil || *deployment.Spec.Replicas > 0) {
+			zero := int32(0)
+			deployment.Spec.Replicas = &zero
+			isChanged = true
+		}
 		containers := deployments[i].Spec.Template.Spec.Containers
 		for j := range containers {
 			switch container := &containers[j]; container.Name {


controllers/managedocs_controller.go
```diff
@@ -1228,8 +1228,16 @@ func (r *ManagedOCSReconciler) reconcileOCSCSV() error {
 		return fmt.Errorf("OCS CSV not found")
 	}
 	var isChanged bool
+	var _0 = int32(0)
 	deployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
 	for i := range deployments {
+		deployment := &deployments[i]
+		// Disable noobaa operator by scaling down the replica of noobaa deploymnet 
+		// in OCS CSV.
+		if deployment.Name == "noobaa-operator" && deployment.Spec.Replicas != &_0 {
```
nb-ohad, line 1237: `deployment.Spec.Replicas != &_0` -> You are checking if the point has the same address of the local variable which will always be false.

The proper check should be:
```suggestion
		if deployment.Name == "noobaa-operator" && 
		(deployment.Spec.Replicas == nil |  *deployment.Spec.Replicas > 0) {
```

controllers/managedocs_controller.go
```diff
@@ -1228,8 +1228,16 @@ func (r *ManagedOCSReconciler) reconcileOCSCSV() error {
 		return fmt.Errorf("OCS CSV not found")
 	}
 	var isChanged bool
+	var _0 = int32(0)
```
nb-ohad, line 1231: This has no use above the condition (see comment about the issue in the condition) so you can just move it inside the if

controllers/managedocs_controller.go
```diff
@@ -1228,8 +1228,16 @@ func (r *ManagedOCSReconciler) reconcileOCSCSV() error {
...
 	deployments := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
 	for i := range deployments {
+		deployment := &deployments[i]
+		// Disable noobaa operator by scaling down the replica of noobaa deploymnet 
+		// in OCS CSV.
+		if deployment.Name == "noobaa-operator" && deployment.Spec.Replicas != &_0 {
+			deployment.Spec.Replicas = &_0
```
nb-ohad, line 1238:
```suggestion
			zero := int32(0)
			deployment.Spec.Replicas = &zero
```

nb-ohad (changes requested):

nb-ohad (approved):

PR #96, opened 2021-10-05
fixed the test

bindrad:

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -94,14 +94,15 @@ type ManagedOCSReconciler struct {
 	Log                logr.Logger
 	Scheme             *runtime.Scheme
 
-	AddonParamSecretName         string
-	AddonConfigMapName           string
-	AddonConfigMapDeleteLabelKey string
-	PagerdutySecretName          string
-	DeadMansSnitchSecretName     string
-	SMTPSecretName               string
-	SOPEndpoint                  string
-	AlertSMTPFrom                string
+	AddonParamSecretName             string
+	AddonConfigMapName               string
+	AddonConfigMapDeleteLabelKey     string
+	PagerdutySecretName              string
+	DeadMansSnitchSecretName         string
+	SMTPSecretName                   string
+	SOPEndpoint                      string
+	AlertSMTPFrom                    string
+	CustomerNotificationTemplatePath string
 
 	ctx                                context.Context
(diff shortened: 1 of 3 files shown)


bindrad:
@kesavanvt @renan-campos Can you review this PR

controllers/managedocs_controller.go
```diff
@@ -121,6 +121,7 @@ type ManagedOCSReconciler struct {
 	k8sMetricsServiceMonitorAuthSecret *corev1.Secret
 	namespace                          string
 	reconcileStrategy                  v1.ReconcileStrategy
+	CustomerNotificationTemplate       string
```
kesavanvt, line 124:
```suggestion
	CustomerNotificationTemplatePath       string
```

also can we move this declaration next to `AlertSMTPFrom`

bindrad: Done

kesavanvt: This change seems not scalable, instead of having the entire path in the `ManagedOCSReconciler` can we have just the template path ?
something like
TemplatePath  = "templates/"

and read the necessary file by concatenating under the appropriate method. In this case, we can avoid declaring multiple  variables under the `ManagedOCSReconciler` when template file grows

renan-campos (commented):
LGTM. As a future request, please add a description to future PRs so that the context of what you're doing is immediately known. Answering 'why the changes are being made', and 'what the expect results will be' would be a great help for reviewing.

bindrad:
> LGTM. As a future request, please add a description to future PRs so that the context of what you're doing is immediately known. Answering 'why the changes are being made', and 'what the expect results will be' would be a great help for reviewing.

@renan-campos Sure, will do next time

controllers/suite_test.go
```diff
@@ -64,6 +64,7 @@ const (
 	testGrafanaFederateSecretName              = "grafana-datasources"
 	testK8sMetricsServiceMonitorAuthSecretName = "k8s-metrics-service-monitor-auth"
 	testOpenshiftMonitoringNamespace           = "openshift-monitoring"
+	testCustomerNotificationTemplatePath       = "../templates/"
```
kesavanvt, line 67:
```suggestion
	testTemplatePath       = "../templates/"
```

kesavanvt (approved):
LGTM

nb-ohad (approved):

PR #104, opened 2021-10-19
Silenced CephOSDCriticallyFull and CephOSDNearFull alerts

bindrad:
These alerts are silenced because there is no scenario in our static deployment configuration where an OSD is getting full without the cluster getting full we will silence the OSD alerts.
CephOSDCriticallyFull
CephOSDNearFull
https://bugzilla.redhat.com/show_bug.cgi?id=2009696

### DIFF
--- a/templates/alertmanagerconfig.go
+++ b/templates/alertmanagerconfig.go
@@ -55,11 +55,16 @@ var smtpAlerts = []string{
 	"CephClusterNearFull",
 	"CephClusterCriticallyFull",
 	"CephClusterReadOnly",
-	"CephOSDCriticallyFull",
-	"CephOSDNearFull",
 	"PersistentVolumeUsageNearFull",
 	"PersistentVolumeUsageCritical",
 }
+
+// List of silenced alerts
+//
+// OSD Full alerts are silenced as there is no scenario in our static deployment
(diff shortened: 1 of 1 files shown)


renan-campos:
In the results of the BZ, Filip got three messages about the OSDs near full, but didn't report getting a message about the cluster being near full, which would come from the CephClusterNearFull alert. Can we verify that he received this alert as well?

renan-campos:
Or is that alert only fired when the OSDs are at 85% utilization?

templates/alertmanagerconfig.go
```diff
@@ -55,12 +55,16 @@ var smtpAlerts = []string{
...
 	"PersistentVolumeUsageNearFull",
 	"PersistentVolumeUsageCritical",
 }
 
+// List of alerts that are silenced because
+// there is no scenario in our static deployment configuration where an OSD is getting
+// full without the cluster getting full we will silence the OSD alerts.
+// CephOSDCriticallyFull
+// CephOSDNearFull
+
```
kesavanvt, line 67:
```suggestion
// List of silenced alerts.

// OSD Full alerts are silenced as there is no scenario in our static deployment configuration where an OSD is getting
// full without the cluster getting full.
// CephOSDCriticallyFull
// CephOSDNearFull

```

PR #110, opened 2021-11-23
Changes to Makefile to create bundle with required info`

bindrad:
Now bundle can be created by passing IMG, VERSION, REPLACES and NAMESPACE parameters, this will help to create downstream bundle easily and will reduce the manual changes.
Example command to create the bundle: make manifests && make bundle IMG=img:tag VERSION=3.3.3 REPLACES=3.3.2 NAMESPACE=openshift-storage


Makefile
```diff
@@ -3,6 +3,10 @@ VERSION ?= 1.1.1
 # Default bundle image tag
 IMAGE_TAG_BASE ?= controller
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
+# Default Namespace
+NAMESPACE ?= openshift-storage
```
nb-ohad, line 7: Why do we need this? If I am not mistaken the resulted namespace is already customizable on `make bundle` run (it is a kustomize feature)

Makefile
```diff
@@ -3,6 +3,10 @@ VERSION ?= 1.1.1
...
+# Default Replaces
+REPLACES ?= 1.1.0
```
nb-ohad, line 9: Why do we need to specify REPLACES? can't we just calc that from the version?

nb-ohad (changes requested):

Makefile
bindrad: We need this to add `namespace: openshift-storage` to the CSV. Currently when we run `make bundle` command the CSV has `namespace: placeholder` but I have used kustomize to add the `namespace: openshift-storage`(default) or we can pass NAMESPACE parameter with make bundle command to set it to something else. So yes this can be done using kustomize but we have to type it manually before running `make bundle` command but now it will run when `make bundle` command is executed.

Makefile
bindrad: We can calculate but not in every scenario. Let's suppose the current version is v1.1.0 then how can we determine the z-stream of the previous version? It can be v1.0.9 or v1.0.1 or anything else.
What do you think?

nb-ohad: Ok, agreed. But can we at least calc the `skipRange` value from this value?

config/manifests/bases/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: openshift-storage
+  
+resources:
+- ocs-osd-deployer.clusterserviceversion.yaml
```
nb-ohad, line 6: Let's not assume downstream conditions and restrictions for upstream builds (which mean let's not hard code a namespace)

config/manifests/kustomization.yaml
```diff
@@ -2,3 +2,4 @@ resources:
 - ../default
 - ../samples
 - ../scorecard
+- bases
```
nb-ohad, line 5: I don't think we need this, see last comment

nb-ohad (changes requested):

config/manifests/bases/kustomization.yaml
bindrad: Okay, I'll remove `namespace: openshift-storage` from kustomization.yaml but every time `make bundle` is executed it'll add `namespace: openshift-storage` to this file if the NAMESPACE parameter is not passed along with the command.
I mean to say that it's not hardcoded, kustomize add it to kustomization.yaml file when we build CSV and the namespace gets added depending on whatever is there in kustomization.yaml file.

Makefile
bindrad: skipRange is calculated depending on VERSION in the Makefile. So it would be `olm.skipRange: ">=0.0.1 <$(VERSION)`.

Makefile
```diff
@@ -163,9 +165,14 @@ endif
 .PHONY: bundle
 bundle: manifests kustomize
 	operator-sdk generate kustomize manifests -q
-	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
-	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --extra-service-accounts prometheus-k8s --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) $(BUNDLE_FLAGS)
+	cp -r config/manifests config/manifests-temp
```
nb-ohad, line 168: Can we start with an rm command in case there are leftovers format failed prev run?

nb-ohad (changes requested):

Makefile
bindrad: Done

nb-ohad:
@bindrad Can we try to revive this PR?

bindrad:
> @bindrad Can we try to revive this PR?

Yes, we can @nb-ohad

Makefile
```diff
@@ -34,7 +36,7 @@ ARCH = $(shell go env GOARCH)
 all: manager readinessServer
 
 export_env_vars:
-export NAMESPACE = openshift-storage
+export NAMESPACE ?= openshift-storage
```
nb-ohad, line 39: This is a downstream namespace, I don't think we want to include it in an upstream project. 
I guess the default should be `default`

config/default/kustomization.yaml
```diff
@@ -1,6 +1,3 @@
-# Adds namespace to all resources.
-namespace: openshift-storage
```
nb-ohad, line 2: Why are w committing a downstream namespace into the upstream repo?

nb-ohad (changes requested):

nb-ohad:
@bindrad Can we close this one?

bindrad:
Closing PR as work was done in another PR

PR #116, opened 2022-01-18
Changes to support multiple deployment type

bindrad:
This PR will help to support other deployment types in the future such as external, consumer, or provider in the same deployer codebase depending on the value of environment variable DEPLOYMENT_TYPE.

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -32,6 +32,7 @@ import (
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -106,6 +107,7 @@ type ManagedOCSReconciler struct {
 	SOPEndpoint                  string
 	AlertSMTPFrom                string
 	CustomerNotificationHTMLPath string
+	DeploymentType               string
 
 	ctx                                context.Context
 	managedOCS                         *v1.ManagedOCS
@@ -141,6 +143,7 @@ type ManagedOCSReconciler struct {
 // +kubebuilder:rbac:groups=operators.coreos.com,namespace=system,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
 // +kubebuilder:rbac:groups="apps",namespace=system,resources=statefulsets,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources={persistentvolumeclaims,secrets},verbs=get;list;watch
+// +kubebuilder:rbac:groups="storage.k8s.io",resources=storageclass,verbs=get;list;watch
 // +kubebuilder:rbac:groups="networking.k8s.io",namespace=system,resources=networkpolicies,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="network.openshift.io",namespace=system,resources=egressnetworkpolicies,verbs=create;get;list;watch;update
 
@@ -606,12 +609,16 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
 
 		// Handle only strict mode reconciliation
 		if r.reconcileStrategy == v1.ReconcileStrategyStrict {
-			// Get an instance of the desired state
-			desired := templates.StorageClusterTemplate.DeepCopy()
-			if err := r.updateStorageClusterFromAddonParamsSecret(desired); err != nil {
-				return err
+			var desired *ocsv1.StorageCluster = nil
+			switch strings.ToLower(r.DeploymentType) {
+			case "converged":
+				var err error
+				if desired, err = r.getDesiredConvergedStorageCluster(); err != nil {
+					return err
+				}
+			default:
+				return fmt.Errorf("Invalid deployment type value: %v", r.DeploymentType)
 			}
-
 			// Override storage cluster spec with desired spec from the template.
 			// We do not replace meta or status on purpose
 			r.storageCluster.Spec = desired.Spec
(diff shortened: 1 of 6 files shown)


controllers/managedocs_controller.go
```diff
@@ -607,11 +609,15 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
...
-			if err := r.updateStorageClusterFromAddonParamsSecret(desired); err != nil {
-				return err
+			var desired *ocsv1.StorageCluster = nil
+			var err error = nil
+			switch r.DeploymentType {
+			case "converged":
+				desired, err = r.getDesiredConvergedStorageCluster()
+				if err != nil {
+					return err
+				}
```
nb-ohad, line 619: I think e should have a default case where we return an error for unsupported values (right now it is everything that is not converged)

controllers/managedocs_controller.go
```diff
@@ -607,11 +609,15 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
 		// Handle only strict mode reconciliation
 		if r.reconcileStrategy == v1.ReconcileStrategyStrict {
 			// Get an instance of the desired state
-			desired := templates.StorageClusterTemplate.DeepCopy()
-			if err := r.updateStorageClusterFromAddonParamsSecret(desired); err != nil {
-				return err
+			var desired *ocsv1.StorageCluster = nil
+			var err error = nil
+			switch r.DeploymentType {
```
nb-ohad, line 614: Can we lowercase before switching? I would prefer for the value to be case insenstive

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -607,11 +609,15 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
 		// Handle only strict mode reconciliation
 		if r.reconcileStrategy == v1.ReconcileStrategyStrict {
 			// Get an instance of the desired state
-			desired := templates.StorageClusterTemplate.DeepCopy()
-			if err := r.updateStorageClusterFromAddonParamsSecret(desired); err != nil {
-				return err
+			var desired *ocsv1.StorageCluster = nil
+			var err error = nil
```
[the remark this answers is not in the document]
bindrad: That was a mistake, I wanted to declare it inside the case.

controllers/managedocs_controller.go
```diff
@@ -606,12 +609,17 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
...
-			// Get an instance of the desired state
-			desired := templates.StorageClusterTemplate.DeepCopy()
-			if err := r.updateStorageClusterFromAddonParamsSecret(desired); err != nil {
-				return err
+			var desired *ocsv1.StorageCluster = nil
+			switch strings.ToLower(r.DeploymentType) {
+			case "converged":
+				var err error = nil
+				desired, err = r.getDesiredConvergedStorageCluster()
+				if err != nil {
```
nb-ohad, line 617:
```suggestion
				if desired, err = r.getDesiredConvergedStorageCluster(); err != nil {
```

controllers/managedocs_controller.go
```diff
@@ -1265,14 +1275,31 @@ func (r *ManagedOCSReconciler) areComponentsReadyForUninstall() bool {
...
-	if err != nil {
+	if err := r.UnrestrictedClient.List(r.ctx, pvcList); err != nil {
 		return false, fmt.Errorf("unable to list pvcs: %v", err)
 	}
+
+	// check if there are any PVCs using OCS storage classes
 	for i := range pvcList.Items {
 		scName := *pvcList.Items[i].Spec.StorageClassName
-		if scName == storageClassCephFSName || scName == storageClassRbdName {
+		if ocsStorageClass[scName]{
```
nb-ohad, line 1302:
```suggestion
		if ocsStorageClass[scName] {
```

controllers/managedocs_controller.go
```diff
@@ -651,16 +659,18 @@ func (r *ManagedOCSReconciler) updateStorageClusterFromAddonParamsSecret(sc *ocs
 		}
 	}
 
+	desired := templates.StorageClusterTemplate.DeepCopy()
```
nb-ohad, line 662: Can we please call this `cs` instead of `desired` inside the function?
This will minimize the changes and in my eyes more meaningful in the cotext if this function

controllers/managedocs_controller.go
```diff
@@ -1278,26 +1280,27 @@ func (r *ManagedOCSReconciler) findOCSVolumeClaims() (bool, error) {
 	if err := r.UnrestrictedClient.List(r.ctx, &storageClassList); err != nil {
 		return false, fmt.Errorf("unable to list storage classes: %v", err)
 	}
-	// create slice of ocs storage classes by checking provisioner
+
+	// create map of ocs storage classes by checking provisioner
 	// Provisioner will be prefixed by namespace
```
nb-ohad, line 1285: Let try to capture the intent with the comment, my suggestion is something similar to the following text:

"Create a set of storage classed names who are using the ODF provisioner (provisioner name prefixed with the namespace name)"

controllers/managedocs_controller.go
```diff
@@ -1278,26 +1280,27 @@ func (r *ManagedOCSReconciler) findOCSVolumeClaims() (bool, error) {
...
-	ocsStorageClass := []string{}
+	ocsStorageClass := make(map[string]bool)
 	for i := range storageClassList.Items {
 		if strings.HasPrefix(storageClassList.Items[i].Provisioner, r.namespace) {
-			ocsStorageClass = append(ocsStorageClass, storageClassList.Items[i].Name)
+			ocsStorageClass[storageClassList.Items[i].Name] = true
 		}
```
nb-ohad, line 1290: I think the above suggestion is more readable

```suggestion 
		storageClass := &storageClassList.Items[i]
		if strings.HasPrefix(storageClass.Provisioner, r.namespace) {
			ocsStorageClass[storageClass.Name] = true
		}
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
bindrad: You mean `sc`, right?
Storage cluster

nb-ohad: yes, sorry

nb-ohad (approved):

PR #120, opened 2022-02-02
Changes to uninstallation flow according to ODF v4.10.0

bindrad:
ODF Operator creates storageSystem for the storageCluster if
storageSystem doesn't exist. While uninstallation deployer needs
to remove storageSystem after storageCluster is removed as storageSystem
has finalizer that prevents uninstallation.
Another resource called CSIAddonNodes has finalizer which is removed by
CSIAddons operator when ceph-csi pods are removed. ceph-csi pods are removed
when rook-operator is removed which is owned by OCS operator CSV. So we need
to remove OCS operator CSV as well.


controllers/managedocs_controller.go
```diff
@@ -1408,6 +1420,37 @@ func (r *ManagedOCSReconciler) removeOLMComponents() error {
...
+		return fmt.Errorf("unable to list storageSystem resource: %v", err)
+	}
+	for i := range storageSystemList.Items {
+		storageSystem := storageSystemList.Items[i]
+		if err := r.delete(&storageSystem); err != nil {
+			return fmt.Errorf("unable to delete storageSystem: %v", err)
+		}
+	}
+	return nil
+}
```
nb-ohad, line 1435: This is a util function that is only used once in a specific location and is very short. 
I see no need for a function in this case. Can we please inline this, with a comment at the start saying:
"Deleting all storage systems from the namespace"?

controllers/managedocs_controller.go
```diff
@@ -1408,6 +1420,37 @@ func (r *ManagedOCSReconciler) removeOLMComponents() error {
...
+	csv := getCSVByPrefix(csvList, ocsOperatorName)
+	if csv != nil {
+		if err := r.delete(csv); err != nil {
+			return fmt.Errorf("Unable to delete csv: %v", err)
+		}
+	}
+
+	r.Log.Info("OCS csv removed successfully")
+	return nil
+}
```
nb-ohad, line 1452: This is going to be the second call sire where we invoke `getCSVByPrefix`. In both cases, the code first lists all CSV. I think we should fold the CSV list operation into the util (same name) then inline this function `removeOCSCSV` in the call site (as it is going to be 5 line function that is invoked only once

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -405,6 +407,16 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
 		if r.verifyComponentsDoNotExist() {
+			// Deleting OCS CSV from the namespace
+			r.Log.Info("deleting OCS CSV")
+			csv, err := r.getCSVByPrefix(ocsOperatorName)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			if err := r.delete(csv); err != nil {
+				return ctrl.Result{}, fmt.Errorf("unable to delete csv: %v", err)
+			}
```
nb-ohad, line 418: Lets not pollute the function scope with an err var

```suggestion
			if csv, err := r.getCSVByPrefix(ocsOperatorName); err == nil {
				if err := r.delete(csv); err != nil {
					return ctrl.Result{}, fmt.Errorf("unable to delete csv: %v", err)
				}
			} else {
				return ctrl.Result{}, err
			}		
```

controllers/managedocs_controller.go
```diff
@@ -1392,16 +1417,13 @@ func (r *ManagedOCSReconciler) updateMCGCSV(csv *opv1a1.ClusterServiceVersion) e
 func (r *ManagedOCSReconciler) removeOLMComponents() error {
 
 	r.Log.Info("deleting deployer csv")
-	csvList := opv1a1.ClusterServiceVersionList{}
-	if err := r.list(&csvList); err != nil {
-		return fmt.Errorf("unable to list csv resources: %v", err)
+	csv, err := r.getCSVByPrefix(deployerCSVPrefix)
```
nb-ohad, line 1420: same as last comment

nb-ohad (approved):

PR #124, opened 2022-02-14
Store the addonParamasecret data in a map

rewantsoni:
We don't need to store the entire secret in the memory, we just need the data from the secret. So, moving the data of addonParamsecret to a map[string]string

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -119,7 +119,6 @@ type ManagedOCSReconciler struct {
 	prometheus                         *promv1.Prometheus
 	dmsRule                            *promv1.PrometheusRule
 	alertmanager                       *promv1.Alertmanager
-	addonParamSecret                   *corev1.Secret
 	pagerdutySecret                    *corev1.Secret
 	deadMansSnitchSecret               *corev1.Secret
 	smtpSecret                         *corev1.Secret
@@ -129,6 +128,7 @@ type ManagedOCSReconciler struct {
 	k8sMetricsServiceMonitorAuthSecret *corev1.Secret
 	namespace                          string
 	reconcileStrategy                  v1.ReconcileStrategy
+	addonParams                        map[string]string
 }
 
 // Add necessary rbac permissions for managedocs finalizer in order to set blockOwnerDeletion.
@@ -328,6 +328,7 @@ func (r *ManagedOCSReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 func (r *ManagedOCSReconciler) initReconciler(ctx context.Context, req ctrl.Request) {
 	r.ctx = ctx
 	r.namespace = req.NamespacedName.Namespace
+	r.addonParams = make(map[string]string)
 
 	r.managedOCS = &v1.ManagedOCS{}
 	r.managedOCS.Name = req.NamespacedName.Name
@@ -360,10 +361,6 @@ func (r *ManagedOCSReconciler) initReconciler(ctx context.Context, req ctrl.Requ
 	r.alertmanager = &promv1.Alertmanager{}
 	r.alertmanager.Name = alertmanagerName
 	r.alertmanager.Namespace = r.namespace
-
-	r.addonParamSecret = &corev1.Secret{}
-	r.addonParamSecret.Name = r.AddonParamSecretName
-	r.addonParamSecret.Namespace = r.namespace
 
 	r.pagerdutySecret = &corev1.Secret{}
 	r.pagerdutySecret.Name = r.PagerdutySecretName
@@ -436,8 +433,15 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			r.reconcileStrategy = v1.ReconcileStrategyNone
 		}
 
-		if err := r.get(r.addonParamSecret); err != nil {
-			return ctrl.Result{}, fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
+		// Read the add-on parameters secret and store it an addonParams map
+		addonParamSecret := &corev1.Secret{}
+		addonParamSecret.Name = r.AddonParamSecretName
+		addonParamSecret.Namespace = r.namespace
+		if err := r.get(addonParamSecret); err != nil {
+			return ctrl.Result{}, fmt.Errorf("Failed to get the addon parameters secret %v", r.AddonParamSecretName)
+		}
+		for key, value := range addonParamSecret.Data {
+			r.addonParams[key] = string(value)
 		}
 
 		// Reconcile the different resources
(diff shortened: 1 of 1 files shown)


controllers/managedocs_controller.go
```diff
@@ -361,9 +361,7 @@ func (r *ManagedOCSReconciler) initReconciler(ctx context.Context, req ctrl.Requ
 	r.alertmanager.Name = alertmanagerName
 	r.alertmanager.Namespace = r.namespace
 
-	r.addonParamSecret = &corev1.Secret{}
-	r.addonParamSecret.Name = r.AddonParamSecretName
-	r.addonParamSecret.Namespace = r.namespace
+	r.addonParams = make(map[string]string)
```
nb-ohad, line 364: Can we move this to the end of the init functions (or the start)? right now it is in the middle of resource initializations and it breaks the reading flow.

controllers/managedocs_controller.go
```diff
@@ -436,9 +434,25 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			r.reconcileStrategy = v1.ReconcileStrategyNone
 		}
 
-		if err := r.get(r.addonParamSecret); err != nil {
+		// Read the secret and store it in addonParams map
+		addonParamSecret := &corev1.Secret{}
+		addonParamSecret.Name = r.AddonParamSecretName
+		addonParamSecret.Namespace = r.namespace
+		if err := r.get(addonParamSecret); err != nil {
 			return ctrl.Result{}, fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
```
nb-ohad, line 442:
```suggestion
			return ctrl.Result{}, fmt.Errorf("Failed to get the addon parameters secret %v", r.AddonParamSecretName)
```

controllers/managedocs_controller.go
```diff
@@ -436,9 +434,25 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
+			enableMCGKey,
+			notificationEmailKeyPrefix + fmt.Sprintf("-%v", 0),
+			notificationEmailKeyPrefix + fmt.Sprintf("-%v", 1),
+			notificationEmailKeyPrefix + fmt.Sprintf("-%v", 2)}
+
+		for _, key := range keys {
+			if value, exists := addonParamSecret.Data[key]; exists {
+				r.addonParams[key] = string(value)
+			}
+		}
```
nb-ohad, line 455: You don't need a key list, just iterate over the map

```suggestion
		for key, value := range addonParamSecret.Data {
			r.addonParams[key] = string(value)
		}
```

controllers/managedocs_controller.go
```diff
@@ -866,7 +878,7 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 		alertingAddressList := []string{}
 		i := 0
 		for {
-			alertingAddress, found := r.addonParamSecret.Data[notificationEmailKeyPrefix+fmt.Sprintf("-%v", i)]
+			alertingAddress, found := r.addonParams[notificationEmailKeyPrefix+fmt.Sprintf("-%v", i)]
```
nb-ohad, line 881:
```suggestion
			alertingAddressKey = fmt.Sprintf("%s-%v", notificationEmailKeyPrefix, i)
			alertingAddress, found := r.addonParams[alertingAddressKey]
```

controllers/managedocs_controller.go
```diff
@@ -119,7 +119,7 @@ type ManagedOCSReconciler struct {
 	prometheus                         *promv1.Prometheus
 	dmsRule                            *promv1.PrometheusRule
 	alertmanager                       *promv1.Alertmanager
-	addonParamSecret                   *corev1.Secret
+	addonParams                        map[string]string
```
nb-ohad, line 122: Can we please move this to the end after all resource var definitions?

controllers/managedocs_controller.go
```diff
@@ -633,14 +647,12 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
...
-	sizeAsString := string(addonParams[storageClassSizeKey])
-	enableMCGRaw, exists := addonParams[enableMCGKey]
+	sizeAsString := r.addonParams[storageClassSizeKey]
+	enableMCGRaw, exists := r.addonParams[enableMCGKey]
 	// Setting hardcoded value here to force no MCG deployment
 	enableMCGAsString := "false"
 	if exists {
-		enableMCGAsString = string(enableMCGRaw)
+		enableMCGAsString = enableMCGRaw
 	}
```
nb-ohad, line 656: This can be refactored into a form that does not pollute the function scope with an `exists` var 
```
	sizeAsString := r.addonParams[storageClassSizeKey]
	
	// Setting hardcoded value here to force no MCG deployment
	enableMCGAsString := "false"
	if enableMCGRaw, exists := r.addonParams[enableMCGKey]; exists {
		enableMCGAsString = enableMCGRaw
	}
```

controllers/managedocs_controller.go
```diff
@@ -436,8 +433,16 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
-			return ctrl.Result{}, fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
+		// Read the secret and store it in addonParams map
+		addonParamSecret := &corev1.Secret{}
+		addonParamSecret.Name = r.AddonParamSecretName
+		addonParamSecret.Namespace = r.namespace
+		if err := r.get(addonParamSecret); err != nil {
+			return ctrl.Result{}, fmt.Errorf("Failed to get the addon parameters secret %v", r.AddonParamSecretName)
+		}
+		for key, value := range addonParamSecret.Data {
+			fmt.Println(key, value)
```
leelavg, line 444: - I believe, it isn't good to either print or log contents of addonParamSecret which may contain onboarding ticket and storage endpoints
- @nb-ohad your take on this?

rewantsoni: I missed that, yes that should be removed. Thanks for pointing it out.

nb-ohad: The onboarding ticket is a one-time use token, so by the time it appears in the log it was already consumed.
But your point remains, especially around user-provided emails.

@rewantsoni What was the intention of printing the secret data into the log at this location?
Usually when we have an error regarding a specific parameter we print the specific parameter with that error, if the parameter is not to be disclosed in a log then we just do not print it

controllers/managedocs_controller.go
```diff
@@ -436,8 +433,15 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			r.reconcileStrategy = v1.ReconcileStrategyNone
 		}
 
-		if err := r.get(r.addonParamSecret); err != nil {
-			return ctrl.Result{}, fmt.Errorf("Failed to get the addon param secret, Secret Name: %v", r.AddonParamSecretName)
+		// Read the secret and store it in addonParams map
```
nb-ohad, line 436: Can we be more specific with the comment? 
Maybe: // Read the add-on parameters secret and store it an addonParams map

controllers/managedocs_controller.go
rewantsoni: I have updated the PR without the print statement

nb-ohad (approved):

nb-ohad:
@rewantsoni Please confirm that unit test are passing

rewantsoni:
> @rewantsoni Please confirm that unit tests are passing

@nb-ohad I ran all the unit tests and they are all passing.

PR #125, opened 2022-02-17
[WIP] Support Provider Deployment type for OCS-OSD deployer

bindrad:


controllers/managedocs_controller.go
```diff
@@ -72,6 +72,7 @@ const (
 	storageClassSizeKey                    = "size"
 	enableMCGKey                           = "enable-mcg"
 	notificationEmailKeyPrefix             = "notification-email"
+	onboardingTicketKey                    = "onboarding-ticket-key"
```
nb-ohad, line 75: Why is the field name have `key` in it? (not the var name). 
Also, providers do not have onboarding tickets. 
I think what you are referring to is "onboarding-validation-key"

controllers/managedocs_controller.go
```diff
@@ -91,6 +92,7 @@ const (
 	openshiftMonitoringNamespace           = "openshift-monitoring"
 	alertRelabelConfigSecretName           = "managed-ocs-alert-relabel-config-secret"
 	alertRelabelConfigSecretKey            = "alertrelabelconfig.yaml"
+	onboardingTicketSecretName             = "onboarding-ticket-key"
```
nb-ohad, line 95: Is this the name the product expects for the secret? even if it is let's name the var something more sane.
Maybe: `onboardingValidationKeySecretName`

controllers/managedocs_controller.go
```diff
@@ -130,6 +132,7 @@ type ManagedOCSReconciler struct {
 	namespace                          string
 	reconcileStrategy                  v1.ReconcileStrategy
 	addonParams                        map[string]string
+	onboardingTicketSecret             *corev1.Secret
```
nb-ohad, line 135: Same as above, suggesting `onboardingValidationKeySecret`

controllers/managedocs_controller.go
```diff
@@ -662,8 +674,38 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
 }
 
 func (r *ManagedOCSReconciler) getDesiredConvergedStorageCluster() (*ocsv1.StorageCluster, error) {
-	sizeAsString := r.addonParams[storageClassSizeKey]
+	sc, err := r.getDesiredStorageCluster(templates.StorageClusterTemplate.DeepCopy())
+	if err != nil {
+		return nil, err
+	}
+	return sc, nil
+}
```
nb-ohad, line 682: Why do we need another function that basically does what this function is supposed to do?
right now this function body is equivalent to:
```
       return r.getDesiredStorageCluster(templates.StorageClusterTemplate.DeepCopy())
```
which is meaningless 


Also, why are we touching converged in the first place?

controllers/managedocs_controller.go
```diff
@@ -662,8 +674,38 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
...
+		r.onboardingTicketSecret.Data = map[string][]byte{
+			"key": []byte(onboardingTicketKeyData),
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to update onboardingTicketSecret: %v", err)
+	}
+
+	sc, err := r.getDesiredStorageCluster(templates.ProviderStorageClusterTemplate.DeepCopy())
```
nb-ohad, line 699: Same as above, why do we need another helper method? provider and converged should have separate implementations that can evolve separately.  

If you need shared functionality between converged and provider please move the specifics into a util function not the entire reconcile function

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -677,46 +690,76 @@ func (r *ManagedOCSReconciler) getDesiredConvergedStorageCluster() (*ocsv1.Stora
...
+func (r *ManagedOCSReconciler) getDesiredProviderStorageCluster() (*ocsv1.StorageCluster, error) {
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.onboardingValidationKeySecret, func() error {
+		if err := r.own(r.onboardingValidationKeySecret); err != nil {
+			return err
+		}
+		r.onboardingValidationKeySecret.Data = map[string][]byte{
+			"key": []byte(r.addonParams[onboardingValidationKey]),
+		}
+		return nil
+	})
```
nb-ohad, line 725: Why do we have a `CreateOrUpdate` here? by design, each reconcile function should only reconcile a single output resource. If we need an `reconcileOnboardinValidationSecret` function  then lets add it

controllers/managedocs_controller.go
```diff
@@ -677,46 +690,76 @@ func (r *ManagedOCSReconciler) getDesiredConvergedStorageCluster() (*ocsv1.Stora
...
 	if err != nil {
-		return nil, fmt.Errorf("Invalid Enable MCG value: %v", enableMCGAsString)
+		return nil, fmt.Errorf("Failed to update k8sMetricsServiceMonitorAuthSecret: %v", err)
```
nb-ohad, line 727: Did you ment to change the error message here to `k8sMetricsServiceMonitorAuthSecret`?

controllers/managedocs_controller.go
```diff
@@ -677,46 +690,76 @@ func (r *ManagedOCSReconciler) getDesiredConvergedStorageCluster() (*ocsv1.Stora
...
+	// Get the desired storage device set from storage cluster template
+	sc := templates.ProviderStorageClusterTemplate.DeepCopy()
+	var ds *ocsv1.StorageDeviceSet = nil
+	if desiredStorageDeviceSet := getStorageDeviceSet(sc.Spec.StorageDeviceSets, deviceSetName); desiredStorageDeviceSet != nil {
+		ds = desiredStorageDeviceSet
+	}
+
+	// Prevent downscaling by comparing count from secret and count from storage cluster
+	r.Log.Info("Setting storage device set count", "Current", currDeviceSetCount, "New", desiredDeviceSetCount)
+	r.updateDeviceSetCount(ds, desiredDeviceSetCount, currDeviceSetCount)
```
nb-ohad, line 758: Can we move the logging into the util?

templates/providerstoragecluster.go
```diff
@@ -0,0 +1,86 @@
...
+		// The label selector is used to select only the worker nodes for
+		// both labeling and scheduling.
+		LabelSelector: &metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{{
+				Key:      "node-role.kubernetes.io/worker",
+				Operator: metav1.LabelSelectorOpExists,
+			}, {
+				Key:      "node-role.kubernetes.io/infra",
+				Operator: metav1.LabelSelectorOpDoesNotExist,
+			}},
```
nb-ohad, line 41: We should not use infra nodes at all, please remove this selector

templates/providerstoragecluster.go
```diff
@@ -0,0 +1,86 @@
...
+						Requests: corev1.ResourceList{
+							"storage": resource.MustParse("2Ti"),
+						},
+					},
+				},
+			},
+			Placement: rook.Placement{},
+			Portable:  true,
+			Replica:   3,
+			Resources: utils.GetResourceRequirements("sds"),
```
nb-ohad, line 78: Are we sure the prev sds requirements matched 2Ti? I rememver we measured that for 1Ti

nb-ohad (changes requested):

templates/providerstoragecluster.go
bindrad: Operator is LabelSelectorOpDoesNotExist(DoesNotExist) which ensure that it won't use infra nodes, we are doing same for converged storage cluster as well.

controllers/managedocs_controller.go
bindrad: That was a mistake, I didn't intend to change it.

controllers/managedocs_controller.go
```diff
@@ -1482,3 +1525,37 @@ func (r *ManagedOCSReconciler) unrestrictedGet(obj client.Object) error {
...
+	if mcgEnable {
+		r.Log.Info("Enabling Multi Cloud Gateway")
+		storageCluster.Spec.MultiCloudGateway.ReconcileStrategy = "manage"
+	} else if storageCluster.Spec.MultiCloudGateway.ReconcileStrategy == "manage" {
+		r.Log.V(-1).Info("Trying to disable Multi Cloud Gateway, Invalid operation")
+	}
+	return nil
+}
+
+func (r *ManagedOCSReconciler) updateDeviceSetCount(deviceSet *ocsv1.StorageDeviceSet, desiredDeviceSetCount int, currDeviceSetCount int) {
```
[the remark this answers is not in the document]
bindrad: Because it needs to log if downscale is attempted

templates/providerstoragecluster.go
bindrad: Sahina mentioned that for Provider we will have 2 TiB for every OSD

templates/providerstoragecluster.go
nb-ohad: Haa, ok. I have missed that detail. Thanks for pointing this out

templates/providerstoragecluster.go
nb-ohad: I think you have misunderstood. 
We will use 2Tib, my question was if `utils.GetResourceRequirements("sds")` fits the 2Tib configuration or do we need  another key for it with different configuration (something like utils.GetResourceRequirements("sds2tib"))

controllers/managedocs_controller.go
```diff
@@ -522,6 +527,12 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 				return ctrl.Result{}, err
 			}
 		}
+		// Reconciles only for provider deployment type
+		if strings.ToLower(r.DeploymentType) == "provider" {
+			if err := r.reconcileOnboardinValidationSecret(); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
```
nb-ohad, line 535: Can we move the check inside the `≈reconcileOnboardinValidationSecret` as an early exit?
So we have something like:

```
if !strings.ToLower(r.DeploymentType) != "provider" {
     r.Log.Info("Non provider deployment, skipping reconcile for onboardin validation secret")
     return nil
}
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: Ok

templates/providerstoragecluster.go
bindrad: When we tested Provider consumer setup OSDs were coming up without any issue so I don't think we need to change resourceRequirements. Do you have any better way to determine resourceRequirements?

nb-ohad: @oritwas can you please provide a recommendation

controllers/managedocs_controller.go
```diff
@@ -405,6 +411,8 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 	// Update the status of the components
 	r.updateComponentStatus()
 
+	r.DeploymentType = strings.ToLower(r.DeploymentType)
```
nb-ohad, line 414: Maybe in the future, we will need to take deployment type into consideration when we `updateComponentStatus` or `checkUninstallCondition`  so this need to be the first thing in the function or even better move it outside of `reconcilePhases`

controllers/managedocs_controller.go
```diff
@@ -509,8 +517,7 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 		if err := r.reconcileOCSInitialization(); err != nil {
 			return ctrl.Result{}, err
 		}
-
-		// fix for non-converged deployment will be provided in upcoming PR
+		// Fix for non-converged deployment type will be provided in upcoming PR
 		if r.DeploymentType == "converged" {
```
nb-ohad, line 521: Please use constants for deployment types ("converged", " consumer", "provider") - this was already asked in a previous review

controllers/managedocs_controller.go
```diff
@@ -646,7 +657,7 @@ func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
 		// Handle only strict mode reconciliation
 		if r.reconcileStrategy == v1.ReconcileStrategyStrict {
 			var desired *ocsv1.StorageCluster = nil
-			switch strings.ToLower(r.DeploymentType) {
+			switch r.DeploymentType {
 			case "converged":
```
nb-ohad, line 661: same as last comment (also for the other cases in the switch)

controllers/managedocs_controller.go
```diff
@@ -772,6 +813,27 @@ func (r *ManagedOCSReconciler) getDesiredConsumerStorageCluster() (*ocsv1.Storag
 	return sc, nil
 }
 
+func (r *ManagedOCSReconciler) reconcileOnboardinValidationSecret() error {
+	if r.DeploymentType != "provider" {
```
nb-ohad, line 817: same as last comment

nb-ohad (changes requested):

nb-ohad (approved):

PR #130, opened 2022-03-03
Tests for provider deployment

bindrad:

### DIFF
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -288,6 +288,12 @@ var _ = Describe("ManagedOCS controller", func() {
 			Namespace: testPrimaryNamespace,
 		},
 	}
+	onboardingValidationKeySecretTemplate := corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      onboardingValidationKeySecretName,
+			Namespace: testPrimaryNamespace,
+		},
+	}
 
 	setupUninstallConditions := func(
 		shouldAddonConfigMapExist bool,
@@ -646,17 +652,47 @@ var _ = Describe("ManagedOCS controller", func() {
 			})
 			When("there is a valid size in the add-on parameter secret", func() {
 				It("should create reconciled resources", func() {
-					if testReconciler.DeploymentType == consumerDeploymentType {
-						Skip(fmt.Sprintf("Skipping the test as it is not required by %v", testReconciler.DeploymentType))
-					}
+
 					// Create a valid add-on parameters secret
 					secret := addonParamsSecretTemplate.DeepCopy()
-					secret.Data["size"] = []byte("1")
-					secret.Data["enable-mcg"] = []byte("false")
+					switch testReconciler.DeploymentType {
+					case convergedDeploymentType:
+						secret.Data["size"] = []byte("1")
+						secret.Data["enable-mcg"] = []byte("false")
+					case consumerDeploymentType:
+						secret.Data["size"] = []byte("1")
+						secret.Data["enable-mcg"] = []byte("false")
+						secret.Data["onboarding-ticket"] = []byte("onboarding-tickets")
+						secret.Data["storage-provider-endpoint"] = []byte("0.0.0.0:36179")
+					case providerDeploymentType:
+						secret.Data["size"] = []byte("1")
+						secret.Data["enable-mcg"] = []byte("false")
+						secret.Data["onboarding-validation-key"] = []byte("   test-validation-key   ")
+					}
 					Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
 
+					if testReconciler.DeploymentType == providerDeploymentType {
+						By("Creating a onboarding validation key secret resource")
+						utils.WaitForResource(k8sClient, ctx, onboardingValidationKeySecretTemplate.DeepCopy(), timeout, interval)
+					}
+
 					By("Creating a storagecluster resource")
-					utils.WaitForResource(k8sClient, ctx, scTemplate.DeepCopy(), timeout, interval)
+					sc := scTemplate.DeepCopy()
+					utils.WaitForResource(k8sClient, ctx, sc, timeout, interval)
+					switch testReconciler.DeploymentType {
+					case convergedDeploymentType:
+						ds := sc.Spec.StorageDeviceSets[0]
+						Expect(ds.Count).Should(Equal(1))
+					case consumerDeploymentType:
+						Expect(sc.Spec.ExternalStorage.Enable).Should(BeTrue())
+						Expect(sc.Spec.ExternalStorage.OnboardingTicket).Should(Equal("onboarding-tickets"))
+						Expect(sc.Spec.ExternalStorage.StorageProviderEndpoint).Should(Equal("0.0.0.0:36179"))
+					case providerDeploymentType:
+						ds := sc.Spec.StorageDeviceSets[0]
+						Expect(ds.Count).Should(Equal(1))
+						Expect(sc.Spec.HostNetwork).Should(BeTrue())
+						Expect(sc.Spec.AllowRemoteStorageConsumers).Should(BeTrue())
+					}
 
 					By("Creating a prometheus resource")
 					utils.WaitForResource(k8sClient, ctx, promTemplate.DeepCopy(), timeout, interval)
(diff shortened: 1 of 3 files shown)


controllers/managedocs_controller_test.go
```diff
@@ -640,7 +651,7 @@ var _ = Describe("ManagedOCS controller", func() {
 			})
 			When("there is a valid size in the add-on parameter secret", func() {
```
nb-ohad, line 652: Why does this case have 3 different its that validate the same "should create reconciled resources" with a different skip?

To me is seems like what we want is a single "it" with an internal switch statement for the difference details

controllers/managedocs_controller_test.go
```diff
@@ -1139,8 +1170,8 @@ var _ = Describe("ManagedOCS controller", func() {
 			})
 			When("a Grafana datasources secret exists in the openshift-monitoring namespace", func() {
 				It("should create k8sMetricsServiceMonitorAuthSecret in primary namespace", func() {
-					grafanaSecret := grafanaDatasourceSecretTemplate.DeepCopy()
-					Expect(k8sClient.Create(ctx, grafanaSecret)).Should(Succeed())
+					secret := grafanaDatasourceSecretTemplate.DeepCopy()
+					Expect(k8sClient.Get(ctx, utils.GetResourceKey(secret), secret)).Should(Succeed())
```
nb-ohad, line 1174: if you assume the secret is already created you should not test for it, 
Expect is used to validate test results, not assumptions

controllers/suite_test.go
```diff
@@ -177,6 +177,22 @@ var _ = BeforeSuite(func() {
...
+		}
+		grafanaDatasources.Datasources[0].BasicAuthPassword = "password"
+		grafanaDatasources.Datasources[0].BasicAuthUser = "internal"
+		grafanaDatasourceSecret := &corev1.Secret{}
+		grafanaDatasourceSecret.Name = testGrafanaFederateSecretName
+		grafanaDatasourceSecret.Namespace = testOpenshiftMonitoringNamespace
+		grafanaDatasourceSecret.Data = map[string][]byte{
+			"prometheus.yaml": utils.ToJsonOrDie(grafanaDatasources),
+		}
+		Expect(k8sClient.Create(ctx, grafanaDatasourceSecret)).ShouldNot(HaveOccurred())
```
nb-ohad, line 194: Why was this moved here?

controllers/managedocs_controller_test.go
```diff
@@ -676,6 +687,26 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					}
+					// Create a valid add-on parameters secret
+					secret := addonParamsSecretTemplate.DeepCopy()
+					secret.Data["size"] = []byte("1")
+					secret.Data["enable-mcg"] = []byte("false")
+					secret.Data["onboarding-validation-key"] = []byte("test-validation-key")
+					Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+
+					By("Creating a storagecluster resource")
+					utils.WaitForResource(k8sClient, ctx, scTemplate.DeepCopy(), timeout, interval)
```
nb-ohad, line 705: We have different deployment types now so we need to validate the content as well. 
I think a deep equal on a desired outcome would be good

controllers/managedocs_controller_test.go
```diff
@@ -536,6 +531,22 @@ var _ = Describe("ManagedOCS controller", func() {
 					Expect(k8sClient.Delete(ctx, secret)).Should(Succeed())
 				})
 			})
+			When("there is no onboarding-validation-key field in the add-on parameters secret", func() {
```
nb-ohad, line 534: Where is the opposite test that verifies successful creation with correct content?

nb-ohad (changes requested):

controllers/suite_test.go
bindrad: This resource should by default exist in the monitoring namespace and it shouldn't be created while the test is running. So added it to BeforeSuite which ensures that the resource exists in the monitoring namespace.

nb-ohad: How did it work before?

bindrad: Before we were creating the secret in monitoring namespace in managedocs_controller_test.go file just before we check for k8sMetricsServiceMonitorAuthSecret
https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/controllers/managedocs_controller_test.go#L1142-L1143

controllers/managedocs_controller_test.go
```diff
@@ -1594,6 +1610,15 @@ var _ = Describe("ManagedOCS controller", func() {
...
 			})
+			When("there is onboarding-validation-key field in the add-on parameters secret", func() {
+				It("should create onboarding-ticket-key secret", func() {
+					if testReconciler.DeploymentType != providerDeploymentType {
+						Skip(fmt.Sprintf("Skipping the test as it is not required by %v", testReconciler.DeploymentType))
+					}
+					secret := onboardingValidationKeySecretTemplate.DeepCopy()
+					Expect(k8sClient.Get(ctx, utils.GetResourceKey(secret), secret)).Should(Succeed())
+				})
+			})
```
nb-ohad, line 1621: This test does not make sense? 
The secret is created whenever we deploy with the correct params, as such this needs to be checked together with the other resources that are created in that test.

Here you have no set up, so it cannot be a `when`

controllers/suite_test.go
nb-ohad: The old way makes more sense as it is the trigger for the test. 
Right now you do not actually have a test as you do not have a trigger for it. 

Let's revert.

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
```diff
@@ -640,13 +646,18 @@ var _ = Describe("ManagedOCS controller", func() {
 			})
 			When("there is a valid size in the add-on parameter secret", func() {
 				It("should create reconciled resources", func() {
-					if testReconciler.DeploymentType == consumerDeploymentType {
-						Skip(fmt.Sprintf("Skipping the test as it is not required by %v", testReconciler.DeploymentType))
-					}
+
```
leelavg, line 649: - shouldn't this test be skipped for converged?

bindrad: No, this test should be for all the deployment types. The resources that gets created might be different depending on deployment types. And the storageCluster spec is different depending on deployment type.

leelavg (approved):
/lgtm

controllers/managedocs_controller_test.go
```diff
@@ -640,44 +646,69 @@ var _ = Describe("ManagedOCS controller", func() {
...
 					By("Creating a prometheus resource")
 					utils.WaitForResource(k8sClient, ctx, promTemplate.DeepCopy(), timeout, interval)
 
 					By("Creating an alertmanager resource")
 					utils.WaitForResource(k8sClient, ctx, amTemplate.DeepCopy(), timeout, interval)
+
+					if testReconciler.DeploymentType == providerDeploymentType {
+						By("Creating a onboarding validation key secret resource")
+						utils.WaitForResource(k8sClient, ctx, onboardingValidationKeySecretTemplate.DeepCopy(), timeout, interval)
+					}
```
nb-ohad, line 695: I think we should validate that before validating the storage cluster

controllers/managedocs_controller_test.go
```diff
@@ -640,44 +646,69 @@ var _ = Describe("ManagedOCS controller", func() {
...
-					By("Creating an alertmanager resource")
-					utils.WaitForResource(k8sClient, ctx, amTemplate.DeepCopy(), timeout, interval)
+					Eventually(func() bool {
+						secret := onboardingValidationKeySecretTemplate.DeepCopy()
+						err := k8sClient.Get(ctx, utils.GetResourceKey(secret), secret)
+						if err != nil {
+							return false
+						}
+						return secret.Data != nil && string(secret.Data["key"]) == "-----BEGIN PUBLIC KEY-----\ntest-validation-key\n-----END PUBLIC KEY-----"
+					}, timeout, interval).Should(BeTrue())
```
nb-ohad, line 711: Why not use the same pattern as in other tests? (`utils.WaitForResource`)

nb-ohad (changes requested):

nb-ohad (approved):

nb-ohad:
@bindrad Please resolve the conflict so I can merge

bindrad:
@nb-ohad Resolved merge conflicts

PR #133, opened 2022-03-09
Disable ceph-csi and default storageclasses on provider deployment

bindrad:

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -1204,104 +1204,100 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
 		return fmt.Errorf("Failed to get Rook ConfigMap: %v", err)
 	}
 
-	if rookConfigMap.Data == nil {
-		rookConfigMap.Data = map[string]string{}
+	cloneRookConfigMap := rookConfigMap.DeepCopy()
+	if cloneRookConfigMap.Data == nil {
+		cloneRookConfigMap.Data = map[string]string{}
 	}
 
-	rbdProvisionerRequirements := utils.MarshalRookResourceRequirements(utils.RookResourceRequirementsList{
-		{
-			Name:     "csi-provisioner",
-			Resource: utils.GetResourceRequirements("csi-provisioner"),
-		},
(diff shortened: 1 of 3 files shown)


controllers/managedocs_controller.go
```diff
@@ -483,7 +483,11 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
 
 		// Reconcile the different resources
-		if err := r.reconcileRookCephOperatorConfig(); err != nil {
+		if err := r.reconcileConvergedRookCephOperatorConfig(); err != nil {
+			return ctrl.Result{}, err
+		}
+		// Reconciles only for provider deployment type
+		if err := r.reconcileProviderRookCephOperatorConfig(); err != nil {
 			return ctrl.Result{}, err
 		}
```
nb-ohad, line 492: The code for both helpers seems almost identical and is not predicated to diverge any further in the future. 
Can we just solve it with a simple if inside the original `reconcileRookCephOperatorConfig` func?

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -1297,11 +1297,19 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
 		rookConfigMap.Data["CSI_CEPHFS_PLUGIN_RESOURCE"] = fsPluginRequirements
+	}
 
-		if err := r.update(rookConfigMap); err != nil {
-			return fmt.Errorf("Failed to update Rook ConfigMap: %v", err)
+	// Update rookConfigMap only for provider deployment
+	// Disable ceph-csi on provider deployment
+	if r.DeploymentType == providerDeploymentType {
+		if rookConfigMap.Data["ROOK_CSI_ENABLE_CEPHFS"] != "false" ||
+			rookConfigMap.Data["ROOK_CSI_ENABLE_RBD"] != "false" {
```
Madhu-1, line 1306: why this check is required? you can directly assign "false" value if it's a providerDeploymentType?

nb-ohad (approved):

PR #134, opened 2022-03-09
monitoring: Add missing alerts to pagerDuty config

pcuzner:
This PR adds 3 alerts to pagerduty escalation;
CephMonQuorumLost
CephOSDCriticallyFull
CephOSDNearFull

Signed-off-by: Paul Cuzner <pcuzner@redhat.com>

### DIFF
--- a/templates/alertmanagerconfig.go
+++ b/templates/alertmanagerconfig.go
@@ -46,9 +46,12 @@ var pagerdutyAlerts = []string{
 	"CephOSDFlapping",
 	"CephOSDDiskNotResponding",
 	"CephOSDDiskUnavailable",
+	"CephOSDCriticallyFull",
+	"CephOSDNearFull",
 	"CephDataRecoveryTakingTooLong",
 	"CephPGRepairTakingTooLong",
 	"CephMonQuorumAtRisk",
+	"CephMonQuorumLost",
 	"CephMonHighNumberOfLeaderChanges",
 }
 var smtpAlerts = []string{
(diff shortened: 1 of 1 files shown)


templates/alertmanagerconfig.go
```diff
@@ -46,9 +46,12 @@ var pagerdutyAlerts = []string{
 	"CephOSDFlapping",
 	"CephOSDDiskNotResponding",
 	"CephOSDDiskUnavailable",
+	"CephOSDCriticallyFull",
+	"CephOSDNearFull",
```
nb-ohad, line 50: These two alerts are already in the ignore section below (L69-L70 - List of silenced alerts)
If you do not agree with the comment explaining why then please remove them from that section.

But I still fail to understand what are we expecting SRE to do when they get such an alert? SRE cannot expand the cluster on behalf of the customer (as we need explicit consent to do so) nor they can delete customer data. If SRE cannot handle an alert there is no need to send the alert to the SRE (PagerDuty) as it just create noise

nb-ohad (changes requested):

templates/alertmanagerconfig.go
pcuzner: I've updated the comments, and avoided the use of "silenced" since in alertmanager terms a silenced alert is something different. What's happening here is that alerts are being ignored/skipped where they don't make sense for the use case.

Confused by your EBS reference for the outlier problem. The issue that the osd_full an osd_nearfull event pick up on is an individual osd utilization breaching a threshold. Although at a small scale data distribution can be regarded as uniform, this is not a valid assumption as the cluster grows, or the data stored (file, block, object) changes.

ODF sets osd_full at 85% (ceph default is 95%) - so there is still freespace in the cluster. From an operational perspective, if the CephOSDCriticallyFull is triggered because of an outlier problem, you could temporarily lift the osd full to allow writes and maintain service, and run a reweight to redistribute the data, then reduce the full threshold again. You would also respond to CephOSDNearFull, to check that the cluster is not subject to outlier utilisation issues - before you reach CephOSDCriticallyFull and end up with a DOS.

PR #135, opened 2022-03-14
Accept public key without header and footer

bindrad:

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -825,8 +825,12 @@ func (r *ManagedOCSReconciler) reconcileOnboardinValidationSecret() error {
 		if err := r.own(r.onboardingValidationKeySecret); err != nil {
 			return err
 		}
+		onboardingValidationData := fmt.Sprintf(
+			"-----BEGIN PUBLIC KEY-----\n%s\n-----END PUBLIC KEY-----",
+			strings.TrimSpace(r.addonParams[onboardingValidationKey]),
+		)
 		r.onboardingValidationKeySecret.Data = map[string][]byte{
-			"key": []byte(r.addonParams[onboardingValidationKey]),
+			"key": []byte(onboardingValidationData),
 		}
 		return nil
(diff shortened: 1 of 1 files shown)


controllers/managedocs_controller.go
```diff
@@ -825,9 +825,11 @@ func (r *ManagedOCSReconciler) reconcileOnboardinValidationSecret() error {
 		if err := r.own(r.onboardingValidationKeySecret); err != nil {
 			return err
 		}
+		onboardingValidationData := "-----BEGIN PUBLIC KEY-----\n" + r.addonParams[onboardingValidationKey] + "\n-----END PUBLIC KEY-----"
```
nb-ohad, line 828: A couple of points
1. use `fmt` instead of manual string concatenation 
2. trim (for whitespaces) the passed `r.addonParams[onboardingValidationKey]` value

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -825,8 +825,10 @@ func (r *ManagedOCSReconciler) reconcileOnboardinValidationSecret() error {
 		if err := r.own(r.onboardingValidationKeySecret); err != nil {
 			return err
 		}
+		r.addonParams[onboardingValidationKey] = strings.TrimSpace(r.addonParams[onboardingValidationKey])
+		onboardingValidationData := fmt.Sprintf("-----BEGIN PUBLIC KEY-----\n%s\n-----END PUBLIC KEY-----", r.addonParams[onboardingValidationKey])
```
nb-ohad, line 829: Do not mutate the actually provided value.

```suggestion
		onboardingValidationData := fmt.Sprintf(
			"-----BEGIN PUBLIC KEY-----\n%s\n-----END PUBLIC KEY-----", 												
			strings.TrimSpace(r.addonParams[onboardingValidationKey]),
		)
```

nb-ohad (approved):

PR #136, opened 2022-03-14
fix: check for empty string in env values

leelavg:

### DIFF
--- a/main.go
+++ b/main.go
@@ -173,7 +173,7 @@ func readEnvVars() (map[string]string, error) {
 	}
 	for envVarName := range envVars {
 		val, found := os.LookupEnv(envVarName)
-		if !found {
+		if !found || val == "" {
 			return nil, fmt.Errorf("%s environment variable must be set", envVarName)
 		}
 		envVars[envVarName] = val


nb-ohad:
@leelavg I believe this is the wrong fix, some env var can and should have empty values on some environments/addons. 
We should find the underlaying issue and fix it

leelavg:
- ack, I went through the env vars that we are using before raising the PR and found that we need all the variables should have values
- agreed! finding underlying issue takes priority

nb-ohad:
The fix can and will break our `make run` or at least will require us to have mock values for every single supported env var.

leelavg:
- I think it'll not break `make run` unless I'm missing something due to https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/Makefile#L36

nb-ohad:
@leelavg I am appreciating your point of view, but I do want to keep the distinction between an empty env var and an undefined one.

Also, I do believe that this change will only hide cases where env vars are defined but empty (by design), even if we do not have a concrete case now. I don't want to limit our future selves, because if we start to rely on this assumption - treating empty and undefined env vars the same - it will be very hard to go back from it.

leelavg:
> treating empty and undefined env vars the same
- understood, we don't want this behavior then, thanks for clarification

PR #137, opened 2022-03-14
Add ingress network policy to support the Provider API Server.

renan-campos:
Signed-off-by: Renan Campos <rcampos@redhat.com>

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -85,6 +85,7 @@ const (
 	egressNetworkPolicyName                = "egress-rule"
 	ingressNetworkPolicyName               = "ingress-rule"
 	cephIngressNetworkPolicyName           = "ceph-ingress-rule"
+	providerApiServerNetworkPolicyName     = "provider-api-server-rule"
 	monLabelKey                            = "app"
 	monLabelValue                          = "managed-ocs"
 	rookConfigMapName                      = "rook-ceph-operator-config"
@@ -125,6 +126,7 @@ type ManagedOCSReconciler struct {
 	egressNetworkPolicy                *openshiftv1.EgressNetworkPolicy
 	ingressNetworkPolicy               *netv1.NetworkPolicy
 	cephIngressNetworkPolicy           *netv1.NetworkPolicy
+	providerAPIServerNetworkPolicy     *netv1.NetworkPolicy
 	prometheus                         *promv1.Prometheus
 	dmsRule                            *promv1.PrometheusRule
 	alertmanager                       *promv1.Alertmanager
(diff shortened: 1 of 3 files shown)


renan-campos:
@bindrad @rewantsoni @nb-ohad Take a gander at this here PR when you have a chance. I look forward to your inputs!

controllers/managedocs_controller.go
```diff
@@ -528,10 +534,15 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
-			if err := r.reconcileCephIngressNetworkPolicy(); err != nil {
+		}
+		if err := r.reconcileIngressNetworkPolicy(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileCephIngressNetworkPolicy(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if r.DeploymentType == providerDeploymentType {
+			if err := r.reconcileProviderAPIServerNetworkPolicy(); err != nil {
```
nb-ohad, line 545: Can we please move the condition into the reconcile function, as a guard at the start of the function? 
This is the pattern we agreed on and we are utilizing it in other reconcile functions. 
You can take a look at `reconcileOnboardingValidationSecret` for an example

controllers/managedocs_controller.go
```diff
@@ -528,10 +534,15 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			if err := r.reconcileEgressNetworkPolicy(); err != nil {
```
nb-ohad, line 534: Can we please move the condition into the reconcile function, as a guard at the start of the function?
This is the pattern we agreed on and we are utilizing it in other reconcile functions.
You can take a look at reconcileOnboardingValidationSecret for an example

controllers/managedocs_controller.go
```diff
@@ -1392,6 +1403,21 @@ func (r *ManagedOCSReconciler) reconcileCephIngressNetworkPolicy() error {
...
+		}
+		desired := templates.ProviderApiServerNetworkPolicyTemplate.DeepCopy()
+		r.providerAPIServerNetworkPolicy.Spec = desired.Spec
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Failed to update provider api server NetworkPolicy: %v", err)
+	}
+	return nil
+}
```
nb-ohad, line 1419: Is there a reason we need a different/new policy for this? can't we just specify this in the ingress/egress policy?

nb-ohad (changes requested):

controllers/managedocs_controller.go
renan-campos: What you suggest would work, but I thought this would be better for readability since it matches the existing pattern: there is a network policy for ingressNetworkPolicyName, and a separate one for cephIngressNetworkPolicyName. 

If you would like me to write it the way you suggest, which of these policies should I put the new ingress rule in?

nb-ohad: Ok, you made your case :) 
let's leave it the way you wrote it

templates/provider_api_server_networkpolicy.go
```diff
@@ -0,0 +1,56 @@
...
+						Port:     &ProviderApiServerPort,
+						Protocol: &ProviderApiServerProtocol,
+					},
+				},
+			},
+		},
+		PolicyTypes: []netv1.PolicyType{
+			netv1.PolicyTypeIngress,
+		},
+		PodSelector: metav1.LabelSelector{
```
leelavg, line 44: - maybe I'm wrong here, IIRC provider api server already exposes service, we better target that rather than the pod?

nb-ohad: Right now, the provider API server is not exposing a service and is exposed via host port on every host, this rule is definitely needed.

leelavg: - ack, the last time I saw, I remember provider api service with NodePort and maybe I missed some commits in b/n.
- no further comments.

PR #138, opened 2022-03-15
Change replacement of the prom CR labels map with just adding the lab…

nb-ohad:
…el to the current map

This is done to prevent overriding of existing labels and allow other agents to add external labels to the CR


leelavg (approved):
/lgtm

PR #142, opened 2022-03-22
deploy: use Recreate strategy for deployment

leelavg:
Background:
- k8s default strategy for deployment is rolling update
- when the readiness of the resource fails and deployment rolled out
  this default strategy will create new pod resource

Issue:
- we are leveraging olm functionality for setting env values in
  deployment
- when deployment is first rolled out the pod fails to progress into
  running state due to absence of required env values
- later the deployment is rolled out with env values and resulting in
  existence of two pods on a lease with incorrect active pod

Solution:
- set deployment strategy to Recreate as our deployer is stateless,
  doesn't do any IO ops and could reconcile the desired state

Testing done:
- deploy the manager with default strategy and perform
  `oc rollout restart deploy` and verify existence of two ocs-osd pods
- set deployment spec.strategy.type to Recreate and perform restart
  again and observe existence of only new ocs-osd pod

Credits: @nb-ohad

Minor fix: use .dockerignore to avoid sending testdata as context to
docker engine while building images


leelavg:
- pls review @nb-ohad @bindrad

nb-ohad (approved):

PR #149, opened 2022-03-28
Adding TLS support to prometheus instance

aruniiird:
TLS support enables endpoints to be more secured.

Signed-off-by: Arun Kumar Mohan <amohan@redhat.com>

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -39,6 +39,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -68,6 +69,11 @@ const (
 	managedOCSName                         = "managedocs"
 	storageClusterName                     = "ocs-storagecluster"
 	prometheusName                         = "managed-ocs-prometheus"
+	prometheusServiceName                  = "managed-ocs-prometheus-service"
+	servingCertSecretBetaAnnotationKey     = "service.beta.openshift.io/serving-cert-secret-name"
+	servingCertSecretAlphaAnnotationKey    = "service.alpha.openshift.io/serving-cert-secret-name"
+	servingCertConfigMapBetaAnnotationKey  = "service.beta.openshift.io/inject-cabundle"
+	servingCertConfigMapAlphaAnnotationKey = "service.alpha.openshift.io/inject-cabundle"
 	alertmanagerName                       = "managed-ocs-alertmanager"
 	alertmanagerConfigName                 = "managed-ocs-alertmanager-config"
 	dmsRuleName                            = "dms-monitor-rule"
@@ -102,6 +108,8 @@ const (
 	convergedDeploymentType                = "converged"
 	consumerDeploymentType                 = "consumer"
 	providerDeploymentType                 = "provider"
+	servingCertCABundleConfigMapName       = "prometheus-serving-cert-ca-bundle"
+	prometheusServiceSecretName            = "managed-ocs-prometheus-service-secret"
 )
 
 // ManagedOCSReconciler reconciles a ManagedOCS object
@@ -130,6 +138,7 @@ type ManagedOCSReconciler struct {
 	cephIngressNetworkPolicy           *netv1.NetworkPolicy
 	providerAPIServerNetworkPolicy     *netv1.NetworkPolicy
 	prometheus                         *promv1.Prometheus
+	prometheusService                  *corev1.Service
 	dmsRule                            *promv1.PrometheusRule
 	alertmanager                       *promv1.Alertmanager
 	pagerdutySecret                    *corev1.Secret
@@ -143,6 +152,7 @@ type ManagedOCSReconciler struct {
 	reconcileStrategy                  v1.ReconcileStrategy
 	addonParams                        map[string]string
 	onboardingValidationKeySecret      *corev1.Secret
+	servingCertCaBundleConfigMap       *corev1.ConfigMap
 }
 
 // Add necessary rbac permissions for managedocs finalizer in order to set blockOwnerDeletion.
@@ -158,6 +168,7 @@ type ManagedOCSReconciler struct {
 // +kubebuilder:rbac:groups=operators.coreos.com,namespace=system,resources=clusterserviceversions,verbs=get;list;watch;delete;update;patch
 // +kubebuilder:rbac:groups="apps",namespace=system,resources=statefulsets,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources={persistentvolumes,secrets},verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources={services},verbs=get;list;watch;create;update
 // +kubebuilder:rbac:groups="storage.k8s.io",resources=storageclass,verbs=get;list;watch
 // +kubebuilder:rbac:groups="networking.k8s.io",namespace=system,resources=networkpolicies,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="network.openshift.io",namespace=system,resources=egressnetworkpolicies,verbs=create;get;list;watch;update
@@ -255,6 +266,7 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&openshiftv1.EgressNetworkPolicy{}).
 		Owns(&netv1.NetworkPolicy{}).
 		Owns(&corev1.Secret{}).
+		Owns(&corev1.Service{}).
 
 		// Watch non-owned resources
 		Watches(
@@ -374,6 +386,10 @@ func (r *ManagedOCSReconciler) initReconciler(ctx context.Context, req ctrl.Requ
 	r.prometheus.Name = prometheusName
 	r.prometheus.Namespace = r.namespace
 
+	r.prometheusService = &corev1.Service{}
+	r.prometheusService.Name = prometheusServiceName
+	r.prometheusService.Namespace = r.namespace
+
 	r.dmsRule = &promv1.PrometheusRule{}
 	r.dmsRule.Name = dmsRuleName
 	r.dmsRule.Namespace = r.namespace
@@ -409,6 +425,10 @@ func (r *ManagedOCSReconciler) initReconciler(ctx context.Context, req ctrl.Requ
 	r.alertRelabelConfigSecret = &corev1.Secret{}
 	r.alertRelabelConfigSecret.Name = alertRelabelConfigSecretName
 	r.alertRelabelConfigSecret.Namespace = r.namespace
+
+	r.servingCertCaBundleConfigMap = &corev1.ConfigMap{}
+	r.servingCertCaBundleConfigMap.Name = servingCertCABundleConfigMapName
+	r.servingCertCaBundleConfigMap.Namespace = r.namespace
 
 	r.onboardingValidationKeySecret = &corev1.Secret{}
 	r.onboardingValidationKeySecret.Name = onboardingValidationKeySecretName
@@ -501,6 +521,12 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			return ctrl.Result{}, err
 		}
 		if err := r.reconcileAlertRelabelConfigSecret(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcilePrometheusService(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileServingCertCABundleConfigMap(); err != nil {
 			return ctrl.Result{}, err
 		}
 		if err := r.reconcilePrometheus(); err != nil {
@@ -885,6 +911,51 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
 	return nil
 }
 
+// reconcilePrometheusService function wait for prometheus Service
+// to start and sets appropriate annotation for 'service-ca' controller
+func (r *ManagedOCSReconciler) reconcilePrometheusService() error {
+	r.Log.Info("Reconciling prometheusService")
+
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.prometheusService, func() error {
+		if err := r.own(r.prometheusService); err != nil {
+			return err
+		}
+
+		r.prometheusService.Spec.Ports = []corev1.ServicePort{
+			{
+				Name:       "odf-prometheus-endpoint",
+				Protocol:   corev1.ProtocolTCP,
+				Port:       9339,
+				TargetPort: intstr.FromString("web"),
+			},
+		}
+		r.prometheusService.Spec.Selector = map[string]string{
+			monLabelKey: monLabelValue,
+		}
+		utils.AddAnnotation(r.prometheusService,
+			servingCertSecretBetaAnnotationKey, prometheusServiceSecretName)
+		utils.AddAnnotation(r.prometheusService,
+			servingCertSecretAlphaAnnotationKey, prometheusServiceSecretName)
+		return nil
+	})
+	return err
+}
+
+func (r *ManagedOCSReconciler) reconcileServingCertCABundleConfigMap() error {
+	r.Log.Info("Reconciling servingCertCABundleConfigMap")
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.servingCertCaBundleConfigMap, func() error {
+		if err := r.own(r.servingCertCaBundleConfigMap); err != nil {
+			return err
+		}
+		utils.AddAnnotation(r.servingCertCaBundleConfigMap,
+			servingCertConfigMapBetaAnnotationKey, "true")
+		utils.AddAnnotation(r.servingCertCaBundleConfigMap,
+			servingCertConfigMapAlphaAnnotationKey, "true")
+		return nil
+	})
+	return err
+}
+
 func (r *ManagedOCSReconciler) reconcilePrometheus() error {
 	r.Log.Info("Reconciling Prometheus")
 
@@ -903,6 +974,11 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
 			},
 			Key: alertRelabelConfigSecretKey,
 		}
+
+		tlsConfig := r.prometheus.Spec.Web.TLSConfig
+		tlsConfig.KeySecret.LocalObjectReference.Name = prometheusServiceSecretName
+		tlsConfig.Cert.Secret.LocalObjectReference.Name = prometheusServiceSecretName
+		tlsConfig.ClientCA.ConfigMap.LocalObjectReference.Name = servingCertCABundleConfigMapName
 
 		return nil
 	})
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -46,6 +46,18 @@ var _ = Describe("ManagedOCS controller", func() {
 			Namespace: testPrimaryNamespace,
 		},
 	}
+	promServiceTemplate := corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      prometheusServiceName,
+			Namespace: testPrimaryNamespace,
+		},
+	}
+	caBundleConfigMap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      servingCertCABundleConfigMapName,
+			Namespace: testPrimaryNamespace,
+		},
+	}
 	promTemplate := promv1.Prometheus{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      prometheusName,
@@ -494,6 +506,19 @@ var _ = Describe("ManagedOCS controller", func() {
 		}
 	}
 
+	assertPrometheusTLSConfigValues := func(ctx context.Context,
+		client client.Client) {
+		prom := promTemplate.DeepCopy()
+		Expect(client.Get(ctx, utils.GetResourceKey(prom), prom)).Should(Succeed())
+		tlsConfig := prom.Spec.Web.TLSConfig
+		Expect(tlsConfig.KeySecret.LocalObjectReference.Name).Should(Equal(prometheusServiceSecretName))
+		Expect(tlsConfig.Cert.Secret.LocalObjectReference.Name).Should(Equal(prometheusServiceSecretName))
+		Expect(tlsConfig.ClientCA.ConfigMap.LocalObjectReference.Name).Should(Equal(servingCertCABundleConfigMapName))
+		Expect(tlsConfig.ClientCA.ConfigMap.Key).Should(Equal("service-ca.crt"))
+		Expect(tlsConfig.KeySecret.Key).Should(Equal("tls.key"))
+		Expect(tlsConfig.Cert.Secret.Key).Should(Equal("tls.crt"))
+	}
+
 	runTests := func(deploymentType string) {
 		Context(fmt.Sprintf("reconcile() %s mode", deploymentType), Ordered, func() {
 			BeforeAll(func() {
@@ -730,8 +755,13 @@ var _ = Describe("ManagedOCS controller", func() {
 						Expect(sc.Spec.AllowRemoteStorageConsumers).Should(BeTrue())
 					}
 
+					By("Creating a prometheus service resource")
+					utils.WaitForResource(k8sClient, ctx, promServiceTemplate.DeepCopy(), timeout, interval)
+					By("Creating serving cert CA bundle configmap")
+					utils.WaitForResource(k8sClient, ctx, caBundleConfigMap.DeepCopy(), timeout, interval)
 					By("Creating a prometheus resource")
 					utils.WaitForResource(k8sClient, ctx, promTemplate.DeepCopy(), timeout, interval)
+					assertPrometheusTLSConfigValues(ctx, k8sClient)
 
 					By("Creating an alertmanager resource")
 					utils.WaitForResource(k8sClient, ctx, amTemplate.DeepCopy(), timeout, interval)
--- a/templates/prometheus.go
+++ b/templates/prometheus.go
@@ -17,8 +17,12 @@ limitations under the License.
 package templates
 
 import (
+	"crypto/tls"
+	"fmt"
+
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/red-hat-storage/ocs-osd-deployer/utils"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
@@ -45,5 +49,23 @@ var PrometheusTemplate = promv1.Prometheus{
 			}},
 		},
 		Resources: utils.GetResourceRequirements("prometheus"),
+		Web: &promv1.WebSpec{
+			TLSConfig: &promv1.WebTLSConfig{
+				KeySecret: corev1.SecretKeySelector{
+					Key: "tls.key",
+				},
+				Cert: promv1.SecretOrConfigMap{
+					Secret: &corev1.SecretKeySelector{
+						Key: "tls.crt",
+					},
+				},
+				ClientAuthType: fmt.Sprint(tls.VerifyClientCertIfGiven),
+				ClientCA: promv1.SecretOrConfigMap{
+					ConfigMap: &corev1.ConfigMapKeySelector{
+						Key: "service-ca.crt",
+					},
+				},
+			},
+		},
 	},
 }
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -16,6 +16,16 @@ rules:
   - list
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - create
+  - get
+  - list
+  - update
+  - watch
+- apiGroups:
   - storage.k8s.io
   resources:
   - storageclass
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -17,8 +17,9 @@ limitations under the License.
 package utils
 
 import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // Contains checks whether a string is contained within a slice
@@ -50,6 +51,15 @@ func AddLabel(obj metav1.Object, key string, value string) {
 		obj.SetLabels(labels)
 	}
 	labels[key] = value
+}
+
+func AddAnnotation(obj metav1.Object, key string, value string) {
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+		obj.SetAnnotations(annotations)
+	}
+	annotations[key] = value
 }
 
 // GetRegexMatcher converts list of alerts to regex matcher


controllers/managedocs_controller.go
```diff
@@ -67,6 +67,8 @@ const (
 	managedOCSName                         = "managedocs"
 	storageClusterName                     = "ocs-storagecluster"
 	prometheusName                         = "managed-ocs-prometheus"
+	prometheusServiceName                  = "prometheus-operated"
+	serviceServingCertAnnotationKeyBeta    = "service.beta.openshift.io/serving-cert-secret-name"
```
nb-ohad, line 71:
```suggestion
	serviceServingCertBetaAnnotationKey   = "service.beta.openshift.io/serving-cert-secret-name"
```

controllers/managedocs_controller.go
```diff
@@ -911,6 +916,26 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+		prometheusService)
+	// as this Service is not created explicitly by us,
+	// we don't raise any error, will wait till it is available
+	if err != nil {
+		return nil
+	}
+	// check whether the Service has appropriate annotation
+	if annotVal := prometheusService.Annotations[serviceServingCertAnnotationKeyBeta]; annotVal != prometheusServiceName {
+		prometheusService.Annotations[serviceServingCertAnnotationKeyBeta] = prometheusServiceName
+	}
```
nb-ohad, line 935: Some issues:
1. The annotation map might be nil
2. What you wrote here is equivalent to just setting the value
3. You are not setting the annotations via the interface

```suggestion
	anno := prometheusService.GetAnnotations()
	if (anno == nil) {
		anno = new map[string]string{}
		prometheusService.SetAnnotations()
	}
	anno[serviceServingCertAnnotationKeyBeta] = prometheusServiceName
```

The above works because a map is a ref type

controllers/managedocs_controller.go
```diff
@@ -911,6 +916,26 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
 }
 
+// reconcilePrometheusService function wait for prometheus Service
+// to start and sets appropriate annotation for 'service-ca' controller
+func (r *ManagedOCSReconciler) reconcilePrometheusService() error {
+	r.Log.Info("Reconciling Prometheus Service")
+	prometheusService := &corev1.Service{}
+	err := r.Client.Get(r.ctx,
+		types.NamespacedName{Name: prometheusServiceName, Namespace: r.namespace},
+		prometheusService)
```
nb-ohad, line 926: Please use our utils instead of r.client directly.

```suggestion
		prometheusService := &corev1.Service{}
		prometheusService.Name = prometheusServiceName
		prometheusService.Namespace = r.namespace
		if err := r.get(prometheusService); err != nil {
			return fmt.Errorf("Failed to get the prometheus service %s secret %v", prometheusServiceName, r.AddonParamSecretName)
		}
```

controllers/managedocs_controller.go
```diff
@@ -911,6 +916,26 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+	// as this Service is not created explicitly by us,
+	// we don't raise any error, will wait till it is available
+	if err != nil {
+		return nil
+	}
```
nb-ohad, line 931: With the current changes, the above statement is inaccurate. 
You need to add a watch for the service in order for it to work (in the watch section at the start of this file)

In addition, not all errors are NotFound errors, and for other errors, you want to return the error so it will requeue another reconcile.

In my eyes it is also ok to just return an error for not found and not rely on a create event, worst case scenario you will only get one additional reconcile.

See last comment for the suggested return error

controllers/managedocs_controller.go
```diff
@@ -911,6 +916,26 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+	// check whether the Service has appropriate annotation
+	if annotVal := prometheusService.Annotations[serviceServingCertAnnotationKeyBeta]; annotVal != prometheusServiceName {
+		prometheusService.Annotations[serviceServingCertAnnotationKeyBeta] = prometheusServiceName
+	}
+	return r.Client.Update(r.ctx, prometheusService)
```
nb-ohad, line 936: We should automatically, this will just cause another requeue to happen (which is not needed and might cause an infinite loop of reconciles) 

Please update only if have to.

controllers/managedocs_controller.go
```diff
@@ -67,6 +67,8 @@ const (
 	managedOCSName                         = "managedocs"
 	storageClusterName                     = "ocs-storagecluster"
 	prometheusName                         = "managed-ocs-prometheus"
+	prometheusServiceName                  = "prometheus-operated"
```
nb-ohad, line 70: is that the actual name -> "operated"

nb-ohad (changes requested):

nb-ohad:
@aruniiird In addition to the above comments on the code. 
I don't see the needed changes to the prom CR to facilitate TLS communication. How is that done?

controllers/managedocs_controller.go
aruniiird: Done

controllers/managedocs_controller.go
aruniiird: This is changed. Now we directly assign into the map (without any `nil` checking), something like this
```go
r.prometheusService.Annotations = map[string]string{
	servingCertSecretBetaAnnotationKey:  prometheusServiceName,
	servingCertSecretAlphaAnnotationKey: prometheusServiceName,
}
```
The thought behind this was we are reconciling to bring the resource to the expected state and the above is the expected state.

controllers/managedocs_controller.go
aruniiird: Again this part is changed. Now we have our own `Service`.

controllers/managedocs_controller.go
aruniiird: Yes, changed this part (as mentioned above)...

aruniiird:
@nb-ohad , @bindrad , please take another look...

aruniiird:
I'm pretty much blocked by these errors while running testcases,
```
• [FAILED] [3.054 seconds]
ManagedOCS controller
/home/amohan/aruns/programming/go/ocs-osd-deployer-project/src/github.com/red-hat-storage/ocs-osd-deployer/contr
ollers/managedocs_controller_test.go:30
  reconcile() converged mode
  /home/amohan/aruns/programming/go/ocs-osd-deployer-project/src/github.com/red-hat-storage/ocs-osd-deployer/con
trollers/managedocs_controller_test.go:506
    a Grafana datasources secret exists in the openshift-monitoring namespace
    /home/amohan/aruns/programming/go/ocs-osd-deployer-project/src/github.com/red-hat-storage/ocs-osd-deployer/c
ontrollers/managedocs_controller_test.go:1245
      [It] should create k8sMetricsServiceMonitorAuthSecret in primary namespace
      /home/amohan/aruns/programming/go/ocs-osd-deployer-project/src/github.com/red-hat-storage/ocs-osd-deployer
/controllers/managedocs_controller_test.go:1246

  Begin Captured GinkgoWriter Output >>
    2022-04-05T05:47:03.780+0530        INFO    controllers.ManagedOCS  Starting reconcile for ManagedOCS      {
"req.Namespace": "primary", "req.Name": "managedocs"}
    2022-04-05T05:47:03.782+0530        INFO    controllers.ManagedOCS  Non provider deployment, skipping reconc
ile for onboarding validation secret
    2022-04-05T05:47:03.782+0530        INFO    controllers.ManagedOCS  Reconciling StorageCluster
    2022-04-05T05:47:03.782+0530        INFO    controllers.ManagedOCS  Reconciling CSVs
    2022-04-05T05:47:03.782+0530        INFO    controllers.ManagedOCS  Reconciling alertRelabelConfigSecret
    2022-04-05T05:47:03.782+0530        INFO    controllers.ManagedOCS  Reconciling Prometheus Service
    2022-04-05T05:47:03.789+0530        INFO    controllers.ManagedOCS  Reconciling CA Bundle ConfigMap
    2022-04-05T05:47:03.789+0530        INFO    controllers.ManagedOCS  Reconciling Prometheus
    2022-04-05T05:47:03.796+0530        INFO    controllers.ManagedOCS  Reconciling Alertmanager
    2022-04-05T05:47:03.801+0530        ERROR   controllers.ManagedOCS  An error was encountered during reconcil
ePhases {"error": "resourceVersion should not be set on objects to be created"}
    sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
```

Search showed this link: https://github.com/cert-manager/cert-manager/issues/3615 , but this is for a full fledged environment not for a test setup. For test setup, didn't find a place to upgrade the cert-manager version.

aruniiird:
Test issue is resolved (through PR: https://github.com/red-hat-storage/ocs-osd-deployer/pull/164). Re-based on top of latest main (which has the fix).
@nb-ohad , @bindrad , please take a look...

controllers/managedocs_controller.go
```diff
@@ -68,6 +69,11 @@ const (
 	managedOCSName                         = "managedocs"
 	storageClusterName                     = "ocs-storagecluster"
 	prometheusName                         = "managed-ocs-prometheus"
+	prometheusServiceName                  = "managed-ocs-prometheus-service"
+	servingCertSecretBetaAnnotationKey     = "service.beta.openshift.io/serving-cert-secret-name"
+	servingCertSecretAlphaAnnotationKey    = "service.alpha.openshift.io/serving-cert-secret-name"
```
leelavg, line 74: - I don't think we also need alpha key

controllers/managedocs_controller.go
```diff
@@ -68,6 +69,11 @@ const (
...
+	servingCertConfigMapBetaAnnotationKey  = "service.beta.openshift.io/inject-cabundle"
+	servingCertConfigMapAlphaAnnotationKey = "service.alpha.openshift.io/inject-cabundle"
```
leelavg, line 76: - same as above, mostly the services providing this functionality were promoted to beta so beta would suffice, let me know otherwise

controllers/managedocs_controller.go
```diff
@@ -102,6 +108,7 @@ const (
 	convergedDeploymentType                = "converged"
 	consumerDeploymentType                 = "consumer"
 	providerDeploymentType                 = "provider"
+	caBundleConfigMapName                  = "cert-ca-bundle"
```
leelavg, line 111: - a different name would be good to state prometheus is using this bundle
- not a hard req, good to have IMHO

controllers/managedocs_controller.go
```diff
@@ -913,6 +977,42 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
 	return nil
 }
 
+func getPrometheusWebSpec() *promv1.WebSpec {
```
leelavg, line 980: - just curious why can't the function body be stored in templates/prometheus.go?
- is it due to cross pkg vars?

controllers/managedocs_controller.go
aruniiird: There was a discussion on old api-s and backward compatibilities. @nb-ohad , can you comment on this

controllers/managedocs_controller.go
aruniiird: when you say `name` did you mean the variable name `caBundleConfigMapName` or the value part `"cert-ca-bundle"`?

controllers/managedocs_controller.go
aruniiird: Yes, you are right. We are referring to the names of  package `controllers` and don't want to have a cyclic reference in the `templates` package by adding this function there.

controllers/managedocs_controller.go
leelavg: value.

controllers/managedocs_controller.go
nb-ohad: This is a bad practice as you are overriding whatever other annotation any other actor/user will add to the service.
Please use the suggestion from the last comment

controllers/managedocs_controller.go
nb-ohad: Why do we need a secondary service if there is one that already exists? Or am I missing something?

controllers/managedocs_controller.go
nb-ohad: Same question as of the last comment: "Why do we need a secondary service if there is one that already exists? Or am I missing something?"

controllers/managedocs_controller.go
nb-ohad: Same question as of the last comment: "Why do we need a secondary service if there is one that already exists? Or am I missing something?"

controllers/managedocs_controller.go
nb-ohad: Please keep both the alpha and beta key. There is no harm in doing so and we get better coverage

controllers/managedocs_controller.go
nb-ohad: same answer as above

controllers/managedocs_controller.go
```diff
@@ -913,6 +977,42 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+			KeySecret: keySelector,
+			Cert: promv1.SecretOrConfigMap{
+				Secret: &certSelector,
+			},
+			ClientAuthType: fmt.Sprint(tls.VerifyClientCertIfGiven),
+			ClientCA: promv1.SecretOrConfigMap{
+				ConfigMap: &caBundleSelector,
+			},
+		},
+	}
```
nb-ohad, line 1011: Most of these are constant values, pelase move into the prom template. 
And in the same breath, there is no need for `getPrometheusWebSpec`

templates/prometheus.go
```diff
@@ -47,3 +48,16 @@ var PrometheusTemplate = promv1.Prometheus{
 		Resources: utils.GetResourceRequirements("prometheus"),
 	},
 }
+
+var PrometheusServiceTemplate = corev1.Service{
+	Spec: corev1.ServiceSpec{
+		Ports: []corev1.ServicePort{
+			{
+				Name:       "managed-prometheus",
```
nb-ohad, line 56: As this is our own service, let's name it with a name that gives more context, for example: `odf-prometheus-endpoint` or `odf-prometheus-service`

controllers/managedocs_controller_test.go
```diff
@@ -730,6 +742,10 @@ var _ = Describe("ManagedOCS controller", func() {
 						Expect(sc.Spec.AllowRemoteStorageConsumers).Should(BeTrue())
 					}
 
+					By("Creating a prometheus service resource")
+					utils.WaitForResource(k8sClient, ctx, promServiceTemplate.DeepCopy(), timeout, interval)
+					By("Creating CA bundle configmap")
```
nb-ohad, line 747: Same comment bout the CS bundle name

controllers/managedocs_controller.go
```diff
@@ -885,6 +909,45 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.prometheusService, func() error {
+		if err := r.own(r.prometheusService); err != nil {
+			return err
+		}
+
+		desired := templates.PrometheusServiceTemplate.DeepCopy()
+		r.prometheusService.Spec.Ports = desired.Spec.Ports
+		r.prometheusService.Spec.Selector = map[string]string{
+			monLabelKey: monLabelValue,
+		}
```
nb-ohad, line 926: Is there a reason why we are not just forcing the entire template's spec and then mutating it?

```suggestion
		desired := templates.PrometheusServiceTemplate.DeepCopy()
		desired.Spec.Selector = map[string]string{
			monLabelKey: monLabelValue,
		}
		r.prometheusService.Spec = desired.Spec		
```

controllers/managedocs_controller.go
```diff
@@ -130,6 +137,7 @@ type ManagedOCSReconciler struct {
 	cephIngressNetworkPolicy           *netv1.NetworkPolicy
 	providerAPIServerNetworkPolicy     *netv1.NetworkPolicy
 	prometheus                         *promv1.Prometheus
+	prometheusService                  *corev1.Service
```
nb-ohad, line 140: As this is the first owned resource of type `Service` we are reconciling, you need to set up an owner watch on this type at the `SetupWithManager` function. It should look something like this:
```Owns(&orev1.Service{})```

**Review by nb-ohad — CHANGES_REQUESTED** [review]

File: controllers/managedocs_controller.go (hunk 11 above)
Comment (aruniiird) [reply controllers/managedocs_controller.go:1011]: We should not do that. These are constants which are in package `controllers` and if we move `getPrometheusWebSpec()` function into `templates` package we have to import `controllers` package which then creates a cyclic dep, which is a bad practice.
If we create a separate package for the common vars and constants (say `defaults`), we can do this.

Comment (nb-ohad) [reply controllers/managedocs_controller.go:1011]: The controller pkg has visibility to all public constants of the template package. 
Just more the constants there

We already did that for other templates in the past

File: controllers/managedocs_controller.go (hunk 14 above)
Comment (aruniiird) [reply controllers/managedocs_controller.go:926]: Yes. As Service`s spec contains other items like `ClusterIP` etc, which we should not reset. So we are only reset the `Ports` part of the Spec.
Previously the code was as you suggested, which brought lot of test failures (like `spec.ClusterIP` is an immutable var and should not be set)

Comment (nb-ohad) [reply controllers/managedocs_controller.go:926]: So in that case there is no need for a template at all. Please remove hard code into the reconcile function

Comment (nb-ohad) [reply controllers/managedocs_controller.go:926]: Or alternatively, override the spec and than copy back the stuff that does not need to be overriden. 
We already used that pattern once in this controller

File: controllers/managedocs_controller_test.go (hunk 13 above)
Comment (aruniiird) [reply controllers/managedocs_controller_test.go]: Done

File: templates/prometheus.go (hunk 12 above)
Comment (aruniiird) [reply templates/prometheus.go]: Done. Changed it to `odf-prometheus-endpoint`

File: controllers/managedocs_controller.go [hunk 16]
```diff
@@ -885,6 +909,45 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+func (r *ManagedOCSReconciler) reconcileCABundleConfigMap() error {
+	r.Log.Info("Reconciling CA Bundle ConfigMap")
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.caBundleConfigMap, func() error {
+		if err := r.own(r.caBundleConfigMap); err != nil {
+			return err
+		}
+		r.caBundleConfigMap.Annotations = map[string]string{
+			servingCertConfigMapBetaAnnotationKey:  "true",
+			servingCertConfigMapAlphaAnnotationKey: "true",
+		}
```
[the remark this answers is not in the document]
aruniiird: Done

controllers/managedocs_controller.go
```diff
@@ -885,6 +909,45 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+			servingCertSecretBetaAnnotationKey:  prometheusServiceName,
+			servingCertSecretAlphaAnnotationKey: prometheusServiceName,
+		}
+		return nil
+	})
+	return err
+}
+
+func (r *ManagedOCSReconciler) reconcileCABundleConfigMap() error {
+	r.Log.Info("Reconciling CA Bundle ConfigMap")
```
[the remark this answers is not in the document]
aruniiird: Done

controllers/managedocs_controller.go
```diff
@@ -885,6 +909,45 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+
+		desired := templates.PrometheusServiceTemplate.DeepCopy()
+		r.prometheusService.Spec.Ports = desired.Spec.Ports
+		r.prometheusService.Spec.Selector = map[string]string{
+			monLabelKey: monLabelValue,
+		}
+		r.prometheusService.Annotations = map[string]string{
+			servingCertSecretBetaAnnotationKey:  prometheusServiceName,
+			servingCertSecretAlphaAnnotationKey: prometheusServiceName,
+		}
```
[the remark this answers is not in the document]
aruniiird: Done

controllers/managedocs_controller.go
```diff
@@ -143,6 +151,7 @@ type ManagedOCSReconciler struct {
 	reconcileStrategy                  v1.ReconcileStrategy
 	addonParams                        map[string]string
 	onboardingValidationKeySecret      *corev1.Secret
+	caBundleConfigMap                  *corev1.ConfigMap
```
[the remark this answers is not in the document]
aruniiird: Done

controllers/managedocs_controller.go
aruniiird: Done

controllers/managedocs_controller.go
```diff
@@ -183,6 +193,14 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		),
 	)
+	servicePredicate := builder.WithPredicates(
+		predicate.NewPredicateFuncs(
+			func(client client.Object) bool {
+				name := client.GetName()
+				return name == prometheusServiceName
+			},
+		),
+	)
```
nb-ohad, line 203: You don't need this, the `Owns` below will listen to any service which has a controller reference (for this controller) on it. Manual watches (with selectors/predicates) are just for non-owned resources.

controllers/managedocs_controller.go
```diff
@@ -296,6 +315,11 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			enqueueManangedOCSRequest,
 			csvPredicates,
 		).
+		Watches(
+			&source.Kind{Type: &corev1.Service{}},
+			enqueueManangedOCSRequest,
+			servicePredicate,
+		).
```
nb-ohad, line 322: You don't need a manual watch, the `Owns` watch will cover this resource. I will suggest you should take a look at the implementation of `Owns` for better understanding

controllers/managedocs_controller.go
```diff
@@ -885,6 +923,47 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+	r.Log.Info("Reconciling serving cert CA Bundle ConfigMap")
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.servingCertCaBundleConfigMap, func() error {
+		if err := r.own(r.servingCertCaBundleConfigMap); err != nil {
+			return err
+		}
+		if r.servingCertCaBundleConfigMap.Annotations == nil {
+			r.servingCertCaBundleConfigMap.Annotations = make(map[string]string)
+		}
+		r.servingCertCaBundleConfigMap.Annotations[servingCertConfigMapBetaAnnotationKey] = "true"
+		r.servingCertCaBundleConfigMap.Annotations[servingCertConfigMapAlphaAnnotationKey] = "true"
```
nb-ohad, line 961: This pattern is starting to repeat itself many times all over the reconciler process. Can you please add a util function, similar to https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/utils/utils.go#L46-L53 in the utils file and use it instead?

controllers/managedocs_controller.go
```diff
@@ -885,6 +923,47 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+		}
+		r.prometheusService.Annotations[servingCertSecretBetaAnnotationKey] = prometheusServiceName
+		r.prometheusService.Annotations[servingCertSecretAlphaAnnotationKey] = prometheusServiceName
+		return nil
+	})
+	return err
+}
+
+func (r *ManagedOCSReconciler) reconcileCABundleConfigMap() error {
+	r.Log.Info("Reconciling serving cert CA Bundle ConfigMap")
```
nb-ohad, line 952: Letter capitalization and "reconciling" messages should be consistent within the reconciler.
```suggestion
	r.Log.Info("Reconciling servingCertCABundleConfigMap")
```

controllers/managedocs_controller.go
```diff
@@ -885,6 +923,47 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+			r.prometheusService.Annotations = make(map[string]string)
+		}
+		r.prometheusService.Annotations[servingCertSecretBetaAnnotationKey] = prometheusServiceName
+		r.prometheusService.Annotations[servingCertSecretAlphaAnnotationKey] = prometheusServiceName
+		return nil
+	})
+	return err
+}
+
+func (r *ManagedOCSReconciler) reconcileCABundleConfigMap() error {
```
nb-ohad, line 951: Function name is still not fixed 
```suggestion
func (r *ManagedOCSReconciler) reconcileServingCertCABundleConfigMap() error {
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
aruniiird: Moved the `getPrometheusWebSpec()` to templates.
Changed it to `PrometheusWebSpecTemplate`.

controllers/managedocs_controller.go
aruniiird: No templates for Service then...

controllers/managedocs_controller.go
aruniiird: I changed it as said, but one question. In this way it will reconcile for any service being created (doesn't matter what name the service has) in the cluster, right?. This goes same with secrets as well (as we own secrets also).
Don't we have to say that reconcile for Service with this particular name (as we know that is Secret we created) and we don't have to reconcile for a user created secret (as we don't take any action with XYZ secret)

controllers/managedocs_controller.go
aruniiird: Done

controllers/managedocs_controller.go
aruniiird: Done

controllers/managedocs_controller.go
aruniiird: Done

controllers/managedocs_controller.go
aruniiird: Removed

controllers/managedocs_controller.go
```diff
@@ -885,6 +910,52 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+
+		r.prometheusService.Spec.Ports = []corev1.ServicePort{
+			{
+				Name:       "odf-prometheus-endpoint",
+				Protocol:   corev1.ProtocolTCP,
+				Port:       9339,
+				TargetPort: intstr.FromString("web"),
+			},
+		}
+
```
nb-ohad, line 931:
```suggestion
```

controllers/managedocs_controller.go
nb-ohad: > I changed it as said, but one question. In this way it will reconcile for any service being created (doesn't matter what name the service has) in the cluster, right?.

No that is not correct. It will only reconcile for services that have an owner reference for the CR managed by this controller, and the only one that matches this condition is the service CR you have added.

templates/prometheus.go
```diff
@@ -45,5 +53,26 @@ var PrometheusTemplate = promv1.Prometheus{
...
+			LocalObjectReference: locSecretRef,
+			Key:                  "tls.key",
+		},
+		Cert: promv1.SecretOrConfigMap{
+			Secret: &corev1.SecretKeySelector{
+				LocalObjectReference: locSecretRef,
+				Key:                  "tls.crt",
+			},
+		},
+		ClientAuthType: fmt.Sprint(tls.VerifyClientCertIfGiven),
```
nb-ohad, line 76: No need for a second template, this is just a sub-struct under the prom spec, please embed it there (currently you create a copy for no reason)

templates/prometheus.go
```diff
@@ -45,5 +53,26 @@ var PrometheusTemplate = promv1.Prometheus{
 			}},
 		},
 		Resources: utils.GetResourceRequirements("prometheus"),
+		Web:       PrometheusWebSpecTemplate.DeepCopy(),
+	},
+}
+
+var locSecretRef = corev1.LocalObjectReference{
+	Name: PrometheusServiceSecretName,
+}
```
nb-ohad, line 62: Shared instances are always tricky, you change in one location, and all location changes. 
Let's remove this and create a separate `LocalObjectReferance` object at each location.

templates/prometheus.go
```diff
@@ -45,5 +53,26 @@ var PrometheusTemplate = promv1.Prometheus{
...
+
+var PrometheusWebSpecTemplate = &promv1.WebSpec{
+	TLSConfig: &promv1.WebTLSConfig{
+		KeySecret: corev1.SecretKeySelector{
+			LocalObjectReference: locSecretRef,
```
nb-ohad, line 67:
```suggestion
			LocalObjectReference: corev1.LocalObjectReference{Name: PrometheusServiceSecretName}
```
Or even better 
```suggestion
			LocalObjectReference: corev1.LocalObjectReference{Name: ""}
```
Then in the reconcile function, you can do
```
r.prometheus.Spec.Web.TLSConfig.KeySecret.LocalObjectReference.Name = PrometheusServiceSecretName
```

Which will allow you to keep `PrometheusServiceSecretName` on the controller where it makes more sense.

But either option is ok.

templates/prometheus.go
```diff
@@ -45,5 +53,26 @@ var PrometheusTemplate = promv1.Prometheus{
...
+			Key:                  "tls.key",
+		},
+		Cert: promv1.SecretOrConfigMap{
+			Secret: &corev1.SecretKeySelector{
+				LocalObjectReference: locSecretRef,
```
nb-ohad, line 72: Same comment as the prev one

templates/prometheus.go
```diff
@@ -45,5 +53,26 @@ var PrometheusTemplate = promv1.Prometheus{
...
+				Key:                  "tls.crt",
+			},
+		},
+		ClientAuthType: fmt.Sprint(tls.VerifyClientCertIfGiven),
 	},
```
nb-ohad, line 77: This should also go into the template:
```suggestion
		ClientCA: promv1.SecretOrConfigMap{
			ConfigMap: &corev1.ConfigMapKeySelector{
				LocalObjectReference: corev1.LocalObjectReference{Name: ""},
				Key: "service-ca.crt",
			},
		},
	},
```

controllers/managedocs_controller.go
```diff
@@ -904,6 +975,15 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
 		}
 
+		r.prometheus.Spec.Web.TLSConfig.ClientCA = promv1.SecretOrConfigMap{
+			ConfigMap: &corev1.ConfigMapKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: servingCertCABundleConfigMapName,
+				},
+				Key: "service-ca.crt",
+			},
+		}
```
nb-ohad, line 985: The main idea behind the templates is that the templates define the entire structure of the desired resource and include all of the data that is constant outside of any specific context. 

In accordance with that, this entire structure should go into the template (to be cloned when we clone the template using `DeepCopy`) and replaced with 
```
r.prometheus.Spec.Web.TLSConfig.ClientCA.ConfigMap.LocalObjectReference.Name = servingCertCABundleConfigMapName
```

nb-ohad (changes requested):

templates/prometheus.go
aruniiird: done

templates/prometheus.go
aruniiird: Done

templates/prometheus.go
aruniiird: Chose the second option (`Or even better` part), but didn't add any `LocalObjectReference` part since it is not a reference/pointer Go should create an empty defaults into it.
Added an assertion in the test to make sure all the values we set are properly propagated to prometheus instance.

templates/prometheus.go
aruniiird: Removed

templates/prometheus.go
aruniiird: Agreed

templates/prometheus.go
nb-ohad: Good observation about not being a pointer, make the template even simpler 💯

controllers/managedocs_controller_test.go
```diff
@@ -494,6 +506,18 @@ var _ = Describe("ManagedOCS controller", func() {
...
+	assertPrometheusTLSConfigValues := func(ctx context.Context,
+		client client.Client) {
+		prom := promTemplate.DeepCopy()
+		Expect(client.Get(ctx, utils.GetResourceKey(prom), prom)).Should(Succeed())
+		Expect(prom.Spec.Web.TLSConfig.KeySecret.LocalObjectReference.Name == prometheusServiceSecretName).Should(BeTrue())
+		Expect(prom.Spec.Web.TLSConfig.Cert.Secret.LocalObjectReference.Name == prometheusServiceSecretName).Should((BeTrue()))
+		Expect(prom.Spec.Web.TLSConfig.ClientCA.ConfigMap.LocalObjectReference.Name == servingCertCABundleConfigMapName).Should(BeTrue())
+		Expect(prom.Spec.Web.TLSConfig.ClientCA.ConfigMap.Key == "service-ca.crt").Should(BeTrue())
+		Expect(prom.Spec.Web.TLSConfig.KeySecret.Key == "tls.key").Should(BeTrue())
+		Expect(prom.Spec.Web.TLSConfig.Cert.Secret.Key == "tls.crt").Should(BeTrue())
```
nb-ohad, line 518: Please use `.Should(Equal(<value>))` instead of `==` and `Should(BeTrue())

For example:
```
Expect(prom.Spec.Web.TLSConfig.KeySecret.LocalObjectReference.Name).Should(Equal(prometheusServiceSecretName))
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
aruniiird: Changed

controllers/managedocs_controller_test.go
aruniiird: Done

controllers/managedocs_controller.go
```diff
@@ -885,6 +911,52 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+
+		r.prometheusService.Spec.Ports = []corev1.ServicePort{
+			{
+				Name:       "odf-prometheus-endpoint",
+				Protocol:   corev1.ProtocolTCP,
+				Port:       9339,
+				TargetPort: intstr.FromString("web"),
+			},
+		}
+
```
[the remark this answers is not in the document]
aruniiird: Done

controllers/managedocs_controller.go
```diff
@@ -904,6 +976,10 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
 			Key: alertRelabelConfigSecretKey,
 		}
 
+		r.prometheus.Spec.Web.TLSConfig.KeySecret.LocalObjectReference.Name = prometheusServiceSecretName
+		r.prometheus.Spec.Web.TLSConfig.Cert.Secret.LocalObjectReference.Name = prometheusServiceSecretName
+		r.prometheus.Spec.Web.TLSConfig.ClientCA.ConfigMap.LocalObjectReference.Name = servingCertCABundleConfigMapName
```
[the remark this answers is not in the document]
aruniiird: Done

leelavg (approved):
lgtm

nb-ohad (approved):

PR #152, opened 2022-03-30
Uninstall will wait untill all PVs using OCS storageClasses are removed

bindrad:

### DIFF
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -192,38 +192,49 @@ var _ = Describe("ManagedOCS controller", func() {
 			Namespace: testPrimaryNamespace,
 		},
 	}
-	pvc1StorageClassName := storageClassRbdName
-	pvc1Template := corev1.PersistentVolumeClaim{
+	pv1StorageClassName := storageClassRbdName
+	pv1Template := corev1.PersistentVolume{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "test-pvc-1",
+			Name:      "test-pv-1",
 			Namespace: testPrimaryNamespace,
 		},
-		Spec: corev1.PersistentVolumeClaimSpec{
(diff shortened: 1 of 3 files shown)


controllers/managedocs_controller_test.go
```diff
@@ -192,38 +192,49 @@ var _ = Describe("ManagedOCS controller", func() {
...
-					corev1.ResourceStorage: resource.MustParse("1Gi"),
+			Capacity: corev1.ResourceList{
+				corev1.ResourceStorage: resource.MustParse("1Gi"),
+			},
+			PersistentVolumeSource: corev1.PersistentVolumeSource{
+				RBD: &corev1.RBDPersistentVolumeSource{
+					CephMonitors: []string{
+						"0.0.0.0:6789",
+					},
+					RBDImage: "test",
```
Madhu-1, line 214: the source should be CSI, not RBD nor Cephfs.

bindrad: But this is just for our unit tests, we just need to make sure that the PVs are using OCS storage classes

nb-ohad: I agree with Dhruv on this one, our unit tests are mocking the stoagecluster and as such we have no notion of CSI.

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
Madhu-1: >But this is just for our unit tests, we just need to make sure that the PVs are using OCS storage classes

there is no relationship between the OCS storageclasses and the PersistentVolumeSource which is RBD and Cephfs mentioned here.

bindrad: Yes, I agree but we just want PVs using OCS storageClasses
It doesn't matter what PersistentVolumeSource we use

PR #153, opened 2022-03-30
Bumpup prometheus operator version 0.49

aruniiird:
This version has an updated prometheus spec which supports TLS configuration.


leelavg:
> Also, why are we removing our ginkgo and tail vendors
- I think `ginkgo` old version is removed as we already updated to v2
- maybe `tail` wasn't a dependency for prometheus v0.49.0

leelavg:
@aruniiird can this PR be prioritised first? rationale being
- the other PR https://github.com/red-hat-storage/ocs-osd-deployer/pull/149 carries too much of diff and making unittests pass there will be relatively not straightforward as you may need to create dummy service and configmap
- however, this PR is a straight version bump up and shouldn't create regression in tests
- as part of https://github.com/red-hat-storage/ocs-osd-deployer/pull/129 ginkgo v.2 is added but surprised to see ginkgo v1.6 hanging around in the go.mod & go.sum files so better concentrate on having this in at the earliest, thanks.

aruniiird:
> @aruniiird Are the unit tests passing? Also, why are we removing our `ginkgo` and `tail` vendors? It is used in our testing framework

I. About the ginkgo and tail removals, 

There is a PR: https://github.com/red-hat-storage/ocs-osd-deployer/pull/129, which updates the ginkgo version to v2, but AFAIU we may not have done a vendor update.

To test this theory, checkout the following in your local repo

  a. switch to the main branch and update it to the latest code base
  b. run `go mod tidy` and `go mod vendor`

if everything is properly updated, then we should not see any change in the main branch.
PS: please run the `make test` if you see changes in main branch

II. About Unit tests
Working on the unit test failures, but I'm not sure whether this is because of this PR or this is because of ginkgo version updates

bindrad:
@aruniiird The test is passing for this PR at current commit and test for main branch is passing as well

aruniiird:
Thanks @bindrad ...
@nb-ohad , just confirming that the `go mod tidy` and `go mod vendor` in main branch will also give this many file changes/deletions.

aruniiird:
Based this PR on top of PR:https://github.com/red-hat-storage/ocs-osd-deployer/pull/159
Now tests are all passing.

nb-ohad:
@aruniiird The fix on #159 is just creating a sleep which somehow (on your machine) is helping in passing the unit tests, but does not solve the underlying issue of why the tests are broken in the first place. 

@bindrad I am not sure if we should merge this PR in this state, wdyt?

PR #154, opened 2022-03-30
Support a storage unit addon parameter for consumer mode deployment

nb-ohad:


bindrad:
> Agreed but this PR is about the addition of the unit addon param not about the size addon param
So let's defer this request to another PR

Even I agree

PR #162, opened 2022-04-05
Editing scorecard for ms release pipeline fix

akshayashu:
Our 'bundle/test/scorecard/config.yaml' expects another field under label that is phase: msp-main. This is a possible fix for our ms-release pipeline.

Signed-off-by: akkhanna akkhanna@redhat.com


nb-ohad:
@akshayashu Why are we mentioning MSP in the upstream project?

akshayashu:
> @akshayashu Why are we mentioning MSP in the upstream project? I believe this is only relevant for downstream

You're right, we must configure it downstream by editing config.yaml. Closing this MR

PR #163, opened 2022-04-06
update prometheus shim-crds to match v0.49.0

leelavg:
- taken from https://github.com/prometheus-operator/prometheus-operator/tree/release-0.49/example/prometheus-operator-crd


leelavg:
- this should be straight forward verification via unit-tests
- however as unit-tests for me were failing from sometime if @bindrad can run tests once that'd be super helpful and so marked this PR as draft

PR #164, opened 2022-04-06
refactor and fix unit tests

leelavg:
- updated crds for unit tests
- delete crds from etcd after test stop

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -18,6 +18,8 @@ package controllers
 
 import (
 	"context"
+	"io"
+	"os"
 	"path/filepath"
 	"testing"
 
@@ -27,6 +29,7 @@ import (
 	opv1a1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	promv1a1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1alpha1"
+	odfv1a1 "github.com/red-hat-data-services/odf-operator/api/v1alpha1"
 	ocsv1 "github.com/red-hat-storage/ocs-operator/api/v1"
 	v1 "github.com/red-hat-storage/ocs-osd-deployer/api/v1alpha1"
 	appsv1 "k8s.io/api/apps/v1"
@@ -45,10 +48,14 @@ import (
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
 // http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
 
-var cfg *rest.Config
-var k8sClient client.Client
-var testEnv *envtest.Environment
-var testReconciler *ManagedOCSReconciler
+var (
+	cfg            *rest.Config
+	k8sClient      client.Client
+	testEnv        *envtest.Environment
+	testReconciler *ManagedOCSReconciler
+	ctx            context.Context
+	cancel         context.CancelFunc
+)
 
 const (
 	testPrimaryNamespace                       = "primary"
@@ -74,15 +81,30 @@ func TestAPIs(t *testing.T) {
 
 var _ = BeforeSuite(func() {
 	done := make(chan interface{})
-	go func() {
 
+	// write logs from reconciler to a log file
+	var logFile io.Writer
+	logFile, err := os.Create("/tmp/ocs-osd-deployer.log")
+	Expect(err).ToNot(HaveOccurred())
+
+	go func(logFile io.Writer) {
+
+		GinkgoWriter.TeeTo(logFile)
 		logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+
+		ctx, cancel = context.WithCancel(context.Background())
 
 		By("bootstrapping test environment")
 		testEnv = &envtest.Environment{
 			CRDDirectoryPaths: []string{
 				filepath.Join("..", "config", "crd", "bases"),
 				filepath.Join("..", "shim", "crds"),
+			},
+			// when tests are re-run even if CRDs are updated, testEnv will only check
+			// for CRD existence and so it's better to delete those from etcd after
+			// testEnv cleanup
+			CRDInstallOptions: envtest.CRDInstallOptions{
+				CleanUpAfterUse: true,
 			},
 		}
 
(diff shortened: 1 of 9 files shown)


controllers/managedocs_controller.go
```diff
@@ -1110,6 +1119,8 @@ func (r *ManagedOCSReconciler) reconcileK8SMetricsServiceMonitorAuthSecret() err
 		var err error = nil
 		if auth, err = r.readGrafanaV1Secret(); err != nil {
 			if errors.IsNotFound(err) {
+				// Log the error irrespective of readGrafanaV2Secret result
+				r.Log.Error(err, "Unable to find grafana-datasources secret")
```
nb-ohad, line 1123: This is not an error by definition. It is ok to log but lets keep that as an `info`

controllers/managedocs_controller.go
```diff
@@ -428,8 +428,17 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
 			r.Log.Info("deleting OCS CSV")
 			if csv, err := r.getCSVByPrefix(ocsOperatorName); err == nil {
-				if err := r.delete(csv); err != nil {
-					return ctrl.Result{}, fmt.Errorf("unable to delete csv: %v", err)
+				// There's a possibility that csv will be deleted in previous reconcile and
+				// getCSVByPrefix will return emtpy csv in those cases, so need to check for value
+				// ex: setting a finalizer in next step might fail due to resourceVersion changes
+				// IOW, there might be resource modification contention, if we aren't checking for
+				// csv value then err will be returned everytime leading to continued reconcile loops
+				if csv != nil {
```
nb-ohad, line 436: I agree with the comment above, but I think the fix needs to go inside `getCSVByPrefix`. if at the end of the loop no CSV was found then we should return a NotFound error to match the behavior of the k8s client `get` operation

controllers/managedocs_controller.go
```diff
@@ -1618,10 +1629,17 @@ func (r *ManagedOCSReconciler) removeOLMComponents() error {
 
 	r.Log.Info("deleting deployer csv")
 	if csv, err := r.getCSVByPrefix(deployerCSVPrefix); err == nil {
-		if err := r.delete(csv); err != nil {
-			return fmt.Errorf("Unable to delete csv: %v", err)
+		// There's a possibility that csv will be deleted in previous reconcile and
+		// getCSVByPrefix will return emtpy csv in those cases, so need to check for value
+		if csv != nil {
```
nb-ohad, line 1634: Same as the comment above about `getCSVByPrefix`

nb-ohad (changes requested):

controllers/managedocs_controller.go
leelavg: ack, then I'll need to check against the error right?

nb-ohad: We are already doing it in the original code

leelavg:
- will raise separate pr for `getCSVByPrefix` fix

controllers/managedocs_controller.go
leelavg: addressed in https://github.com/red-hat-storage/ocs-osd-deployer/pull/165

controllers/managedocs_controller.go
leelavg: same as above

controllers/suite_test.go
```diff
@@ -74,17 +81,30 @@ func TestAPIs(t *testing.T) {
 
 var _ = BeforeSuite(func() {
 	done := make(chan interface{})
-	go func() {
 
+	// log msgs from reconciler to a log file
+	var f io.Writer
+	f, err := os.Create("/tmp/ocs-osd-deployer.log")
+	Expect(err).ToNot(HaveOccurred())
```
nb-ohad, line 88: Please take a look at the ginkgo `-v` option, that according to docs should stream all of the writer output to stdout

controllers/suite_test.go
```diff
@@ -74,17 +81,30 @@ func TestAPIs(t *testing.T) {
 
 var _ = BeforeSuite(func() {
 	done := make(chan interface{})
-	go func() {
 
+	// log msgs from reconciler to a log file
+	var f io.Writer
+	f, err := os.Create("/tmp/ocs-osd-deployer.log")
```
nb-ohad, line 87: If we decide to go with this implementation, let's add a timestamp to the file name, 
`/tmp/ocs-osd-deployer-YYYYMMDD`

nb-ohad (changes requested):

controllers/suite_test.go
leelavg: - Unfortunately I don't see any log o/p in stdout with this change 
``` diff
# gd Makefile 
diff --git a/Makefile b/Makefile
index 0b6646a..c61f7b3 100644
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ ENVTEST_ASSETS_DIR = $(shell pwd)/testbin
 test: generate fmt vet manifests
        mkdir -p $(ENVTEST_ASSETS_DIR)
        test -f $(ENVTEST_ASSETS_DIR)/setup-envtest.sh || curl -sSLo $(ENVTEST_ASSETS_DIR)/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/v0.6.3/hack/setup-envtest.sh
-       source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out
+       source $(ENVTEST_ASSETS_DIR)/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); go test ./... -coverprofile cover.out -ginkgo.v
 
 # Build manager binary
 manager: generate fmt vet```

File: controllers/suite_test.go (hunk 5 above)
Comment (leelavg) [reply controllers/suite_test.go:87]: - May I know the reason behind it? I don't think we need to have timestamped logs, if need may arise we can rename the file manually

File: controllers/suite_test.go (hunk 4 above)
Comment (nb-ohad) [reply controllers/suite_test.go]: Ok, so we will go with your original method

File: controllers/suite_test.go [hunk 6]
```diff
@@ -74,16 +81,31 @@ func TestAPIs(t *testing.T) {
...
+	var logFile io.Writer
+	logFile, err := os.Create("/tmp/ocs-osd-deployer.log")
+	Expect(err).ToNot(HaveOccurred())
+
+	go func(logFile io.Writer) {
+
+		GinkgoWriter.TeeTo(logFile)
 		logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
+		ctx, cancel = context.WithCancel(context.TODO())
```
nb-ohad, line 95: Please change to 
```suggestion
		ctx, cancel = context.WithCancel(context.Background())
```

For a full explanation you can take a look here:
https://pkg.go.dev/context#Background

nb-ohad (changes requested):

PR #168, opened 2022-04-13
BZ: 2070898 Added rook  override configs during Consumer storage cluster creation

ezio-auditore:
Added the following options to be created in a ConfigMap before Storage cluster creation.
```
[global]
osd_pool_pg_autoscale_mode = off
osd_pool_default_pg_num = 128
osd_pool_default_pgp_num = 128
```

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -110,6 +110,7 @@ const (
 	providerDeploymentType                 = "provider"
 	servingCertCABundleConfigMapName       = "prometheus-serving-cert-ca-bundle"
 	prometheusServiceSecretName            = "managed-ocs-prometheus-service-secret"
+	rookOverrideConfigMapName              = "rook-config-override"
 )
 
 // ManagedOCSReconciler reconciles a ManagedOCS object
@@ -153,6 +154,7 @@ type ManagedOCSReconciler struct {
 	addonParams                        map[string]string
 	onboardingValidationKeySecret      *corev1.Secret
 	servingCertCaBundleConfigMap       *corev1.ConfigMap
+	rookConfigOverride                 *corev1.ConfigMap
 }
 
 // Add necessary rbac permissions for managedocs finalizer in order to set blockOwnerDeletion.
@@ -433,6 +435,10 @@ func (r *ManagedOCSReconciler) initReconciler(ctx context.Context, req ctrl.Requ
 	r.onboardingValidationKeySecret = &corev1.Secret{}
 	r.onboardingValidationKeySecret.Name = onboardingValidationKeySecretName
 	r.onboardingValidationKeySecret.Namespace = r.namespace
+
+	r.rookConfigOverride = &corev1.ConfigMap{}
+	r.rookConfigOverride.Name = rookOverrideConfigMapName
+	r.rookConfigOverride.Namespace = r.namespace
 }
 
 func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
@@ -512,6 +518,9 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 		}
 		// Reconciles only for provider deployment type
 		if err := r.reconcileOnboardingValidationSecret(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileRookOverrideConfig(); err != nil {
 			return ctrl.Result{}, err
 		}
 		if err := r.reconcileStorageCluster(); err != nil {
@@ -1785,4 +1794,30 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
 		r.Log.V(-1).Info("Requested storage device set count will result in downscaling, which is not supported. Skipping")
 		deviceSet.Count = currDeviceSetCount
 	}
+}
+
+func (r *ManagedOCSReconciler) reconcileRookOverrideConfig() error {
+	if r.DeploymentType != providerDeploymentType {
+		r.Log.Info("Non provider deployment, skipping reconcile for rook override config")
+		return nil
+	}
+	r.Log.Info("Reconciling Rook Override Config Map")
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.rookConfigOverride, func() error {
+		if err := r.own(r.rookConfigOverride); err != nil {
+			return err
+		}
+
+		r.rookConfigOverride.Data = map[string]string{
+			"config": "" +
+				"[global]\n" +
+				"osd_pool_pg_autoscale_mode = off\n" +
+				"osd_pool_default_pg_num = 128\n" +
+				"osd_pool_default_pgp_num = 128\n",
+		}
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Unable to create Rook Override Config Map")
+	}
+	return nil
 }
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -318,6 +318,12 @@ var _ = Describe("ManagedOCS controller", func() {
 			Namespace: testPrimaryNamespace,
 		},
 	}
+	rookOverrideConfigMapTemplate := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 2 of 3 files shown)


controllers/managedocs_controller.go
```diff
@@ -1786,3 +1795,30 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
 		deviceSet.Count = currDeviceSetCount
 	}
 }
+
+func (r *ManagedOCSReconciler) reconcileRookOverrideConfig() error {
+	if r.DeploymentType != consumerDeploymentType {
```
nb-ohad, line 1800: I believe this change needs to go on provider deployment, not consumer deployments.

controllers/managedocs_controller.go
```diff
@@ -1786,3 +1795,30 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		r.Log.Info("Non consumer deployment, skipping reconcile for rook override config")
+		return nil
+	}
+	r.Log.Info("Reconciling Rook Override Config Map")
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.rookConfigOverride, func() error {
```
nb-ohad, line 1805: Isn't the config map already there? if so we cannot use `createOrUpdate`

controllers/managedocs_controller.go
```diff
@@ -1786,3 +1795,30 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		if err := r.own(r.rookConfigOverride); err != nil {
```
nb-ohad, line 1806: If we are not the ones to create it we should not own it.
Please note that if we do not own we will need to add a manual watch for this

controllers/managedocs_controller.go
```diff
@@ -1786,3 +1795,30 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+			return err
+		}
+		configMapData := make(map[string]string, 0)
+		globaldata := `
+		    [global]
+			osd_pool_pg_autoscale_mode = off
+			osd_pool_default_pg_num = 128
+			osd_pool_default_pgp_num = 128
+			`
+		configMapData["config"] = globaldata
```
nb-ohad, line 1816: 1. No need to use make
2. all of the extra spaces (indentations) might create issues

```suggestion
		r.rookConfigOverride.Data := map[string]string {
			"config": `[global]
osd_pool_pg_autoscale_mode = off
osd_pool_default_pg_num = 128
osd_pool_default_pgp_num = 128` ,
		}
```

controllers/managedocs_controller.go
```diff
@@ -507,6 +513,9 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 		}
 
 		// Reconcile the different resources
+		if err := r.reconcileRookOverrideConfig(); err != nil {
+			return ctrl.Result{}, err
+		}
```
nb-ohad, line 518: Is there a reason to add it here?

nb-ohad (changes requested):

nb-ohad:
@ezio-auditore I don't see changes to unit tests to capture the changes

controllers/managedocs_controller.go
ezio-auditore: Done

controllers/managedocs_controller.go
ezio-auditore: We are the ones creating it. Since rook will look on this at clusttr creation and apply the settings.

controllers/managedocs_controller.go
ezio-auditore: Done

ezio-auditore:
Added unit tests as well

controllers/managedocs_controller.go
```diff
@@ -1786,3 +1795,25 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		return nil
+	}
+	r.Log.Info("Reconciling Rook Override Config Map")
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.rookConfigOverride, func() error {
+		if err := r.own(r.rookConfigOverride); err != nil {
+			return err
+		}
+
+		r.rookConfigOverride.Data = map[string]string{
+			"config": "[global]\nosd_pool_pg_autoscale_mode = off\nosd_pool_default_pg_num = 128\nosd_pool_default_pgp_num = 128",
```
nb-ohad, line 1811: This seems problematic to read and change in the future, can we just do the following instead?
```suggestion
			"config": "" +
				"[global]\n" +
				"osd_pool_pg_autoscale_mode = off\n" +
				"osd_pool_default_pg_num = 128\n" +
				"osd_pool_default_pgp_num = 128\n",
```

controllers/managedocs_controller_test.go
```diff
@@ -777,6 +785,17 @@ var _ = Describe("ManagedOCS controller", func() {
...
 			})
+			When("it is a provider deployment", func() {
+				It("should have rook-config-override configmap settings", func() {
+					if testReconciler.DeploymentType != providerDeploymentType {
+						Skip(fmt.Sprintf("Skipping the test as it is not required by %v", testReconciler.DeploymentType))
+					}
+					configMap := rookOverrideConfigMapTemplate.DeepCopy()
+					utils.WaitForResource(k8sClient, ctx, configMap, timeout, interval)
+					fmt.Println(configMap.Data["config"])
+					Expect(configMap.Data["config"]).Should(Equal("[global]\nosd_pool_pg_autoscale_mode = off\nosd_pool_default_pg_num = 128\nosd_pool_default_pgp_num = 128"))
```
nb-ohad, line 796: Same reason as the last comment:
```suggestion
					Expect(configMap.Data["config"]).Should(Equal("" +
						"[global]\n" +
						"osd_pool_pg_autoscale_mode = off\n" +
						"osd_pool_default_pg_num = 128\n" +
						"osd_pool_default_pgp_num = 128\n"))
```

controllers/managedocs_controller_test.go
```diff
@@ -777,6 +785,17 @@ var _ = Describe("ManagedOCS controller", func() {
...
 				})
 			})
+			When("it is a provider deployment", func() {
+				It("should have rook-config-override configmap settings", func() {
+					if testReconciler.DeploymentType != providerDeploymentType {
+						Skip(fmt.Sprintf("Skipping the test as it is not required by %v", testReconciler.DeploymentType))
+					}
+					configMap := rookOverrideConfigMapTemplate.DeepCopy()
+					utils.WaitForResource(k8sClient, ctx, configMap, timeout, interval)
+					fmt.Println(configMap.Data["config"])
```
nb-ohad, line 795: Why are you using `fmt.Println` inside a test?

nb-ohad (changes requested):

controllers/managedocs_controller.go
ezio-auditore: Done

controllers/managedocs_controller_test.go
ezio-auditore: Should'nt be there. Have removed it.

controllers/managedocs_controller_test.go
ezio-auditore: Done

templates/providerstoragecluster.go
```diff
@@ -95,6 +95,9 @@ var ProviderStorageClusterTemplate = ocsv1.StorageCluster{
 			CephFilesystems: ocsv1.ManageCephFilesystems{
 				DisableStorageClass: false,
 			},
+			CephConfig: ocsv1.ManageCephConfig{
+				ReconcileStrategy: "ignore",
```
ezio-auditore, line 99: I am not sure about the side effects of this.But this need to be there.  what do you think @nb-ohad ?

ezio-auditore:
Closing this as this is not a fruitful solution. The config map is managed by ocs-operator and it will be difficult to modify the existing cm and restart all the ceph pods.
Adding it during cluster creation will result in additional configs missing from ocs-operator.

PR #172, opened 2022-04-18
bz#2073297 - Sendgrid PVC utilization mails should contain PVC name

akshayashu:
Sendgrid Alerts only send emails/Alerts mentioning the PVC utilization, but the message must also contain the corresponding PVC name


nb-ohad:
@akshayashu Why do you have 11 commits on this PR?

akshayashu:
I pulled v2.0 for the latest commits, I think that is why all these commits are here too

nb-ohad:
why v2.0 and not main?

akshayashu:
I thought changes would first go into v2.0 and then collectively in main
I'll raise another PR

nb-ohad:
All patches/features are going first into main, then we decided what and when to backport into versioned/release branches

PR #175, opened 2022-04-21
Fix tls prometheus

bindrad:


controllers/managedocs_controller.go
```diff
@@ -913,6 +924,67 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
 
+func (r *ManagedOCSReconciler) reconcileKubeRbacConfigMap() error {
+	r.Log.Info("Reconciling kubeRbacConfigMap")
+
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.kubeRbacConfigMap, func() error {
+		if err := r.own(r.kubeRbacConfigMap); err != nil {
+			return err
+		}
+
+		kubeRbacConfig := struct {
```
leelavg, line 935: - maybe use this directly https://github.com/brancz/kube-rbac-proxy/blob/0475166d4bf55981fa704052fd0d2077df6c75f6/pkg/authz/auth.go#L32-L37 and json marshalling?

controllers/managedocs_controller.go
```diff
@@ -913,6 +924,67 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+	if err != nil {
+		return fmt.Errorf("unable to create kubeRbacConfig config map: %v", err)
+	}
+
+	return nil
+}
+
 // reconcilePrometheusService function wait for prometheus Service
 // to start and sets appropriate annotation for 'service-ca' controller
 func (r *ManagedOCSReconciler) reconcilePrometheusService() error { //nolint:deadcode,unused
```
leelavg, line 990: - the nolint directives can be removed now i think
- pls do the same in remaining places

templates/prometheus.go
```diff
@@ -45,25 +52,50 @@ var PrometheusTemplate = promv1.Prometheus{
...
+		Resources:   utils.GetResourceRequirements("prometheus"),
+		ListenLocal: true,
+		Containers: []corev1.Container{{
+			Name:  "kube-rbac-proxy",
+			Image: "quay.io/brancz/kube-rbac-proxy:v0.12.0",
+			Args: []string{
+				"--secure-listen-address=0.0.0.0:9339",
+				"--upstream=http://127.0.0.1:9090/",
+				"--logtostderr=true",
+				"--v=10",
```
leelavg, line 64: - too much verbosity i guess 😅

templates/prometheus.go
```diff
@@ -45,25 +52,50 @@ var PrometheusTemplate = promv1.Prometheus{
...
-		// 				Key: "service-ca.crt",
-		// 			},
-		// 		},
-		// 	},
-		// },
+		Resources:   utils.GetResourceRequirements("prometheus"),
+		ListenLocal: true,
+		Containers: []corev1.Container{{
+			Name:  "kube-rbac-proxy",
+			Image: "quay.io/brancz/kube-rbac-proxy:v0.12.0",
```
leelavg, line 59: - maybe use an env value to get this image? we can swap this with https://catalog.redhat.com/software/containers/openshift4/ose-kube-rbac-proxy/5cdb2634dd19c778293b4d98 when needed via env

controllers/managedocs_controller_test.go
```diff
@@ -507,19 +506,18 @@ var _ = Describe("ManagedOCS controller", func() {
...
-	// 	Expect(client.Get(ctx, utils.GetResourceKey(prom), prom)).Should(Succeed())
-	// 	tlsConfig := prom.Spec.Web.TLSConfig
-	// 	Expect(tlsConfig.KeySecret.LocalObjectReference.Name).Should(Equal(prometheusServiceSecretName))
-	// 	Expect(tlsConfig.Cert.Secret.LocalObjectReference.Name).Should(Equal(prometheusServiceSecretName))
-	// 	Expect(tlsConfig.ClientCA.ConfigMap.LocalObjectReference.Name).Should(Equal(servingCertCABundleConfigMapName))
-	// 	Expect(tlsConfig.ClientCA.ConfigMap.Key).Should(Equal("service-ca.crt"))
-	// 	Expect(tlsConfig.KeySecret.Key).Should(Equal("tls.key"))
-	// 	Expect(tlsConfig.Cert.Secret.Key).Should(Equal("tls.crt"))
-	// }
+	assertPrometheusTLSConfigValues := func(ctx context.Context,
```
leelavg, line 509: - i don't think this is the correct unit test for the functionality
- pls have a re-look at rest of the unit test pieces added in this pr

templates/prometheus.go
```diff
@@ -19,11 +19,18 @@ package templates
...
 	promv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/red-hat-storage/ocs-osd-deployer/utils"
+	corev1 "k8s.io/api/core/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	ServingCertificaKeyVolumeName  = "serving-cert"
```
leelavg, line 29:
```suggestion
	ServingCertificateKeyVolumeName  = "serving-cert"
```

templates/prometheus.go
```diff
@@ -45,25 +52,50 @@ var PrometheusTemplate = promv1.Prometheus{
...
-		// 		},
-		// 		ClientAuthType: fmt.Sprint(tls.VerifyClientCertIfGiven),
-		// 		ClientCA: promv1.SecretOrConfigMap{
-		// 			ConfigMap: &corev1.ConfigMapKeySelector{
-		// 				Key: "service-ca.crt",
-		// 			},
-		// 		},
-		// 	},
-		// },
+		Resources:   utils.GetResourceRequirements("prometheus"),
```
leelavg, line 55: we can reduce the requirements, however can be part of another pr

controllers/managedocs_controller.go
```diff
@@ -977,11 +1049,33 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
 		}
 
-		// prometheus tls -- will be uncommented
-		// tlsConfig := r.prometheus.Spec.Web.TLSConfig
-		// tlsConfig.KeySecret.LocalObjectReference.Name = prometheusServiceSecretName
-		// tlsConfig.Cert.Secret.LocalObjectReference.Name = prometheusServiceSecretName
-		// tlsConfig.ClientCA.ConfigMap.LocalObjectReference.Name = servingCertCABundleConfigMapName
+		volumes := r.prometheus.Spec.Volumes
+		for key := range volumes {
+			volume := &volumes[key]
```
leelavg, line 1054: - why not just range on volumes and switch on volume.Name rather than multiple conditionals?

leelavg (changes requested):
- looks straightforward, thanks for taking this up!

config/rbac/prom_kube_rbac_proxy_role_binding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: kube-rbac-proxy
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: proxy-role
```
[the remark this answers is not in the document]
bindrad: Yes, I was thinking to create a separate one for this but then I realized kube-rbac-proxy doesn't require anything else so why not use the same role. What do you think?

controllers/managedocs_controller.go
bindrad: Yes, I missed to remove these

templates/prometheus.go
bindrad: Yes, I just didn't want to commit downstream image to upstream repo. But using the env variable will solve the issue.

config/rbac/prom_kube_rbac_proxy_role_binding.yaml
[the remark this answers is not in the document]
leelavg: - fine with me.

controllers/managedocs_controller.go
nb-ohad: +1 to replace the `if else` construct with a switch statement

nb-ohad:
@bindrad Please separate vendor bump to a different PR (to be merged before this one)

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
bindrad: Yes, I didn't actually work on implementing tests for this PR. I'll do it after all the changes are done to the controller.

controllers/managedocs_controller.go
```diff
@@ -925,26 +1008,28 @@ func (r *ManagedOCSReconciler) reconcilePrometheusService() error { //nolint:dea
...
 			},
 		}
 		r.prometheusService.Spec.Selector = map[string]string{
-			monLabelKey: monLabelValue,
+			prometheusServiceSelectorLabelKey: prometheusServiceSelectorLabelValue,
 		}
 		utils.AddAnnotation(r.prometheusService,
 			servingCertSecretBetaAnnotationKey, prometheusServiceSecretName)
 		utils.AddAnnotation(r.prometheusService,
 			servingCertSecretAlphaAnnotationKey, prometheusServiceSecretName)
```
nb-ohad, line 1023: These are known annotations defined outside of our product, so I see no point in hiding them in a constant as they are not going to change throughout the lifetime of the operator.

Also, long lines should be split in accordance with this repo coding style

 ```suggestion
		utils.AddAnnotation(
			r.prometheusService,
			"service.beta.openshift.io/serving-cert-secret-name", 
			prometheusServiceSecretName,
		)
		utils.AddAnnotation(
			r.prometheusService,
			"service.alpha.openshift.io/serving-cert-secret-name", 
			prometheusServiceSecretName,
		)
```

File: controllers/managedocs_controller.go [hunk 11]
```diff
@@ -925,26 +1008,28 @@ func (r *ManagedOCSReconciler) reconcilePrometheusService() error { //nolint:dea
...
+		utils.AddLabel(r.prometheusService,
+			prometheusServiceLabelKey, prometheusServiceLabelValue)
```
nb-ohad, line 1025: No need to invent a new set of labels, please use our existing `monLabelKey`, `monLabelValue` 

```suggestion
		utils.AddLabel(r.prometheusService, monLabelKey, monLabelValue)
```

controllers/managedocs_controller.go
```diff
@@ -925,26 +1008,28 @@ func (r *ManagedOCSReconciler) reconcilePrometheusService() error { //nolint:dea
...
 		return nil
 	})
 	return err
 }
 
-func (r *ManagedOCSReconciler) reconcileServingCertCABundleConfigMap() error { //nolint:deadcode,unused
-	r.Log.Info("Reconciling servingCertCABundleConfigMap")
+func (r *ManagedOCSReconciler) reconcileServingCertCABundleConfigMap() error {
```
nb-ohad, line 1031: The annotation names in this function should be inlined as of the reasoning in the comment above about the service serving cert annotations 

Also, line splitting styling.

controllers/managedocs_controller.go
```diff
@@ -977,16 +1062,45 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+			volume := &volumes[key]
+			switch volume.Name {
+			case templates.ServingCertificateKeyVolumeName:
+				volume.VolumeSource.Secret.SecretName = prometheusServiceSecretName
+			case templates.ServiceCACertificateVolumeName:
+				volume.VolumeSource.ConfigMap.LocalObjectReference.Name = servingCertCABundleConfigMapName
+			case templates.KubeRbacProxyConfigVolumeName:
+				volume.VolumeSource.ConfigMap.LocalObjectReference.Name = kubeRbacConfigMap
+			}
+		}
```
nb-ohad, line 1098: All of this is constant, no need to run code to set this up.
Please move this as static configuration into the Prometheus template

controllers/managedocs_controller.go
```diff
@@ -1754,6 +1885,21 @@ func (r *ManagedOCSReconciler) deleteCSVByPrefix(name string) error {
...
+	}
+
+	for index := range csvList.Items {
+		candidate := &csvList.Items[index]
+		if strings.HasPrefix(candidate.Name, name) {
+			return candidate, nil
+		}
+	}
+	return nil, fmt.Errorf("unable to find the csv prefixed with %s", name)
+}
```
nb-ohad, line 1901: This is an extract from `deleteCSVByPrefix` so please refactor `deleteCSVByPrefix` to use this method internally so we will not need to solve bugs in 2 distinct places.

controllers/managedocs_controller.go
```diff
@@ -977,16 +1062,45 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+			prometheusContainers := r.prometheus.Spec.Containers
+			for key := range prometheusContainers {
+				container := &prometheusContainers[key]
+				if container.Name == "kube-rbac-proxy" {
+					container.Image = kubeRbacImage
+				}
+			}
+		} else {
+			return fmt.Errorf("Unable to set image for kube-rbac-proxy container: %v", err)
+		}
```
nb-ohad, line 1085: Error handling should come first to remove one label of indentation from the large code block.

```suggestion
		if deployerCSV, err := r.getCSVByPrefix(deployerCSVPrefix); err != nil {
			return fmt.Errorf("Unable to set image for kube-rbac-proxy container: %v", err)
		}
		
		deployerCSVDeployment := &deployerCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0]
		deployerCSVContainers := deployerCSVDeployment.Spec.Template.Spec.Containers
		var kubeRbacImage string
		for key := range deployerCSVContainers {
			container := deployerCSVContainers[key]
			if container.Name == "kube-rbac-proxy" {
				kubeRbacImage = container.Image
			}
		}

		prometheusContainers := r.prometheus.Spec.Containers
		for key := range prometheusContainers {
			container := &prometheusContainers[key]
			if container.Name == "kube-rbac-proxy" {
				container.Image = kubeRbacImage
			}
		}
```

controllers/managedocs_controller.go
```diff
@@ -977,16 +1062,45 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+				}
+			}
+
+			prometheusContainers := r.prometheus.Spec.Containers
+			for key := range prometheusContainers {
+				container := &prometheusContainers[key]
+				if container.Name == "kube-rbac-proxy" {
+					container.Image = kubeRbacImage
+				}
+			}
```
nb-ohad, line 1082: If the container already exists why are we forcing a specific kube rbac proxy? Why can't we use the rbac proxy that is already there?

templates/prometheus.go
```diff
@@ -19,11 +19,18 @@ package templates
...
+	corev1 "k8s.io/api/core/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
 
+const (
+	ServingCertificateKeyVolumeName = "serving-cert"
+	ServiceCACertificateVolumeName  = "ca-cert"
+	KubeRbacProxyConfigVolumeName   = "kube-rbac-config"
```
nb-ohad, line 31: Please make these private. 
We should not export if we do not have to.

```suggestion
	servingCertificateKeyVolumeName = "serving-cert"
	serviceCACertificateVolumeName  = "ca-cert"
	kubeRbacProxyConfigVolumeName   = "kube-rbac-config"
```

templates/prometheus.go
```diff
@@ -45,25 +52,62 @@ var PrometheusTemplate = promv1.Prometheus{
...
+			},
+			{
+				Name: KubeRbacProxyConfigVolumeName,
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{},
+					},
+				},
+			},
+		},
```
nb-ohad, line 111: This needs to come out of the box in our prom package, the operator should not concern itself with adding containers into the Prometheus bundle/deployment.

templates/prometheusproxynetworkpolicy.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+var PrometheusProxyPort = intstr.FromInt(9339)
+var PrometheusProxyProtocol = corev1.ProtocolTCP
```
nb-ohad, line 27: grouping + local scoping 
```suggestion
var (
    prometheusProxyPort = intstr.FromInt(9339)
    prometheusProxyProtocol = corev1.ProtocolTCP
)
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
bindrad: We need that label on the service to use the Managed-tenats federation mechanism

nb-ohad: Are they targeting that specific label? Isn't that configurable in the addon manifest?

bindrad: Yes, they are targeting that label and it is not configurable.

nb-ohad: In that case there is no need for constants just add the label inline with a comment above explaining the need for this label on the service

controllers/managedocs_controller.go
```diff
@@ -1731,27 +1820,26 @@ func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
...
 	}
-
-	// There might be a chance that csv was deleted in previous reconcile, so
-	// check for it's existence before deleting
-	if csv != nil {
-		return r.delete(csv)
-	}
-
-	return nil
+	return nil, fmt.Errorf("unable to find the csv prefixed with %s", name)
```
nb-ohad, line 1842:
```suggestion
	return nil, fmt.Errorf("unable to find a csv prefixed with %s", name)
```

templates/prometheus.go
```diff
@@ -45,25 +46,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
-		// },
+		Resources:   utils.GetResourceRequirements("prometheus"),
+		ListenLocal: true,
+		Containers: []corev1.Container{{
+			Name: "kube-rbac-proxy",
+			Args: []string{
+				"--secure-listen-address=0.0.0.0:9339",
+				"--upstream=http://127.0.0.1:9090/",
+				"--logtostderr=true",
+				"--v=10",
```
nb-ohad, line 57: Do we really need `--v=10`? isn't 10 too high?

templates/prometheus.go
```diff
@@ -45,25 +46,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
-		// 			},
-		// 		},
-		// 	},
-		// },
+		Resources:   utils.GetResourceRequirements("prometheus"),
+		ListenLocal: true,
+		Containers: []corev1.Container{{
+			Name: "kube-rbac-proxy",
+			Args: []string{
+				"--secure-listen-address=0.0.0.0:9339",
```
nb-ohad, line 54: The port number is used more than one time in this template and in the service reconciliation. 
Can we please move it into a public constant to be used in all places?

templates/prometheuskuberbacproxyconfig.go
```diff
@@ -0,0 +1,52 @@
...
+				ResourceRequest: false,
+				Verb:            "get",
+			},
+			{
+				Path:            "/federate",
+				ResourceRequest: false,
+				Verb:            "get",
+			},
+		},
+	},
```
nb-ohad, line 51: A couple of notes: 
1. Why are we creating our own struct and using one provided by the lib?
2. Can't we just have a template for the entire config map and not just the config itself?

templates/prometheusproxynetworkpolicy.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+var (
+	prometheusProxyPort     = intstr.FromInt(9339)
```
nb-ohad, line 27: See my prev comment regarding the port number.

controllers/managedocs_controller_test.go
```diff
@@ -757,15 +755,14 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					By("Creating a prometheus service resource")
+					utils.WaitForResource(k8sClient, ctx, promServiceTemplate.DeepCopy(), timeout, interval)
+					By("Creating serving cert CA bundle configmap")
+					utils.WaitForResource(k8sClient, ctx, caBundleConfigMap.DeepCopy(), timeout, interval)
 					By("Creating a prometheus resource")
 					utils.WaitForResource(k8sClient, ctx, promTemplate.DeepCopy(), timeout, interval)
-					// prometheus tls -- will be uncommented
-					// assertPrometheusTLSConfigValues(ctx, k8sClient)
+
+					assertPrometheusTLSConfigValues(ctx, k8sClient)
```
nb-ohad, line 765: No need for a function here. please assert what is needed to be asserted in the body of the test case.

nb-ohad (changes requested):

templates/prometheus.go
nb-ohad: After discussing it with @bindrad I agree we need to add the container in code.

templates/prometheuskuberbacproxyconfig.go
bindrad: We can't use the struct provided by the library because of two reasons
1. ResourceRequest needs to be set to false but when we try to Marshal it ResourceRequest is removed as the default is set to false. Usually, they should have used pointer to the boolean variable but they haven't so if try to set the default value Marshal just removes it.  I'm unclear then please refer to this: https://stackoverflow.com/questions/37756236/json-golang-boolean-omitempty
2. The outermost key-value pair in the json should be authorization but it is not present in the struct they have provided so the container goes in failed state. So basically it is marshalled into `{
      "static":[
        {
          "verb":"get",
          "resourceRequest":false,
          "path":"/metrics"
        },
        {
          "verb":"get",
          "resourceRequest":false,
          "path":"/federate"
        }
      ]
    }`
    whereas it should marshal as `{
      "authorization":
      {
        "static":[
          {
            "verb":"get",
            "resourceRequest":false,
            "path":"/metrics"
          },
          {
            "verb":"get",
            "resourceRequest":false,
            "path":"/federate"
          }
        ]
      }
    }`
    
    I have used the entire configMap in the template file in my latest commit

templates/prometheus.go
bindrad: I used the one used by in-cluster monitoring, I don't have any problem in reducing it but it helps to debug.

controllers/managedocs_controller.go
```diff
@@ -977,16 +1032,33 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
-		// tlsConfig := r.prometheus.Spec.Web.TLSConfig
-		// tlsConfig.KeySecret.LocalObjectReference.Name = prometheusServiceSecretName
-		// tlsConfig.Cert.Secret.LocalObjectReference.Name = prometheusServiceSecretName
-		// tlsConfig.ClientCA.ConfigMap.LocalObjectReference.Name = servingCertCABundleConfigMapName
+		deployerCSV, err := r.getCSVByPrefix(deployerCSVPrefix)
+		if err != nil {
+			return fmt.Errorf("Unable to set image for kube-rbac-proxy container: %v", err)
+		}
+
+		deployerCSVDeployment := &deployerCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0]
```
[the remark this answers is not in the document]
bindrad: Done in the latest commit

controllers/managedocs_controller.go
```diff
@@ -69,11 +69,7 @@ const (
 	managedOCSName                         = "managedocs"
 	storageClusterName                     = "ocs-storagecluster"
 	prometheusName                         = "managed-ocs-prometheus"
-	prometheusServiceName                  = "managed-ocs-prometheus-service"
-	servingCertSecretBetaAnnotationKey     = "service.beta.openshift.io/serving-cert-secret-name"
-	servingCertSecretAlphaAnnotationKey    = "service.alpha.openshift.io/serving-cert-secret-name"
-	servingCertConfigMapBetaAnnotationKey  = "service.beta.openshift.io/inject-cabundle"
-	servingCertConfigMapAlphaAnnotationKey = "service.alpha.openshift.io/inject-cabundle"
+	prometheusServiceName                  = "prometheus"
```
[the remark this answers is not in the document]
bindrad: We can't use any name for this service. If we want to use Managed-tenants tooling for federation then we need "prometheus"[1] as the name of service and "app: prometheus"[2] label on that service(we have configured this in the addon manifest).
[1]: https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/data/selectorsyncset.yaml.j2#L220
[2]: https://github.com/mt-sre/managed-tenants-cli/blob/main/managedtenants/data/selectorsyncset.yaml.j2#L232

controllers/managedocs_controller.go
```diff
@@ -913,10 +927,36 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
 
+func (r *ManagedOCSReconciler) reconcileKubeRBACConfigMap() error {
+	r.Log.Info("Reconciling kubeRBACConfigMap")
+
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.kubeRBACConfigMap, func() error {
+		if err := r.own(r.kubeRBACConfigMap); err != nil {
+			return err
+		}
+
+		config, err := yaml.Marshal(templates.KubeRBACProxyConfig)
```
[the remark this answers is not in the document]
bindrad: Done in latest commit

controllers/managedocs_controller.go
```diff
@@ -913,10 +927,36 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+		if err != nil {
+			panic(fmt.Sprintf("Failed to Marshal kubeRBACConfig: %v", err))
+		}
+
+		r.kubeRBACConfigMap.Data = map[string]string{
+			"config-file.yaml": string(config),
+		}
```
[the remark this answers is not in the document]
bindrad: Done in latest commit

templates/prometheus.go
bindrad: I used the one used by in-cluster monitoring, I don't have any problem in reducing it but it helps to debug.

templates/prometheusproxynetworkpolicy.go
bindrad: Done in latest commit

controllers/managedocs_controller_test.go
bindrad: I haven't worked on test yet so please don't review unit test.

templates/prometheusproxynetworkpolicy.go
bindrad: Done in latest commit

controllers/managedocs_controller.go
```diff
@@ -977,11 +1049,33 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+			} else if volume.Name == templates.KubeRbacProxyConfigVolumeName {
+				volume.VolumeSource = corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: kubeRbacConfigMap,
+						},
+					},
+				}
+			}
+		}
```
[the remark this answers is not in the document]
bindrad: Done in the latest commit

templates/prometheus.go
bindrad: Not using variable anymore for these values and directly adding in the template

controllers/managedocs_controller.go
bindrad: This was a mistake, I intended to add containerImage to the prometheus template and not to the runtime CR.
`desired := templates.PrometheusTemplate.DeepCopy()
prometheusContainers := desired.Spec.Containers`

controllers/managedocs_controller.go
bindrad: Done in latest commit

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: For [2] I see that we choose the labels based on our own addon metadata, so why do we have to use `"app: prometheus"`?

controllers/managedocs_controller.go
```diff
@@ -108,8 +105,9 @@ const (
 	convergedDeploymentType                = "converged"
 	consumerDeploymentType                 = "consumer"
 	providerDeploymentType                 = "provider"
-	servingCertCABundleConfigMapName       = "prometheus-serving-cert-ca-bundle"
-	prometheusServiceSecretName            = "managed-ocs-prometheus-service-secret"
+	servingCertCABundleConfigMapName       = "prometheus-cert-bundle"
+	prometheusServiceSecretName            = "managed-ocs-prom-svc-secret"
```
nb-ohad, line 109: Why is the value of this constant (secret name) is widely different than the names convention on the surrounding constants

templates/prometheus.go
```diff
@@ -45,25 +50,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
+					MountPath: "/etc/kube-rbac-proxy",
+				},
+			},
+		}},
+		Volumes: []corev1.Volume{
+			{
+				Name: "serving-cert",
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{
+						SecretName: "managed-ocs-prom-svc-secret",
```
nb-ohad, line 91: Isn't this need to match the name of the secret we reconcile, if so use the same constant here and in the reconciler.

templates/prometheus.go
```diff
@@ -45,25 +50,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
+					},
+				},
+			},
+			{
+				Name: "ca-cert",
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: "prometheus-cert-bundle",
```
nb-ohad, line 100: Same as last comment

templates/prometheus.go
```diff
@@ -45,25 +50,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
+						},
+					},
+				},
+			},
+			{
+				Name: "kube-rbac-config",
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: "prometheus-tls-config",
```
nb-ohad, line 110: Same as last comment

templates/prometheuskuberbacproxyconfig.go
```diff
@@ -0,0 +1,42 @@
...
+					{
+						"verb":"get",
+						"resourceRequest":false,
+						"path":"/federate"
+					}
+				]
+			}
+		}`,
+	},
+}
```
nb-ohad, line 42: This is basically a string and we have no way to force a typed structure, every single type/structure error (like a missing comma, or special char) will be invisible until run time. I would suggest doing something similar to  

```suggestion
var KubeRBACProxyConfigMap = corev1.ConfigMap{
	Data: map[string]string{
		"config-file.json": (func() string {
			var config struct {
				Authorization struct {
					Static [2]struct {
						Path            string `json:"path"`
						ResourceRequest bool   `json:"resourceRequest"`
						Verb            string `json:"verb"`
					} `json:"static"`
				} `json:"authorization"`
			}

			item := &config.Authorization.Static[0]
			item.Verb = "get"
			item.Path = "/metrics"
			item.ResourceRequest = false

			item = &config.Authorization.Static[1]
			item.Verb = "get"
			item.Path = "/federate"
			item.ResourceRequest = false

			raw, _ := json.Marshal(config)
			return string(raw)
	})()
}
```

I do admit that defining and immediately calling an anonymous function is an inconvenient, but it does bring  benefits: 
- Everything is typed
- The anonymous struct is well defined and localized to the new func
- Everything Is defined only once
- init code is very clear
- the inlined function is called only once when the operator process starts
- The Marshal operation should never go fail, as we are marshaling a known typed struct

nb-ohad (changes requested):

templates/prometheusproxynetworkpolicy.go
```diff
@@ -0,0 +1,58 @@
...
+			netv1.PolicyTypeIngress,
+		},
+		PodSelector: metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{
+				{
+					Key:      "prometheus",
+					Operator: metav1.LabelSelectorOpIn,
+					Values: []string{
+						"managed-ocs-prometheus",
+					},
```
leelavg, line 53: - shouldn't this be as below as I don't see a label `prometheus: managed-ocs-promethues` being set on prom pod, correct me if I'm wrong.
``` go
					Key:      "app.kubernetes.io/name",
					Operator: metav1.LabelSelectorOpIn,
					Values: []string{
						"prometheus",
					},
```

leelavg (changes requested):

leelavg:
- pls resolve addressed conversations, thanks.

templates/prometheusproxynetworkpolicy.go
bindrad: I think prometheus operator sets this label `prometheus: managed-ocs-prometheus` on the prometheus-server pod(This is just a guess). But I checked Prometheus-server pod and it has the label `prometheus: managed-ocs-prometheus`  and `app: prometheus`. If you suggest using other one I'm fine with it, I just used `prometheus: managed-ocs-promethues` because it signifies that this is managed-ocs prometheus and not some other prometheus.

templates/prometheus.go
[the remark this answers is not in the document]
bindrad: I want to use the one created by us because this way we are not dependent on OCP versions as we already add annotations to support a wider range of OCP versions.

controllers/managedocs_controller_test.go
```diff
@@ -46,19 +46,18 @@ var _ = Describe("ManagedOCS controller", func() {
...
-	// 		Namespace: testPrimaryNamespace,
-	// 	},
-	// }
-	// caBundleConfigMap := corev1.ConfigMap{
-	// 	ObjectMeta: metav1.ObjectMeta{
-	// 		Name:      servingCertCABundleConfigMapName,
-	// 		Namespace: testPrimaryNamespace,
-	// 	},
-	// }
+	promServiceTemplate := corev1.Service{
```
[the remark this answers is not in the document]
bindrad: Will look into it when I update the tests

templates/prometheusproxynetworkpolicy.go
leelavg: ack, if the label exists on the prom pod then that's fine.

controllers/managedocs_controller.go
```diff
@@ -1731,27 +1823,26 @@ func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
 }
 
 func (r *ManagedOCSReconciler) deleteCSVByPrefix(name string) error {
+	csv, err := r.getCSVByPrefix(name)
```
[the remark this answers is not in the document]
nb-ohad: @leelavg Why? if we need to refactor in a PR we should refactor. 
If you see an issue with the refactor we need to fix it in this PR as well.

controllers/managedocs_controller.go
```diff
@@ -968,6 +1017,38 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
 		desired := templates.PrometheusTemplate.DeepCopy()
 		utils.AddLabel(r.prometheus, monLabelKey, monLabelValue)
+
+		deployerCSV, err := r.getCSVByPrefix(deployerCSVPrefix)
+		if err != nil {
+			return fmt.Errorf("Unable to set image for kube-rbac-proxy container: %v", err)
+		}
+
+		deployerCSVDeployments := deployerCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs
+		var deployerCSVDeployment *opv1a1.StrategyDeploymentSpec = nil
```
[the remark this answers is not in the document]
nb-ohad: Being very verbose should not be considered a fault. 
I vote to keeping it

controllers/managedocs_controller.go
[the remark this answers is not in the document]
bindrad: I'll use the label that we use for all monitoring resources. Updated in latest commit.

controllers/managedocs_controller.go
bindrad: Updated the name to `prometheus-svc-secret` in latest commit.

templates/prometheus.go
bindrad: Created a Public constant and used it in the reconciler as well. Updated in the latest commit.

templates/prometheus.go
bindrad: Same as last comment

templates/prometheus.go
bindrad: Same as last comment

templates/prometheuskuberbacproxyconfig.go
bindrad: Updated the code.

templates/prometheus.go
```diff
@@ -45,25 +50,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
+					Name:      "serving-cert",
+					MountPath: "/etc/tls-secret",
+				},
+				{
+					Name:      "ca-cert",
+					MountPath: "/etc/tls-cert",
+				},
+				{
+					Name:      "kube-rbac-config",
+					MountPath: "/etc/kube-rbac-proxy",
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit.

controllers/managedocs_controller.go
```diff
@@ -968,6 +1017,38 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
 
 		desired := templates.PrometheusTemplate.DeepCopy()
 		utils.AddLabel(r.prometheus, monLabelKey, monLabelValue)
+
+		deployerCSV, err := r.getCSVByPrefix(deployerCSVPrefix)
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
bindrad: I agree kept it as `var deployerCSVDeployment *opv1a1.StrategyDeploymentSpec = nil`

controllers/managedocs_controller.go
```diff
@@ -925,34 +959,49 @@ func (r *ManagedOCSReconciler) reconcilePrometheusService() error { //nolint:dea
...
+			prometheusServiceSecretName,
+		)
+		utils.AddAnnotation(
+			r.prometheusService,
+			"service.alpha.openshift.io/serving-cert-secret-name",
+			prometheusServiceSecretName,
+		)
+		// This label is required to enable us to use metrics federation
+		// mechanism provided by Managed-tenants
+		utils.AddLabel(r.prometheusService, "app", "prometheus")
```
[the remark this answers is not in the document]
bindrad: Updated in latest commit and using the label that we use for other monitoring resources.

leelavg (approved):

controllers/managedocs_controller.go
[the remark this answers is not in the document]
leelavg: - uh oh, this comment was folded and approved without seeing this
- ack, got the intention. How should we go about fixing this?
- I just mentioned `deleteCSVByPrefix` no need to use `getCSVByPrefix` if it needs to use I'd like to see below changes

1. look at call site of these two funcs and decide should `deleteCSVByPrefix` will be called only when CSV exists or it needs to be idempotent
2. Should `getCSVByPrefix` will be called only when CSV exists, IOW will this be invoked during uninstall code path as well?

controllers/managedocs_controller.go
```diff
@@ -1731,27 +1822,26 @@ func (r *ManagedOCSReconciler) own(resource metav1.Object) error {
 }
 
 func (r *ManagedOCSReconciler) deleteCSVByPrefix(name string) error {
+	csv, err := r.getCSVByPrefix(name)
+	if err != nil {
+		return err
+	}
+	return r.delete(csv)
```
nb-ohad, line 1829:
```suggestion
	if err == nil {
		return r.delete(csv)
	} else if errrors.IsNotFound(err) {
		return nil
	} else {
		return err
	}
```

templates/prometheus.go
```diff
@@ -31,6 +34,13 @@ var resourceSelector = metav1.LabelSelector{
 	},
 }
 
+var (
+	KubeRBACProxyPortNumber          int32  = 9339
+	PrometheusServiceSecretName      string = "prometheus-svc-secret"
+	ServingCertCABundleConfigMapName string = "prometheus-cert-bundle"
+	KubeRBACConfigMapName            string = "prometheus-tls-config"
```
nb-ohad, line 41:
```suggestion
	PrometheusKubeRBACPoxyConfigMapName            string = "prometheus-kube-rbac-proxy-config"
```

templates/prometheus.go
```diff
@@ -31,6 +34,13 @@ var resourceSelector = metav1.LabelSelector{
 	},
 }
 
+var (
+	KubeRBACProxyPortNumber          int32  = 9339
+	PrometheusServiceSecretName      string = "prometheus-svc-secret"
```
nb-ohad, line 39:
```suggestion
	PrometheusServingCertSecret      string = "prometheus-serving-cert-secret"
```

templates/prometheus.go
```diff
@@ -45,25 +55,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
+		Containers: []corev1.Container{{
+			Name: "kube-rbac-proxy",
+			Args: []string{
+				fmt.Sprintf("--secure-listen-address=0.0.0.0:%d", KubeRBACProxyPortNumber),
+				"--upstream=http://127.0.0.1:9090/",
+				"--logtostderr=true",
+				"--v=10",
+				"--tls-cert-file=/etc/tls-secret/tls.crt",
+				"--tls-private-key-file=/etc/tls-secret/tls.key",
+				"--client-ca-file=/etc/tls-cert/service-ca.crt",
```
nb-ohad, line 69:
```suggestion
				"--client-ca-file=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.cert",
```

templates/prometheus.go
```diff
@@ -45,25 +55,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
+			}},
+			VolumeMounts: []corev1.VolumeMount{
+				{
+					Name:      "serving-cert",
+					MountPath: "/etc/tls-secret",
+				},
+				{
+					Name:      "ca-cert",
+					MountPath: "/etc/tls-cert",
+				},
```
nb-ohad, line 84: Please remove (see comments above)

templates/prometheus.go
```diff
@@ -45,25 +55,68 @@ var PrometheusTemplate = promv1.Prometheus{
...
+			{
+				Name: "ca-cert",
+				VolumeSource: corev1.VolumeSource{
+					ConfigMap: &corev1.ConfigMapVolumeSource{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: ServingCertCABundleConfigMapName,
+						},
+					},
+				},
+			},
```
nb-ohad, line 109: Please remove (see comments above)

templates/prometheus.go
```diff
@@ -31,6 +34,13 @@ var resourceSelector = metav1.LabelSelector{
 	},
 }
 
+var (
+	KubeRBACProxyPortNumber          int32  = 9339
```
nb-ohad, line 38: Please use `int` instead of `int32`, and convert to the specific storage class when needed
```suggestion
	KubeRBACProxyPortNumber          int  = 9339
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -153,6 +149,7 @@ type ManagedOCSReconciler struct {
 	addonParams                        map[string]string
 	onboardingValidationKeySecret      *corev1.Secret
 	servingCertCaBundleConfigMap       *corev1.ConfigMap
+	kubeRBACConfigMap                  *corev1.ConfigMap
```
nb-ohad, line 152: This is not a general kube rbac config map, it is specifically for prometheus
```suggestion
	prometheusKubeRBACConfigMap                  *corev1.ConfigMap
```

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: @leelavg I think your concern is addressed with the newest code changes

controllers/managedocs_controller_test.go
```diff
@@ -531,7 +523,7 @@ var _ = Describe("ManagedOCS controller", func() {
 				deployerCSV.Name = testDeployerCSVName
 				deployerCSV.Namespace = testPrimaryNamespace
 				deployerCSV.Spec.InstallStrategy.StrategyName = "test-strategy"
-				deployerCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs = []opv1a1.StrategyDeploymentSpec{}
+				deployerCSV.Spec.InstallStrategy.StrategySpec.DeploymentSpecs = getMockOCSOSDDeployerCSVDeploymentSpec()
```
nb-ohad, line 526: Why do we need to refer to a struct that lives in another file? Why can't we just embed the necessary struct here?

controllers/managedocs_controller_test.go
```diff
@@ -973,6 +964,73 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					// Get an updated configMap
+					kubeRBACConfigMap := prometheusKubeRBACPoxyConfigMapTemplate.DeepCopy()
+					kubeRBACConfigMapKey := utils.GetResourceKey(kubeRBACConfigMap)
+					Expect(k8sClient.Get(ctx, kubeRBACConfigMapKey, kubeRBACConfigMap)).Should(Succeed())
+
+					// Update to the configMap with dummy values
+					data := kubeRBACConfigMap.Data
+					kubeRBACConfigMap.Data = map[string]string{
+						"test-key": "test-value",
+					}
```
nb-ohad, line 978: No need to add mock data, we just want to verify that it is reconciling back. An empty object will work just fine
```suggestion
					kubeRBACConfigMap.Data = map[string]string{}
```

controllers/managedocs_controller_test.go
```diff
@@ -973,6 +964,73 @@ var _ = Describe("ManagedOCS controller", func() {
...
+					// Update to empty spec
+					spec := prometheusService.Spec
+					prometheusService.Spec = corev1.ServiceSpec{
+						Ports: []corev1.ServicePort{
+							{
+								Port: 1234,
+							},
+						},
+						ClusterIP: spec.ClusterIP,
+					}
```
nb-ohad, line 1014: No need to add mock data, we just want to verify that it is reconciling back. An empty object will work just fine
```suggestion
					prometheusService.Spec = corev1.ServiceSpec{}
```

controllers/managedocs_controller_test.go
```diff
@@ -1742,6 +1800,35 @@ var _ = Describe("ManagedOCS controller", func() {
...
+
+					// Wait for the spec changes to be reverted
+					Eventually(func() *netv1.NetworkPolicySpec {
+						prometheusProxyIngress := prometheusProxyIngressNetworkPolicyTemplate.DeepCopy()
+						Expect(k8sClient.Get(ctx, prometheusProxyIngressKey, prometheusProxyIngress)).Should(Succeed())
+						return &prometheusProxyIngress.Spec
+					}, timeout, interval).Should(Equal(spec))
+				})
+			})
+			When("the prometheus proxy ingress  NetworkPolicy resource is deleted", func() {
```
nb-ohad, line 1823:
```suggestion
			When("the prometheus proxy ingress NetworkPolicy resource is deleted", func() {
```

controllers/suite_test.go
```diff
@@ -279,6 +279,33 @@ func getMockOCSCSVDeploymentSpec() []opv1a1.StrategyDeploymentSpec {
...
+								Image: "test",
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	return deploymentSpec
+}
```
nb-ohad, line 307: This is used in a single test/location and should be embedded there

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
```diff
@@ -973,6 +964,73 @@ var _ = Describe("ManagedOCS controller", func() {
...
+						return kubeRBACConfigMap.Data
+					}, timeout, interval).Should(Equal(data))
+				})
+			})
+			When("the kube-rbac-proxy ConfigMap resource is deleted", func() {
+				It("should create a new kube-rbac-proxy ConfigMap in the namespace", func() {
+					// Delete the NetworkPolicy resource
+					Expect(k8sClient.Delete(ctx, prometheusKubeRBACPoxyConfigMapTemplate.DeepCopy())).Should(Succeed())
+
+					// Wait for the NetworkPolicy to be recreated
```
leelavg, line 994: - in comment `s/NetworkPolicy/prometheusKubeRBACPoxyConfigMap/g`

controllers/managedocs_controller.go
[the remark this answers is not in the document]
leelavg: ack, @bindrad change lgtm, can be resolved.

controllers/managedocs_controller_test.go
bindrad: It can't be done for service because ports is a required field and clusterIP is an immutable field. So we need to add a dummy port and keep the clusterID as whatever it was before.

nb-ohad: @bindrad Ok, please add a comment explaining exactly that

PR #176, opened 2022-04-22
Added new metrics to be federated according to new alertRules

bindrad:
This fixes two alerts
1. CephMonQuorumLost
2. CephMgrIsMissingReplicas

As there are two replicas of the Prometheus pod in the openshift-monitoring namespace and when we execute PersistentVolumeUsageNearFull and PersistentVolumeUsageCritical alertRule we get this error `Error executing query: found duplicate series for the match group {storageclass="ocs-storagecluster-cephfs"} on the right hand-side of the operation: [{__name__="kube_storageclass_info", container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", prometheus="openshift-monitoring/k8s", provisioner="openshift-storage.cephfs.csi.ceph.com", reclaim_policy="Delete", service="kube-state-metrics", storageclass="ocs-storagecluster-cephfs", volume_binding_mode="Immediate"}, {__name__="kube_storageclass_info", container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", pod="prometheus-k8s-1", prometheus="openshift-monitoring/k8s", provisioner="openshift-storage.cephfs.csi.ceph.com", reclaim_policy="Delete", service="kube-state-metrics", storageclass="ocs-storagecluster-cephfs", volume_binding_mode="Immediate"}];many-to-many matching not allowed: matching labels must be unique on one side` so we added labeldrop for pod label.

But now CephMonQuorumLost alert requires pod label so we are relabelling in such a way that the metrics that have pod label whose value is prometheus-k8s-0 or prometheus-k8s-1 will be replaced with blank value.
```
Metric before change:
    kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", pod="prometheus-k8s-1", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}
Metrics after change:
    kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}
```

### DIFF
--- a/templates/k8smetricsservicemonitor.go
+++ b/templates/k8smetricsservicemonitor.go
@@ -24,6 +24,8 @@ var params = map[string][]string{
 		"{__name__='kube_persistentvolume_info'}",
 		"{__name__='kube_storageclass_info'}",
 		"{__name__='kube_persistentvolumeclaim_info'}",
+		"{__name__='kube_deployment_spec_replicas'}",
+		"{__name__='kube_pod_status_phase'}",
 		"{__name__='kubelet_volume_stats_capacity_bytes'}",
 		"{__name__='kubelet_volume_stats_used_bytes'}",
 		"{__name__='node_disk_read_time_seconds_total'}",
@@ -45,14 +47,19 @@ var K8sMetricsServiceMonitorTemplate = promv1.ServiceMonitor{
 				ScrapeTimeout: "1m",
 				Interval:      "2m",
 				HonorLabels:   true,
-				MetricRelabelConfigs: []*promv1.RelabelConfig{
+				RelabelConfigs: []*promv1.RelabelConfig{
 					{
 						Action: "labeldrop",
 						Regex:  "prometheus_replica",
 					},
 					{
-						Action: "labeldrop",
-						Regex:  "pod",
+						Action:      "replace",
+						Regex:       "prometheus-k8s-.*",
+						Replacement: "",
+						SourceLabels: []string{
+							"pod",
+						},
+						TargetLabel: "pod",
 					},
 				},
 				TLSConfig: &promv1.TLSConfig{


bindrad:
Based on our call regarding this PR, I don't think it is possible to drop the label with specific key-value pair
The other alternative is to replace prometheus-k8s-0 and prometheus-k8s-1 values with something generic such as promethueus-k8s and drop metrics that have prometheus-k8s-0 and prometheus-k8-1 values for the pod label

leelavg:
@bindrad does this pr needs a re-work?

bindrad:
> @bindrad does this pr needs a re-work?

@leelavg I re-worked on it and pushed the changes 2 days ago, waiting for it to get reviewed.

leelavg:
- Ack, will take a look at the earliest.
- I read last comment and pinged for an update, thanks for the confirmation.

templates/k8smetricsservicemonitor.go
```diff
@@ -51,8 +53,20 @@ var K8sMetricsServiceMonitorTemplate = promv1.ServiceMonitor{
...
+							"pod",
+						},
+						TargetLabel: "pod",
+					},
+					{
+						SourceLabels: []string{
+							"pod",
+						},
+						Action: "drop",
+						Regex:  "prometheus-k8s-.*",
```
leelavg, line 69: 1. CephMonQuorumLost alert requires pod label 
2. drop metrics that have prometheus-k8s-0 and prometheus-k8-1

- I see above two as contradicting statements and this PR seems to implement second statement?

---
- If the first config is replacing the label (`pod`) value as emtpy string and how will the second rules get's matched?

ps: prom relabel config is a new reading to me so I might've misunderstood somethings here.

bindrad: Let me explain using an example
Without any relabelling, the metrics will look something like this
- `kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", pod="prometheus-k8s-1", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}`
- `kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", pod="prometheus-k8s-0", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}`

After first relabelling is applied
```
Action:      "replace",
Regex:       "prometheus-k8s-.*",
Replacement: "",
SourceLabels: []string{
	"pod",
},
TargetLabel: "pod",

```
the metrics will look something like this
- `kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", pod="prometheus-k8s-1", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}`
- `kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", pod="prometheus-k8s-0", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}`
- `kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}`

You can see one more metric is added that does not have pod label but we still we need to remove the metric that has pod label with value prometheus-k8-* so we need second relabel rule where we will drop the metric that has pod label with value prometheus-k8-*

After second relabelling 
```
SourceLabels: []string{
	"pod",
},
Action: "drop",
Regex:  "prometheus-k8s-.*",

```
the metrics will look something like this
- `kube_pod_status_phase{container="kube-rbac-proxy-main", endpoint="https-main", job="kube-state-metrics", namespace="openshift-monitoring", phase="Succeeded", prometheus="openshift-monitoring/k8s", service="kube-state-metrics", uid="071feb6a-7d97-4ec3-93fd-938d27c5a097"}`

This observation is based on testing relabelling on a live cluster

bindrad: > CephMonQuorumLost alert requires pod label
> drop metrics that have prometheus-k8s-0 and prometheus-k8-1
> I see above two as contradicting statements and this PR seems to implement second statement?
CephMonQuorumLost alert uses kube_pod_status_phase metric{pod="rook-ceph-mon-.*"} metric
In the current codebase, we are just dropping pod label from all metrics that being federated so alertRule for CephMonQuorumLost doesn't work because it cannot get this metric kube_pod_status_phase metric{pod="rook-ceph-mon-.*"}

leelavg: 1. ack, I thought the initial config modifies the metrics in-place, that doesn't seem to be case and it creates a new metric omitting the key pair. Understood, thanks a lot for the explanation.
> we are just dropping pod label from all metrics
- Doesn't your PR effectively dropping the label `pod` from a metric family which has `pod: prometheus...` and when `CephMonQuorumLost` query is run doesn't it fail while querying the metric with not existing label (or) it's like the query will be successful even for a single match unlike `labeldrop` which the existing code is doing?

bindrad: > Doesn't your PR effectively dropping the label pod from a metric family which has pod: prometheus...

Yes, PR is effectively dropping the label from the metrics that have pod label values as `prometheus-k8s.*`

> when CephMonQuorumLost query is run doesn't it fail while querying the metric with not existing label

The query doesn't fail, the output of the query is nothing/nil
Something similar to SQL if you have a where condition and the SQL query doesn't return anything that does not mean that query has failed, it means that the query returns nothing.

leelavg (dismissed):
/lgtm

templates/k8smetricsservicemonitor.go
nb-ohad: @bindrad Why are we relabeling before ingest and not during scraping? I believe that if we will do it during scraping you will have no need to delete duplicate metrics as they will not exist in the first place. 

see
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
vs 
https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs

bindrad: I tested using relabel_config, it works without dropping the metrics(and just replacing)

nb-ohad (approved):

bindrad:
@nb-ohad This PR can be merged

PR #180, opened 2022-05-05
Mechanism to fetch AWS data to be used by the deployer.

renan-campos:
In provider/consumer configurations, information about the underlying network infrastructure is needed to correctly regulate traffic to the cluster. This PR introduces a mechanism to get this data from AWS without the need for AWS credentials. The deployer uses this data to configure the egressNetworkPolicy.


Makefile
```diff
@@ -123,7 +126,7 @@ generate: controller-gen
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
 # Build the docker image
-docker-build: test
```
renan-campos, line 126: I don't think we need to test the code every time we want to make a new local mage. Most of the times this command is being run because a local image is going to be tested manually.

Dockerfile
```diff
@@ -16,19 +16,27 @@ COPY controllers/ controllers/
 COPY utils/ utils/
 COPY templates/ templates/
 COPY readinessProbe/ readinessProbe/
+COPY cmd/awsDataGather/ awsDataGather/
+COPY pkg/ pkg/
```
renan-campos, line 20: I put the main program for aws data gathering in `cmd/` and its auxiliary objects (which is also used by the deployer) in `pkg/` to try out a [common layout pattern for go projects](https://github.com/golang-standards/project-layout). I'm interested to know others thoughts on this. If it looks promising, I can move the readinessProbe/, utils/ and templates/ as well.

Dockerfile
```diff
@@ -16,19 +16,27 @@ COPY controllers/ controllers/
...
+
+RUN mkdir bin
```
renan-campos, line 22: I put the main program for aws data gathering in `cmd/` and its auxiliary objects (which is also used by the deployer) in `pkg/` to try out a [common layout pattern for go projects](https://github.com/golang-standards/project-layout). I'm interested to know others thoughts on this. If it looks promising, I can move the readinessProbe/, utils/ and templates/ as well.

controllers/managedocs_controller.go
```diff
@@ -1467,6 +1497,9 @@ func (r *ManagedOCSReconciler) reconcileEgressNetworkPolicy() error {
 		if err != nil {
 			return fmt.Errorf("Unable to parse DMS url: %v", err)
 		}
+		if snitchURL.Hostname() == "" {
+			return fmt.Errorf("DMS snitch url %q does not have a valid hostname", snitchURL)
+		}
```
renan-campos, line 1502: This was a separate issue I found while making these changes.

controllers/managedocs_controller.go
```diff
@@ -507,6 +510,19 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
+		if err := r.reconcileIngressNetworkPolicy(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileCephIngressNetworkPolicy(); err != nil {
+			return ctrl.Result{}, err
+		}
+		if err := r.reconcileProviderAPIServerNetworkPolicy(); err != nil {
+			return ctrl.Result{}, err
+		}
+
```
renan-campos, line 525: By reconciling security related objects earlier in the stack, we can ensure that the service isn't missing any security configuration when the storage is available.

pkg/aws/gather.go
```diff
@@ -0,0 +1,92 @@
...
+		if k8serrors.IsAlreadyExists(err) {
+			err = k8sClient.Update(context.Background(), configMap)
+		}
+	}
+	return err
+}
+
+func httpGet(address string, log logr.Logger) string {
+	// To be robust to network hiccups, this get function will retry on failures
+	for tries := 0; tries < 3; tries++ {
```
leelavg, line 73: - these will be continuous http requests without any delay, no issues with that?

leelavg (commented):
- I only skimmed through the PR and I'll review after removal of WIP

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,51 @@
...
+
+	k8sClient, err := client.New(config.GetConfigOrDie(), options)
+	if err != nil {
+		log.Error(err, "error creating client")
+		os.Exit(1)
+	}
+
+	aws.GatherData(aws.IMDSv1Server, k8sClient, namespace, log)
+
+	fmt.Println("AWS data gathering successfully completed!")
```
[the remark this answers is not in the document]
renan-campos: Why's that?

Dockerfile
```diff
@@ -16,19 +16,27 @@ COPY controllers/ controllers/
...
+# Build aws data gathering binary
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o bin/awsDataGather awsDataGather/main.go
+
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
 FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/manager .
 COPY --from=builder /workspace/readinessServer .
+COPY --from=builder /workspace/bin/ /bin/
```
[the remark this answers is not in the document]
leelavg: - That's also fine, moving all to `/bin`
- I just meant above to be consistent with existing usage
- In general executables can be in many places, `/bin`, `/usr/local/bin` etc, however in containers usually we don't get `which` to verify binary path existence 
- And if we are not sure where the binary exist then need to refer Dockerfile before mentioning `command: <path/to/binary>` in deployments, so either not to fiddle `PATH` env or remember path/to/binary we just copy executables to `/` particularly single binary compiled code afaik

cmd/awsDataGather/main.go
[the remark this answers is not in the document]
leelavg: - just thought you wanted to log rather than print
- all good!

pkg/aws/gather.go
renan-campos: @leelavg I added sleep to this function, please let me know what you think and resolve the conversation if it looks good to you.

pkg/aws/gather.go
```diff
@@ -0,0 +1,92 @@
...
+		defer resp.Body.Close()
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			log.Error(err, "Failed to read body of GET")
+			continue
+		}
+		return string(body), nil
+	}
+	return "", fmt.Errorf("http get call failed")
+}
```
nb-ohad, line 92: This function has a couple of drawbacks. 
1. The call site does not know anything about the fact that more than one HTTP get call is going to be done.
2. The `defer` will defer closing of all attempts to the end function, which in this case is 3 different tries
3. The function implements both the concern of retring and HTTP get. 

I would suggest separating the retry functionality into a util and moving it into the utils file:
```
func retry(attempts int, sleep time.Duration, f func() error) {
	for i := 0; ; i++ {
		if err = f(); err == nil {
			return 
		}
		if i >= (attempts - 1) {
        		break
		}
		time.Sleep(sleep)
		log.Println("retrying after error:", err)
	}
}
```

and for the call sites, you can do:
```
var mac string;
if err := retry(3, 1, func() err {
	var endpoint := fmt.Sprintf("%s/latest/meta-data/mac", imdsServerAddr)
	res, err := http.Get(endpoint)
	if err != nil {
		return fmt.Errorf("Failed to GET %s, endpoint, %v", err)
	}
	defer res.Body.Close()
	
	mac, err := io.ReadAll(resp.Body)
	if err != nil {
		return fmt.Errorf("Failed to read response body, %v", err)
	}
})
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
```diff
@@ -1564,17 +1568,43 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
+				Type: openshiftv1.EgressNetworkPolicyRuleAllow,
+				To: openshiftv1.EgressNetworkPolicyPeer{
+					CIDRSelector: vpcCIDR,
+				},
+			}
+
+			desired.Spec.Egress = append(
+				[]openshiftv1.EgressNetworkPolicyRule{
+					vpcEgressRule,
+				}, desired.Spec.Egress...)
```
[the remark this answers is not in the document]
renan-campos: What is the advantage of doing it the way you suggest? What is the issue appending the rule?

pkg/aws/gather.go
```diff
@@ -0,0 +1,92 @@
+package aws
```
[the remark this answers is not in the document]
nb-ohad: @renan-campos I didn't mean to comment on the `package aws` i ment to comment on this entire `gather.go` file content. 

Sorry for the confusion, but please retry to read my original comment with this new context

pkg/aws/gather.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	IMDSv1Server      = "http://169.254.169.254"
```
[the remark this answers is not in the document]
renan-campos: Yes. I can add a comment to point to where it says this is the case.

pkg/aws/gather.go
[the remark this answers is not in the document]
renan-campos: `DataConfigMapName` in `gather.go` is used in both the aws data gather program as well as in the controller code.
I would prefer to separate the function implementation and definitions from the main program.

pkg/aws/gather.go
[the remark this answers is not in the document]
nb-ohad: if this is an AWS public documentation, please do so

controllers/managedocs_controller.go
renan-campos: Actually since this is breaking tests, I'll revert it.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: The same reasons we do everything in templates. 
The entire role can be described as template that just needs to be filled and the code part becomes
 more concise. 
 So this:
 ```
vpcEgressRule := openshiftv1.EgressNetworkPolicyRule{
	Type: openshiftv1.EgressNetworkPolicyRuleAllow,
	To: openshiftv1.EgressNetworkPolicyPeer{
		CIDRSelector: vpcCIDR,
	},
}

desired.Spec.Egress = append(
	[]openshiftv1.EgressNetworkPolicyRule{
		vpcEgressRule,
	}, desired.Spec.Egress...)
 ```
becomes:
```
desired.Spec.Egress[0].To.CIDRSelector: vpcCIDER
```

pkg/aws/gather.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		defer resp.Body.Close()
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("Failed to read response body: %v", err)
+		}
+		mac = string(body)
+		return nil
+	})
```
nb-ohad, line 44: this code is repeating between the 2 get operations, this will be a good candidate for a refactor. You can introduce another func 
```
func getAndParseBody(endpoint) (string, error) {
	resp, err := http.Get(endpoint)
	if err != nil {
		return "", fmt.Errorf("Failed to GET %s: %v", endpoint, err)
	}
	defer resp.Body.Close()

	body, err := io.ReadAll(resp.Body)
	if err != nil {
		return "", fmt.Errorf("Failed to read response body: %v", err)
	}
	return string(body), nil
}
```
then 
```suggestion
	err := utils.Retry(3, time.Second, func() error {
		endpoint := fmt.Sprintf("%s/latest/meta-data/mac", imdsServerAddr)
		if mac, err = getAndParseBody(); err != nil {
			return err
		} 
	})
```

pkg/aws/gather.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		defer resp.Body.Close()
+
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("Failed to read response body: %v", err)
+		}
+		cidr = string(body)
+		return nil
+	})
```
nb-ohad, line 67: Continuation of the suggestion from the last comment
```suggestion
	err := utils.Retry(3, time.Second, func() error {
		endpoint := fmt.Sprintf(
			"%s/latest/meta-data/network/interfaces/macs/%s/vpc-ipv4-cidr-block",
			imdsServerAddr, 
			mac
		)
		if cidr, err = getAndParseBody(endoint); err != nil {
			return err
		} 
	})
```

pkg/aws/gather.go
```diff
@@ -0,0 +1,89 @@
...
+	log.Info("VPC CIDR found.", "vpc-cidr", cidr)
+
+	log.Info("Creating Config Map with AWS data")
+	configMap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      DataConfigMapName,
+			Namespace: namespace,
+		},
+		Data: map[string]string{},
+	}
```
nb-ohad, line 80: No need to creat the map here as the entire map needs to be overwriten inside the CreateOrUpdate callback to reconcile out any other changes to the configmap
```suggestion
	configMap := corev1.ConfigMap{}
	configMap.Name = DataConfigMapName
	configMap.Namespace = namespace
```

pkg/aws/gather.go
```diff
@@ -0,0 +1,89 @@
...
+	_, err = ctrl.CreateOrUpdate(context.Background(), k8sClient, &configMap, func() error {
+		configMap.Data[CidrKey] = cidr
```
nb-ohad, line 82: Reconcile out any other changes to the configmap data
```suggestion
		configMap.Data map[string]string{
			CidrKey: cidr
		} 
```

pkg/aws/gather.go
```diff
@@ -0,0 +1,89 @@
...
+
+	log.Info("Creating Config Map with AWS data")
+	configMap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      DataConfigMapName,
+			Namespace: namespace,
+		},
+		Data: map[string]string{},
+	}
+	_, err = ctrl.CreateOrUpdate(context.Background(), k8sClient, &configMap, func() error {
```
nb-ohad, line 81: The first thing you should do in the callback is to set an owner for this configmap. 
As this deployment is a proxy of deployer I think it makes sense to have the ManagedOCS cr be the owner.
This will guarantee proper ownership and garbage collection. 


Another option is to set the deployment resource as the owner.

nb-ohad (changes requested):

pkg/aws/gather.go
renan-campos: Thanks a lot for this @nb-ohad , your suggested refactor significantly improves code readability. While looking at it, I had an idea for a code-review strategy that I wanted to run by you to understand the disadvantages of the approach. The following comment will be what I would add if I were the reviewer, let me know what you think:

renan-campos:
```suggestion
// TODO: Needs refactor
	err := utils.Retry(3, time.Second, func() error {
		endpoint := fmt.Sprintf("%s/latest/meta-data/mac", imdsServerAddr)
		resp, err := http.Get(endpoint)
		if err != nil {
			return fmt.Errorf("Failed to GET %s: %v", endpoint, err)
		}
		defer resp.Body.Close()

		body, err := io.ReadAll(resp.Body)
		if err != nil {
			return fmt.Errorf("Failed to read response body: %v", err)
		}
		mac = string(body)
		return nil
	})
```

renan-campos: Here's my reasoning for this:

- It saves the reviewer from having to refactor the code for the contributor.
- It gives the contributor the opportunity to refactor the code themselves.
- It doesn't block merging the current PR.
- If the original contributor doesn't want to do the refactor, it gives other contributors a chance to do so.
- If no other contributor knows how to do the refactor, they can reach out to the reviewer that placed the suggestion there to learn how to do it.

I see it like a "teach a man to fish vs. giving the man a fish" scenario.

nb-ohad: @renan-campos 
As a reviewer, you identify places that might benefit from refactoring only when you have some notion of the refactoring that you would like to see. Putting a comment saying `Needs refactor` does not help the contributor to understand what kind of refactoring you have in mind, or even evaluate or rebuttal the refactoring request. 
In addition, my assumption is that the contributor didn't find the desired way to refactor the code because if he did, he would have submitted the refactored code in the first place. Just pointing to the fact that some piece of code need refactoring will not help the contributor to come up with a refactoring strategy. 

This all boils down to the conclusion that as a reviewer if you think a specific refactoring is needed - just propose it. And this is exactly the kind of review comments I tend to leave in these scenarios.

nb-ohad: Now as a maintainer I will say that you have the responsibility to gate code that does not meet a specific standard, so:
> It doesn't block merging the current PR.
> f the original contributor doesn't want to do the refactor, it gives other contributors a chance to do so.
These are not valid points, as this code will not be approved for merging without a proper rebuttal

renan-campos: Isn't the point 
> * It doesn't block merging the current PR.
Valid since this suggestion blocks function for form? I might be wrong, but I think the code that is there and the code you suggested does the exact same thing, only yours does so more clearly?

nb-ohad: No, you are correct both forms are accomplishing the exact same thing and even the original code (before all refactors) did the exact same thing. There are infinite ways to achieve the exact same behavior.
But a review is not only about the correctness of a contribution, it is also about other properties of the contributions. Contributed code is measured by multiple factors:

* Correctness
* Performance - when relevant and across multiple dimensions 
* future maintenance cost
* Standards - which relates to the prev bullet
* Readability - which then again is measured across multiple dimensions 
* Reusability - when relevant
and so on. 

And of course some of these are subjective measurments - but this is why a maintainer on one project might not fit the position of a maintainer on a different project

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: Can you please elaborate?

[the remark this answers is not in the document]
renan-campos: Sure: If you add the EgressNetworkPolicyRule to the template, you need to leave the value for CIDRSelector as "". That's fine for the non-converged cases since it will get mutated to the vpcCIDR once its determined, but for the converged case it will remain blank. How the EgressNetworkPolicyRule is handled when the CIDRSelector is blank is not defined behavior (At least from what I can tell in the [documentation](https://docs.openshift.com/container-platform/4.10/networking/openshift_sdn/configuring-egress-firewall.html)). 

It would be more accurate to say that it _may_ break converged, but I think its safe to go with the assumption that the undefined behavior will break it as opposed to testing that it doesn't, and then having to make sure that every update to openshift sdn doesn't change that behavior.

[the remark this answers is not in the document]
renan-campos: I can use a template if you want, I just think it would be better to use a separate template for non-converged cases then.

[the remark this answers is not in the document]
nb-ohad: I see. But we do have a way to separate code paths in our code base. so why can't we just mutate the template in accordance with the deployment type? 
```
r.DeploymentType == convergedDeploymentType {
    desired.Spec.Egress[0] = nil
} else {
    desired.Spec.Egress[0].To.CIDRSelector: vpcCIDER
}
```
But I guess my suggestion is suffering from the problem of relying on a constant placement in the array (position 0), and there is no way to query a rule (in the template) by some sort of identity. So for now I think your suggestion is better with a small request for change. Instead of 


I see. But we do have a way to separate code paths in our code base. so why can't we just mutate the template in accordance with the deployment type? 
```
r.DeploymentType == convergedDeploymentType {
    desired.Spec.Egress[0] = nil
} else {
    desired.Spec.Egress[0].To.CIDRSelector: vpcCIDER
}
```
But I guess my suggestion is suffering from the problem of relying on a constant placement in the array (position 0), and there is no way to query a rule (in the template) by some sort of identity. So, for now, I think your suggestion is better with a small request for change. Instead of adding the new rule at the start of the array, let's append at the end:
```
desired.Spec.Egress = append(
	desired.Spec.Egress,
	openshiftv1.EgressNetworkPolicyRule{
		Type: openshiftv1.EgressNetworkPolicyRuleAllow,
		To: openshiftv1.EgressNetworkPolicyPeer{
			CIDRSelector: vpcCIDR,
		},
	},
)
```
As it results in a more concise and readable code.
Unless there is a concrete reason for it to be the first rule.

pkg/aws/gather.go
renan-campos: I see the value of the suggestion : improved readability reduces future maintenance cost.

I don't want to keep you from your weekend, I had one last thought on this if you want to talk more monday:
The factors that need to be taken into account when reviewing contributed code affect how the number of contributors and reviewers can scale. If the standards are too stringent or subjective:
- fewer contributors will be able to confidently and independently make changes to the code
- fewer other reviewers will be able to match the quality of reviews of the OG reviewer

In an extreme case, the long term result could be reviewers and contributors that can only code and review in exactly the same way as the original.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
renan-campos: There is a reason why I inserted the rule in the front instead of appending it: [The order of the rules matter](https://docs.openshift.com/container-platform/4.10/networking/openshift_sdn/configuring-egress-firewall.html#policy-rule-order_openshift-sdn-egress-firewall). This rule must come before the EgressNetworkPolicyRuleDeny rule. Inserting the rule in the front ensures that the vpc policy comes before it.

I can add a comment explaining this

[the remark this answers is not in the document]
nb-ohad: Please do so, It will clarify this bit for a future reader.

pkg/aws/gather.go
nb-ohad: I am not sure that I agree here. If I was to implement this feature I would have done it differently. All of the review suggestions are amounting to a couple of requested changes to some pieces of code that I find a little bit lacking. The Idea was never to push you to implement my vision for a solution. 

Saying that I do agree this entire discussion boils down to an exercise of balance. 
As you have suggested, let's continue this meta-discussion outside of the PR comments.

pkg/aws/gather.go
renan-campos: To avoid the dependence on the ManagedOCS resource, I'll set the deployment resource as the owner.

utils/utils.go
```diff
@@ -66,3 +71,33 @@ func AddAnnotation(obj metav1.Object, key string, value string) {
 func GetRegexMatcher(alerts []string) string {
 	return "^" + strings.Join(alerts, "$|^") + "$"
 }
+
+func Retry(attempts int, sleep time.Duration, f func() error) error {
+	var err error
+	for i := 0; ; i++ {
```
leelavg, line 77:
```suggestion
	for i := 0; i < attempts ; i++ {
```

utils/utils.go
```diff
@@ -66,3 +71,33 @@ func AddAnnotation(obj metav1.Object, key string, value string) {
...
+		if err = f(); err == nil {
+			return nil
+		}
+		if i >= (attempts - 1) {
+			break
+		}
```
leelavg, line 83:
```suggestion
```
- can be moved into `for` condition

pkg/aws/gather.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+func GatherData(imdsServerAddr string, log logr.Logger) (map[string]string, error) {
+	var data map[string]string = map[string]string{}
+
+	// A mac address for one of the interfaces on the instance is needed
+	// to query for the VPC CIDR block.
+	log.Info("Determining mac address on instance")
+	var mac string
+	err := utils.Retry(3, time.Second, func() error {
```
leelavg, line 25: - even though it's verbose, having `1 * time.Second` shows intent more accurately 🤔

pkg/aws/gather.go
```diff
@@ -0,0 +1,57 @@
...
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to find mac address of instance")
+	}
+	log.Info("MAC address found.", "mac", mac)
+
+	log.Info("Determining VPC CIDR")
+	var cidr string
+	err = utils.Retry(3, time.Second, func() error {
```
leelavg, line 40: - even though it's verbose, having 1 * time.Second shows intent more accurately

pkg/aws/gather_test.go
```diff
@@ -0,0 +1,63 @@
...
+	configMap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      DataConfigMapName,
+			Namespace: testNamespace,
+		},
+	}
+
+	When("aws data gathering successfully completes", func() {
+		BeforeEach(func() {
+			err := GatherData("http://"+testMockerAddr, k8sClient, testNamespace, log)
```
leelavg, line 29: - seems the function signature isn't changed in test

pkg/aws/gather_test.go
```diff
@@ -0,0 +1,63 @@
...
+	log := ctrl.Log.WithName("aws-data-gather-test")
+	ctx := context.Background()
+
+	configMap := corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      DataConfigMapName,
+			Namespace: testNamespace,
+		},
+	}
+
```
leelavg, line 26: - if possible, before this test can we have below two tests?
1. Mocker is reachable within the time period
2. Mocker is unreachable within time period and assert the flow stops

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,94 @@
...
+		log.Error(err, "Failed to create configmap")
+		os.Exit(1)
+	}
+
+	log.Info("AWS data gathering successfully completed!")
+
+	// Set up signal handler, so that a clean up procedure will be run if the pod running this code is deleted.
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM)
+	<-signalChan
```
leelavg, line 91: - need to re-look, just a thought, why can't we set `restartPolicy: OnFailure` on the pod spec and remove this signalling thereby reducing a background process which is only awaiting to be killed?

config/rbac/aws_data_gather.yaml
```diff
@@ -0,0 +1,56 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: aws-data-gather
+rules:
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
```
leelavg, line 15: - I think the resources here are inadequate as I see `pod: get, cm: get` in code
- Easy one would be to add `- pod` here or separate `pod` roles

controllers/managedocs_controller.go
```diff
@@ -1457,17 +1460,44 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
 			return err
 		}
 		desired := templates.EgressNetworkPolicyTemplate.DeepCopy()
 
+		if r.DeploymentType != convergedDeploymentType {
+			// Get the aws config map
+			awsConfigMap := corev1.ConfigMap{}
+			awsConfigMap.Name = aws.DataConfigMapName
+			awsConfigMap.Namespace = r.namespace
+			if err := r.get(&awsConfigMap); err != nil {
```
leelavg, line 1474:
```suggestion
			awsConfigMap := &corev1.ConfigMap{}
			awsConfigMap.Name = aws.DataConfigMapName
			awsConfigMap.Namespace = r.namespace
			if err := r.get(awsConfigMap); err != nil {
```

pkg/aws/gather_test.go
renan-campos: @leelavg I can't really think of a simple way to do this, would it be alright if we deferred this to perhaps a later pr?

cmd/awsDataGather/main.go
renan-campos: Actually there was another reason I did this: Creating a deployment through a bundle does not support a restartPolicy of OnFailure. I'll add a comment with this in it.

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,110 @@
...
+
+	k8sClient, err := client.New(config.GetConfigOrDie(), options)
+	if err != nil {
+		log.Error(err, "error creating client")
+		os.Exit(1)
+	}
+
+	// This will later be used as the OwnerReference for the data ConfigMap
+	var deployment appsv1.Deployment
+	utils.Retry(5, 1*time.Second, func() error {
```
leelavg, line 54:
```suggestion
	err = utils.Retry(5, 1*time.Second, func() error {
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,110 @@
...
+	// so terminating this program will cause the pod to keep getting restarted and needlessly running.
+	// To avoid this, a signal handler is set up to only terminate the program when the pod is getting deleted.
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM)
+	<-signalChan
+
+	log.Info("AWS data gather program terminating")
+}
+
+func gatherData(imdsServer string, deployment appsv1.Deployment, k8sClient client.Client) error {
```
leelavg, line 86: - maybe func rename is suited better as it's also creating configmap?
```suggestion
func gatherAndSaveData(imdsServer string, deployment appsv1.Deployment, k8sClient client.Client) error {
```

leelavg (commented):
- what issues are you facing currently for removing `wip`?

cmd/awsDataGather/main.go
leelavg: ack. thanks for checking.

pkg/aws/gather_test.go
leelavg: sure.

Dockerfile
```diff
@@ -16,19 +16,25 @@ COPY controllers/ controllers/
...
+COPY pkg/ pkg/
 
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
 
 # Build readiness probe binary
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o readinessServer readinessProbe/main.go
 
+# Build aws data gathering binary
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o awsDataGather/awsDataGather awsDataGather/main.go
```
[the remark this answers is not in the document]
renan-campos: Without the extra directory, the awsDataGather binary isn't getting copied over... I can build it into the bin directory instead if you'd like.

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,110 @@
...
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	log := ctrl.Log.WithName("main")
+
+	namespace, found := os.LookupEnv("NAMESPACE")
+	if !found {
+		fmt.Fprintf(os.Stderr,
+			"NAMESPACE environment variable not found\n")
+		os.Exit(1)
+	}
+	deploymentName, found := os.LookupEnv("DEPLOYMENT_NAME")
```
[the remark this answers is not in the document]
renan-campos: This was on Ohad's suggestion. So that the ConfigMap doesn't go away if the pod gets deleted or restarts, only when the deployment is deleted should the ConfigMap be removed.

cmd/awsDataGather/main.go
renan-campos: Thanks I missed this.

Dockerfile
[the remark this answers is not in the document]
renan-campos: It's an odd silent error of go... from the source directory you can recreate it by running something like this:
`go build -o cmd cmd/awsDataGather/main.go`
It won't return an error, but the binary won't be built either.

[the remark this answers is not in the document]
renan-campos: actually it'll build the binary as `cmd/main`

bindrad (approved):

renan-campos:
@nb-ohad the pr is blocked on your requested changes, which have been addressed.Do you have to clear the requested changes flag?

nb-ohad (changes requested):

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,123 @@
...
+	log.Info("AWS data gathering successfully completed!")
+
+	// Set up signal handler, so that a clean up procedure will be run if the pod running this code is deleted.
+	// This code is run as part of a deployment that is pushed to Kubernetes through an OLM Bundle.
+	// At this time, OLM Bundles do not support restartPolicy: OnFailure,
+	// so terminating this program will cause the pod to keep getting restarted and needlessly running.
+	// To avoid this, a signal handler is set up to only terminate the program when the pod is getting deleted.
+	signalChan := make(chan os.Signal, 1)
+	signal.Notify(signalChan, syscall.SIGTERM)
+	<-signalChan
```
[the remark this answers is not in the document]
renan-campos: It is used to block completion of main until the pod receives the SIGTERM signal during deletion. You may have missed the comment explaining the reason in the lines above this:
```
	// Set up signal handler, so that a clean up procedure will be run if the pod running this code is deleted.
	// This code is run as part of a deployment that is pushed to Kubernetes through an OLM Bundle.
	// At this time, OLM Bundles do not support restartPolicy: OnFailure,
	// so terminating this program will cause the pod to keep getting restarted and needlessly running.
	// To avoid this, a signal handler is set up to only terminate the program when the pod is getting deleted.
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,123 @@
...
+
+	k8sClient, err := client.New(config.GetConfigOrDie(), options)
+	if err != nil {
+		log.Error(err, "error creating client")
+		os.Exit(1)
+	}
+
+	// This will later be used as the OwnerReference for the data ConfigMap
+	var deployment appsv1.Deployment
+	err = utils.Retry(5, 1*time.Second, func() error {
```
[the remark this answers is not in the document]
renan-campos: After checking my notes, I can't find a reason for adding this. My guess was I put this in just in case some network hiccup caused the get call to fail. I'll remove it

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,123 @@
...
+		Kind:       "Deployment",
+		Name:       deployment.Name,
+		UID:        deployment.UID,
+	}
+	if err := imdsClient.CreateConfigMap(deployment.Namespace, owner); err != nil {
+		return fmt.Errorf("Failed to create configmap: %v", err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
renan-campos: So that this functionality could receive test-coverage. `main` looks at its environment to gather the parameters to call this function. Under test, some of the parameters are faked, and the behavior is checked. see `main_test.go`

utils/utils.go
```diff
@@ -67,3 +74,41 @@ func MapItems(source []string, transform func(string) string) []string {
...
+func DeploymentNameFromPodName(podName string) (string, error) {
+	//Pod names created from deployments follow the convention:
+	// <deployment_name>-<pod-template-hash>-<uid>
+	// Therefore, we can get the deployment_name by omitting the last hyphened two sections
+	tmpStr := strings.Split(podName, "-")
+	if len(tmpStr) < 3 {
+		return "", fmt.Errorf("invalid pod name")
+	}
+	return strings.Join(tmpStr[:len(tmpStr)-2], "-"), nil
+}
```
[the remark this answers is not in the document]
renan-campos: Readability:
```
 43 --deploymentName, err := utils.DeploymentNameFromPodName(podName)
 44 --if err != nil {
 45 ----fmt.Fprintf(os.Stderr,
 46 ------"Could not determine deployment name from pod name (%s)",
 47 ------podName)
 48 ----os.Exit(1)
 49 --}
``` 
Being called this way, it is clear what the function is doing without having to understand how it is doing it.

Your one liner does not cover deployment names with hyphens or ensures that the podName is valid.

controllers/managedocs_controller.go
```diff
@@ -1367,27 +1370,27 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
-			{
+			utils.RookResourceRequirements{
 				Name:     "csi-rbdplugin",
 				Resource: utils.GetResourceRequirements("csi-rbdplugin"),
 			},
-			{
+			utils.RookResourceRequirements{
 				Name:     "liveness-prometheus",
 				Resource: utils.GetResourceRequirements("liveness-prometheus"),
 			},
```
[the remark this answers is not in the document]
renan-campos: The linter fails if this is not done.

pkg/aws/imds_client.go
```diff
@@ -0,0 +1,92 @@
...
+		// Setting the owner of the configmap resource to this deployment that creates it.
+		// That way, the configmap will go away when the deployment does.
+		configMap.OwnerReferences = owners
+
+		configMap.Data = client.data
+
+		return nil
+	})
+	return err
+}
```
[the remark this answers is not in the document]
renan-campos: I made it a method for the client because we want a specifically named configmap. That name for the configmap is used here, and by the controller. Having the client take care of it provides a central place to define the name of that configmap.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
nb-ohad: Ack

pkg/aws/imds_client.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return nil
+}
+
+func (client *IMDSClient) CreateConfigMap(namespace string, owners ...metav1.OwnerReference) error {
+	configMap := corev1.ConfigMap{}
+	configMap.Name = DataConfigMapName
+	configMap.Namespace = namespace
+
+	_, err := ctrl.CreateOrUpdate(context.Background(), client.k8sClient, &configMap, func() error {
```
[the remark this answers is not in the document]
nb-ohad: Even after the change you are still using different contexts during a single pass. 
See https://github.com/renan-campos/ocs-osd-deployer/blob/f18bac040c4db324a599e357a3d295bc0ec63e6c/cmd/awsDataGather/main.go#L64 

Please move the creation of the background context to main and then provide/pass  it down for every operation that need a go context. If feel the need to specialize a context  (by creating a sub-context, e.g. a context with a timeout), for a specific operation, please do so. But make sure an entire pass is using the same root context.

cmd/awsDataGather/main.go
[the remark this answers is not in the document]
nb-ohad: @renan-campos I am not sure I understand the issue you are trying to solve here. If the pod is getting deleted the program will terminate anyway. Why do we need to track this via an OS signal?

cmd/awsDataGather/main.go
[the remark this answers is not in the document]
nb-ohad: ack

utils/utils.go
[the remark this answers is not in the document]
nb-ohad: The readability argument could be applied to every single line of code on main. That does not mean we need multiple single-op util functions. Better just put a comment on top of the line if you think an explanation is necessary.

But your second point still stands. In that case, I will move from the wasteful implementation using split/merge into something less wasteful, like the following:

```suggestion
func DeploymentNameFromPodName(podName string) (string, error) {
	//Pod names created from deployments follow the convention:
	// <deployment_name>-<replica-set-hash>-<pod-uid>
	// Therefore, we can get the deployment_name by omitting the last hyphened two sections

	var i int
	if i = strings.LastIndex(podName, "-"); i == -1 {
		return "", fmt.Errorf("invalid pod name")
	}
	if i = strings.LastIndex(podName[0:i], "-"); i == -1 {
		return "", fmt.Errorf("invalid pod name")
	}
	return podName[0:i], nil
}
```

controllers/managedocs_controller.go
```diff
@@ -1461,17 +1464,44 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
+			}
+
+			// Inserting the VPC Egress rule in front of all other egress rules.
+			// The order or rules matter
+			// https://docs.openshift.com/container-platform/4.10/networking/openshift_sdn/configuring-egress-firewall.html#policy-rule-order_openshift-sdn-egress-firewall
+			// Inserting this rule in the front ensures it comes before the EgressNetworkPolicyRuleDeny rule.
+			desired.Spec.Egress = append(
+				[]openshiftv1.EgressNetworkPolicyRule{
+					vpcEgressRule,
+				}, desired.Spec.Egress...)
```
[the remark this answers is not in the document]
nb-ohad: The updated code (https://github.com/red-hat-storage/ocs-osd-deployer/blob/f18bac040c4db324a599e357a3d295bc0ec63e6c/controllers/managedocs_controller.go#L1499-L1503) does not comply with the repo formatting. When doing a multiline func invocation each argument should have it's own line (or multi lines), as shown in the above suggestion

pkg/aws/imds_client.go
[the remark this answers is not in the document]
nb-ohad: But that will also be true for any kind of information sharing between the 2 sides. even a simple constant name could do it. The point is that the client right now is mixing 2 different concerns. 

One concern is fetching the needed info from AWS and the other is to persist that knowledge into k8s. These 2 concerns have nothing to do with each other, and the only piece of info needed to fetch is the name. 

Also, this client does not handle fetching of the data from k8s (for a good reason), so it is not usable for the controller at all, and even if you will provide such functionality I would argue it should go somewhere else. You do not want to force the users of this client to also start using specific k8s implementation.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
renan-campos: Sorry I missed this.

pkg/aws/imds_client.go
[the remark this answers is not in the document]
renan-campos: ack

cmd/awsDataGather/main.go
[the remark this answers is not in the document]
renan-campos: We only want the program to terminate when the pod is deleted. Having this signal handler here ensures that happens. If main wasn't blocked waiting for the signal, then it would complete and the pod would endlessly restart

utils/utils.go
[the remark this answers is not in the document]
renan-campos: ack

cmd/awsDataGather/main.go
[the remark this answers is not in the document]
renan-campos: After discussion in-call discussion: this signal handling won't be needed after switching to a looping structure for main.

pkg/aws/imds_client.go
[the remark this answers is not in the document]
renan-campos: After in-call discussion: The two concerns that this client is addressing will be decoupled.

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,145 @@
...
+			continue
+		}
+
+		// Reset the backoff counter since data gathering succeeded.
+		backoff = 1
+
+		log.Info("AWS data gathering successfully completed!")
+		time.Sleep(5 * time.Minute)
+
+	}
```
nb-ohad, line 97: 1. This is a very aggressive backoff ramp, lets move to backoff = 2 * backoff instead
2. 256 is 4.2 minutes which is not very round, I will suggest rounding up to 5 min (300 seconds)
3.  Could be written using a max operation
4. Resetting the backoff to 1 will result in a constant backoff in the current scheme (1 ^ 2 = 1)
5. There is no need to handle the sleep in  2 separate cases. It complicates the code
6. logging the sleep duration is also important for happy path

Adding a suggestion to address all of these problems at once:
```
const maxSleep = 300
```
```suggestion
 	backoff = 1;
	for {
		log.Info("Gathering AWS data")
		if err := gatherAndSaveData(utils.IMDSv1Server, deployment, k8sClient, mainContext); err == nil {
			log.Info("AWS data gathering successfully completed!")
			sleep = maxSleep
			backoff = 1
		} else {
			log.Error(err, "Failed to gather AWS data")
			sleep = backoff
			backoff = math.Max(2 * backoff, maxSleep)
		}
		log.Info("Sleeping for %d seconds before next fetch...", sleep)
		time.Sleep(sleep)
	}
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,145 @@
...
+
+	log.Info("Fetching AWS VPC IPv4 CIDR data")
+	cidr, err := utils.IMDSFetchIPv4CIDR(imdsServer)
+	if err != nil {
+		return fmt.Errorf("Failed to get VPC IPv4 CIDR: %v", err)
+	}
+
+	awsData := map[string]string{
+		utils.CIDRKey: cidr,
+	}
```
nb-ohad, line 111: This is used only once inside `CreateOrUpdate`. There is no reason to define this here when the first and only usage is there.
```suggestion
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,145 @@
...
+	// Setting the owner of the configmap resource to this deployment that creates it.
+	// That way, the configmap will go away when the deployment does.
+	// The version and kind are being manually inserted because the deployment struct doesn't have it.
+	// See: https://github.com/kubernetes/client-go/issues/861#issuecomment-686806279
+	owner := metav1.OwnerReference{
+		APIVersion: "apps/v1",
+		Kind:       "Deployment",
+		Name:       deployment.Name,
+		UID:        deployment.UID,
+	}
```
nb-ohad, line 123: This is used only once inside CreateOrUpdate. There is no reason to define this here when the first and only usage is there.
```suggestion
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,145 @@
...
+	configMap := corev1.ConfigMap{}
+	configMap.Name = utils.DataConfigMapName
+	configMap.Namespace = deployment.Namespace
+
+	_, err = ctrl.CreateOrUpdate(context, k8sClient, &configMap, func() error {
+		// Setting the owner of the configmap resource to this deployment that creates it.
+		// That way, the configmap will go away when the deployment does.
+		configMap.OwnerReferences = []metav1.OwnerReference{
+			owner,
+		}
```
nb-ohad, line 134:
```suggestion
		configMap.OwnerReferences = []metav1.OwnerReference{
			metav1.OwnerReference{
				APIVersion: "apps/v1",
				Kind:       "Deployment",
				Name:       deployment.Name,
				UID:        deployment.UID,
			},
		}
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,145 @@
...
+
+		configMap.Data = awsData
```
nb-ohad, line 136:
```suggestion
		configMap.Data = map[string]string{
			utils.CIDRKey: cidr,
		}
```

controllers/managedocs_controller.go
```diff
@@ -207,6 +207,8 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *c
 					}
 				} else if name == rookConfigMapName {
 					return true
+				} else if name == utils.DataConfigMapName {
```
nb-ohad, line 210: The name `DataConfigMapName` is very generic and does not convey intent
```suggestion
				} else if name == utils.IMDSConfigMapName {
```

controllers/managedocs_controller.go
```diff
@@ -1461,17 +1463,46 @@ func (r *ManagedOCSReconciler) reconcileRookCephOperatorConfig() error {
...
-		r.Log.Info("Non converged deployment, skipping reconcile for egress network policy")
-		return nil
-	}
 	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.egressNetworkPolicy, func() error {
 		if err := r.own(r.egressNetworkPolicy); err != nil {
 			return err
 		}
 		desired := templates.EgressNetworkPolicyTemplate.DeepCopy()
 
+		if r.DeploymentType != convergedDeploymentType {
```
nb-ohad, line 1472: Do we need the new code on both provider and consumer?

utils/imds.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+const (
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html
+	IMDSv1Server      = "http://169.254.169.254"
+	DataConfigMapName = "aws-data"
```
nb-ohad, line 11: As this is a public const that is going to be imported in other packages please name it with a name that identifies the content
```suggestion
	IMDSDataConfigMapName = "aws-data"
```

utils/imds.go
```diff
@@ -0,0 +1,45 @@
...
+	CIDRKey           = "vpc-cidr"
+)
+
+func IMDSFetchIPv4CIDR(imdsServerAddr string) (string, error) {
+	// This method needs an instance mac address to get the VPC ipv4 CIDR.
+	var mac string
+	err := Retry(3, 1*time.Second, func() error {
```
nb-ohad, line 18: We decided to  handle retry logic with backoff from the outside. Why do we need to introduce additional (up to 2 seconds) of sleep here?

utils/imds.go
```diff
@@ -0,0 +1,45 @@
...
+			return err
+		}
+		return nil
+	})
+	if err != nil {
+		return "", fmt.Errorf("failed to determine mac address of instance")
+	}
+
+	var cidr string
+	err = Retry(3, 1*time.Second, func() error {
```
nb-ohad, line 31: We decided to  handle retry logic with backoff from the outside. Why do we need to introduce additional (up to 2 seconds) of sleep here?

nb-ohad (changes requested):

controllers/managedocs_controller.go
renan-campos: yes, both modes should set the network policy.

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,138 @@
...
+		Name:      deploymentName,
+		Namespace: namespace,
+	}, &deployment)
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Failed to find deployment '%s' in namespace '%s'", deploymentName, namespace))
+		os.Exit(1)
+	}
+
+	var backoff time.Duration = 1
+	var maxSleep time.Duration = 5
```
nb-ohad, line 76: Please move this to the top and make it a const

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,138 @@
...
+	namespace, found := os.LookupEnv("NAMESPACE")
+	if !found {
+		fmt.Fprintf(os.Stderr,
+			"NAMESPACE environment variable not found\n")
+		os.Exit(1)
+	}
+	podName, found := os.LookupEnv("POD_NAME")
+	if !found {
+		fmt.Fprintf(os.Stderr,
+			"POD_NAME environment variable not found\n")
```
nb-ohad, line 37:
```suggestion
		fmt.Fprintf(os.Stderr, "POD_NAME environment variable not found\n")
```
or
```suggestion
		fmt.Fprintf(
			os.Stderr,
			"POD_NAME environment variable not found\n",
		)
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,138 @@
...
+func main() {
+	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
+	log := ctrl.Log.WithName("main")
+
+	mainContext := context.Background()
+
+	namespace, found := os.LookupEnv("NAMESPACE")
+	if !found {
+		fmt.Fprintf(os.Stderr,
+			"NAMESPACE environment variable not found\n")
```
nb-ohad, line 31:
```suggestion
		fmt.Fprintf(os.Stderr, "NAMESPACE environment variable not found\n")
```
or
```suggestion
		fmt.Fprintf(
			os.Stderr,
			"NAMESPACE environment variable not found\n"
		)
```

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,138 @@
...
+			sleep = maxSleep
+			// Reset the backoff counter since data gathering succeeded.
+			backoff = 1
+		} else {
+			log.Error(err, "Failed to gather AWS data")
+			sleep = backoff
+			backoff = 2 * backoff
+			if backoff > maxSleep {
+				backoff = maxSleep
+			}
```
nb-ohad, line 91: Out of curiosity. Why did you decide to not use `math.Max`?

controllers/managedocs_controller.go
nb-ohad: Ok

cmd/awsDataGather/main.go
nb-ohad, line 76: Please  move `maxSleep` to the top and make it a const

cmd/awsDataGather/main.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		log.Info("Sleeping for %d minutes before next the fetch...", sleep)
+		time.Sleep(sleep * time.Minute)
```
nb-ohad, line 94: This is supposed to be in Seconds not minutes. 
I guess this change got in by accident via a suggestion I made on a previous review.

```suggestion
		time.Sleep(sleep * time.Second)
```

nb-ohad (changes requested):

cmd/awsDataGather/main.go
renan-campos: math.Max takes a float and the format for backoff and maxSleep must be time.Duration (int64). This way is cleaner without having to make all of the conversions.

cmd/awsDataGather/main.go
renan-campos: I did this intentionally. Why does it need to be in seconds?
As it is written, backoff will go 1, 2, 4, 5 minutes.

renan-campos: I don't want to get stuck on this point, I changed it to seconds.

nb-ohad: Even a single minute is a very long time to wait when you have an error and wants to repeat.
Thank you for changing to seconds

cmd/awsDataGather/main.go
nb-ohad: Thanks for clarifying

nb-ohad (approved):

PR #182, opened 2022-05-12
Add dummyTemplates and update Makefile

bindrad:
Currently, we are generating and creating the CSVs in the Makefile and if we need to add additional CSV such as deployer CSV then the Makefile looks clumsy so I moved the templates into a new folder called dummyTemplates so that we can create resources that are required by deployer to run reconciles.

TLS for Prometheus change requires deployer CSV to be created when running `make run` command
Grafana secret is expected in openshift-monitoring namespace when running `make run` command


Makefile
```diff
@@ -64,31 +64,9 @@ run: generate fmt vet manifests export_env_vars
...
-		  "\n          template:" \
-		  "\n            metadata:" \
-		  "\n              labels:" \
-		  "\n                app: test" \
-		  "\n            spec:" \
-		  "\n              containers:" \
-		  "\n              - name: test" \
-		  "\n    strategy: deployment" | kubectl apply -f -; \
-	 	done
+	cd dummyTemplates/CSVs/ && kubectl apply -f . -n ${NAMESPACE}
```
leelavg, line 67: - still mentions `dummyTemplates` pls change to `makefileutils`
- pls update the commit description as well

leelavg:
hi @bindrad if this is need pls update or close the pr?

bindrad:
I think it is a need because `make run` command won't work as controller requires ocs-osd-deployer template with kube-rbac-proxy image

Makefile
```diff
@@ -64,31 +64,9 @@ run: generate fmt vet manifests export_env_vars
...
-		  "\n              labels:" \
-		  "\n                app: test" \
-		  "\n            spec:" \
-		  "\n              containers:" \
-		  "\n              - name: test" \
-		  "\n    strategy: deployment" | kubectl apply -f -; \
-	 	done
+	cd makefileutils/CSVs/ && kubectl apply -f . -n ${NAMESPACE}
+	kubectl create namespace openshift-monitoring --dry-run=client -o yaml | kubectl apply -f -
+	cd makefileutils/secrets/ && kubectl apply -f .
```
nb-ohad, line 69: in this case, why do we need to set up every single CRD via a separate file? why can't we just have a single file with all the mock resources and apply it

nb-ohad (changes requested):

Makefile
bindrad: We can have a single file with all the mock resources

nb-ohad: And let's find it a better place. Maybe adjacent to the makefile file itself.

nb-ohad (approved):

nb-ohad:
@bindrad please sqaush

bindrad:
@nb-ohad Done

nb-ohad (approved):

PR #184, opened 2022-05-12
GitHub actions configuration to run docker build

bindrad:

### DIFF
--- a/.github/workflows/dockerbuild.yaml
+++ b/.github/workflows/dockerbuild.yaml
@@ -1,0 +1,21 @@
+---
+name: image-builder
+
+on:
+  pull_request:
+    branches: ['*']
+
+jobs:
+  docker-build:
+    name: docker-build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set GO_VERSION
+        run: echo "GO_VERSION=$(grep -m 1 -r 'go' go.mod | awk '{print $2}')" >> $GITHUB_ENV
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Build operator docker image
+        run: make docker-build
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,8 @@ VERSION ?= 2.0.2
 # Default bundle image tag
 IMAGE_TAG_BASE ?= controller
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
+# Makefile uses sh shell by default but we want to use bash
+SHELL := /bin/bash
 # Options for 'bundle-build'
 ifneq ($(origin CHANNELS), undefined)
 BUNDLE_CHANNELS := --channels=$(CHANNELS)


.github/dockerbuild.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  docker-build:
+    name: docker-build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.16
```
nb-ohad, line 18: This is now defined in 2 places, here and in the make file.

Can we get the go-version from an env var, and set it from somewhere within the project so both make and the workflow with use it? It will also be helpful to allow us to override the go-version when running locally or dispatching manually

Makefile
```diff
@@ -3,6 +3,8 @@ VERSION ?= 2.0.2
 # Default bundle image tag
 IMAGE_TAG_BASE ?= controller
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
+# Makefile uses sh shell by default but we want to use bash
+SHELL := /bin/bash
```
nb-ohad, line 7: Why?

.github/dockerbuild.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+      - uses: actions/checkout@v3
```
nb-ohad, line 20: Please checkout first, so the entirety of the project can be used in the next steps

nb-ohad (changes requested):

Makefile
nb-ohad: How did it work up until now?

bindrad: I think on our local systems the env var SHELL is already set to bash shell. But virtual machine provided by github actions don't have SHELL env var set so by default Makefile uses sh shell. We can solve the issue by either setting it via Makefile or having a job in github actions that sets SHELL env var.
I think we should set it via Makefile because it will help to run test without issue if someone is using any other shell that doesn't support the source.

bindrad: https://www.gnu.org/software/make/manual/make.html#Choosing-the-Shell

.github/dockerbuild.yaml
bindrad: I don't think it is good to set the go version using the environment variable in go.mod and Dockerfile.

Makefile
nb-ohad: ok

.github/dockerbuild.yaml
nb-ohad: I think you misunderstand me. 
My ask is to read it from one of these sources and use it here, you are already checking-out the entire source repo so you can read it from the go.mod file

bindrad: Ahh... Yes, I misunderstood your suggestion. I'll update the PR with the necessary changes.

.github/workflows/dockerbuild.yaml
```diff
@@ -0,0 +1,23 @@
+---
+name: image-builder
+
+on:
+  push:
+    branches: ['*']
+  pull_request:
+    branches: ['*']
```
nb-ohad, line 8: Why both events? does it make sense?

bindrad: Yes, I can remove the push event as whenever the head of pull_request is updated the workflow will run. I added it to test it on my fork otherwise I need to raise a PR to test it.
Will update this

nb-ohad:
@bindrad Can you please squash, also for some reason I don't see a unit test rn indication on this PR. Shouldn't we expect one as this change introduce them?

bindrad:
> @bindrad Can you please squash, also for some reason I don't see a unit test rn indication on this PR. Shouldn't we expect one as this change introduce them?

@nb-ohad You can see the workflows in the action tab: https://github.com/red-hat-storage/ocs-osd-deployer/actions/runs/2409848716
You can even see it in the PR checks
image

nb-ohad (approved):

nb-ohad:
Ok, but the name of the PR is "GitHub actions configuration to run unit tests" while the actual code just ensures a build is successful. Am I missing something?

bindrad:
The actual code will run the unit tests and then build the deployer image. This way we can ensure that our Dockerfile has no issues along with unit tests passing.
https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/Makefile#L140

nb-ohad (approved):

PR #199, opened 2022-05-24
Added Cluster ID to sendgrid email template

akshayashu:
The new email template will look like this:

sendgrid

Note: I used a new library `github.com/openshift/api/config/v1` to be able to use `clusterVersions`

Signed-off-by: akkhanna <akkhanna@redhat.com>


controllers/managedocs_controller.go
```diff
@@ -170,6 +172,7 @@ type ManagedOCSReconciler struct {
 // +kubebuilder:rbac:groups="network.openshift.io",namespace=system,resources=egressnetworkpolicies,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="coordination.k8s.io",namespace=system,resources=leases,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="odf.openshift.io",namespace=system,resources=storagesystems,verbs=list;watch;delete
+// +kubebuilder:rbac:groups="config.openshift.io",resources=clusterversions,verbs=get;list;watch;delete;update;patch
```
bindrad, line 175: According to the use-case, we don't need to delete, update and patch the resource.

controllers/managedocs_controller.go
```diff
@@ -1030,6 +1033,19 @@ func (r *ManagedOCSReconciler) reconcilePrometheus() error {
...
+		err1 := r.UnrestrictedClient.Get(context.Background(), types.NamespacedName{Name: "version"}, clusterVersion)
+
+		if err1 != nil {
+			r.Log.Info(err1.Error())
+		} else {
+			r.Log.Info("Cluster ID - " + string(clusterVersion.Spec.ClusterID))
+
+			clusterID := string(clusterVersion.Spec.ClusterID)
+			desired.Spec.ExternalLabels["clusterid"] = clusterID
+		}
```
bindrad, line 1047:
```suggestion
		clusterVersion := &configv1.ClusterVersion{}
		if err := r.UnrestrictedClient.Get(context.Background(), types.NamespacedName{Name: "version"}, clusterVersion); err != nil {
			return err
		}
		clusterID := string(clusterVersion.Spec.ClusterID)
		desired.Spec.ExternalLabels["clusterid"] = clusterID
```
You should use `unrestrictedGet` function.
I don't think we need to log error and clusterID so I have removed it in the suggestion.

ezio-auditore (changes requested):
Please add go.mod and go.sum files to the commit.

controllers/managedocs_controller.go
bindrad: You can even return a meaningful error message along with err(use fmt.Errorf).

templates/customernotification.html
```diff
@@ -68,6 +68,9 @@
                         <br>
                 {{ end }}
                 <ul>
+                    {{if eq .Name "clusterid" }}
+                        <li>Cluster ID: {{ .Value }}</li>
+                    {{ end }}
```
bindrad, line 73: I think we should provide the clusterID in the start of email.
Based on your change, I think if we have multiple PVCs then it will print clusterID multiple times.

bindrad (changes requested):

templates/prometheus.go
```diff
@@ -42,6 +42,7 @@ var (
 
 var PrometheusTemplate = promv1.Prometheus{
 	Spec: promv1.PrometheusSpec{
+		ExternalLabels:         map[string]string{},
```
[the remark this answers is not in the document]
akshayashu: This will help us to add extra/external labels to our Alertmanager-labels for alerts. We are adding clusterID in it

templates/customernotification.html
akshayashu: This will break the email template, suppose we have the same alert for diff cluster/namespace
Acc to you, this will look like:

- clusterID1
- alertmessage

- pvc name1

- clusterID2
- pvc name2

bindrad: Customers will always receive separate mail for different clusters.

PR #111, opened 2021-11-26
Added operator-sdk target to Makefile

bindrad:
This will be helpful for the downstream build and will help to run the `make bundle` command if operator-sdk is not present on the system.

Signed-off-by: bindrad <dbindra@redhat.com>

PR #204, opened 2022-05-27
install and use local operator-sdk in Makefile targets

leelavg:
ref: #111

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # Current Operator version
 VERSION ?= 2.0.2
+OPERATOR_SDK_VERSION ?= v1.8.0
 # Default bundle image tag
 IMAGE_TAG_BASE ?= controller
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
@@ -29,7 +30,6 @@ endif
 
 OS = $(shell go env GOOS)
 ARCH = $(shell go env GOARCH)
-
 
 all: manager readinessServer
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -176,14 +178,27 @@ else
...
 endif
 
+OPERATOR_SDK = $(BINDIR)/operator-sdk
+operator-sdk:
+ifeq (,$(wildcard $(OPERATOR_SDK)))
+	@{ \
+	set -e ;\
+	mkdir -p $(dir $(OPERATOR_SDK)) ;\
+	curl -sSLo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/$(OPERATOR_SDK_VERSION)/operator-sdk_${OS}_${ARCH};\
+	chmod +x $(OPERATOR_SDK) ;\
```
nb-ohad, line 188: Don't we need to `chmod` the actual `OPERATOR_SDK` binary? doesn't `OPERATOR_SDK` point to the directory?

nb-ohad (changes requested):

Makefile
```diff
@@ -176,14 +178,27 @@ else
 KUSTOMIZE=$(shell which kustomize)
 endif
 
+OPERATOR_SDK = $(BINDIR)/operator-sdk
```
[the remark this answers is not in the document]
leelavg: not working for me 
```
operator-sdk:
	@@echo $(CURRDIR)
```

```
# make operator-sdk && echo 1

1
```

Makefile
leelavg: - it's path to the binary not to the directory

Makefile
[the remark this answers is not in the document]
ezio-auditore: you might need to set the CURRDIR value.
CURRDIR = $(shell pwd)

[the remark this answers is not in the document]
leelavg: - well, that's what I did `BINDIR=$(shell pwd)/bin` in the first place. 
- I believe Ohad might've gave the comment as `CURRDIR` will be available by default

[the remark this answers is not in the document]
nb-ohad: `pwd` is problematic because it will point to the directory where the make was invoked not where the makefile resided.
`CURDIR` is supposed to be a built-in and will always point to the directory where the make file resides.

I think the issue is that I misspelled CURDIR as CURRDIR in my last comment

Makefile
nb-ohad: ok

Makefile
```diff
@@ -176,14 +178,26 @@ else
...
-	operator-sdk generate kustomize manifests -q
+bundle: manifests kustomize operator-sdk
+	$(OPERATOR_SDK) generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
-	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --extra-service-accounts prometheus-k8s --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) $(BUNDLE_FLAGS)
+	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --extra-service-accounts prometheus-k8s --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) $(BUNDLE_FLAGS)
 	cp config/metadata/* $(OUTPUT_DIR)/metadata/
-	operator-sdk bundle validate $(OUTPUT_DIR)
+	$(OPERATOR_SDK) bundle validate $(OUTPUT_DIR)
+	sed -i "s/builder=.*/builder=operator-sdk-$(OPERATOR_SDK_VERSION)/" bundle.Dockerfile
```
nb-ohad, line 200: @leelavg Why do we need this? Up until now, some other mechanism was responsible to update the docker file annotations.

leelavg: some other mechanism --> I'm not sure what that mechanism does as here I'm only updating the bundle version to match sdk version as I was told there's an issue with the sdk version which updates builder version incorrectly.

- when we upgrade the sdk and no issues are observed i'll remove this line

``` diff
# gd bundle.Dockerfile 
diff --git a/bundle.Dockerfile b/bundle.Dockerfile
index bd3ac93..4789190 100644
--- a/bundle.Dockerfile
+++ b/bundle.Dockerfile
@@ -6,7 +6,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.8.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.7.1+git
 LABEL operators.operatorframework.io.metrics.mediatype.v1=metrics+v1
 LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v2
```

nb-ohad: You are fighting against the builder. If the builder wants `operator-sdk-v1.7.1+git` it should be `operator-sdk-v1.7.1+git` and not `operator-sdk-v1.8.0`

leelavg: ack, done.

nb-ohad (approved):

PR #205, opened 2022-05-27
ability to make bundle with required params

leelavg:
ref: #110

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 # Current Operator version
 VERSION ?= 2.0.2
 OPERATOR_SDK_VERSION ?= v1.8.0
+REPLACES ?= 2.0.1
 # Default bundle image tag
 IMAGE_TAG_BASE ?= controller
 BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
@@ -194,6 +195,12 @@ endif
 bundle: manifests kustomize operator-sdk
 	$(OPERATOR_SDK) generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	cd config/manifests/bases && \
+		rm -rf kustomization.yaml && \
+		$(KUSTOMIZE) create --resources ocs-osd-deployer.clusterserviceversion.yaml && \
+		$(KUSTOMIZE) edit add annotation --force 'olm.skipRange':">=0.0.1 <$(VERSION)" && \
+		$(KUSTOMIZE) edit add patch --name ocs-osd-deployer.v0.0.0 --kind ClusterServiceVersion \
+		--patch '[{"op": "replace", "path": "/spec/replaces", "value": "$(REPLACES)"}]'
 	$(KUSTOMIZE) build config/manifests | $(OPERATOR_SDK) generate bundle -q --extra-service-accounts prometheus-k8s --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) $(BUNDLE_FLAGS)
 	cp config/metadata/* $(OUTPUT_DIR)/metadata/
 	$(OPERATOR_SDK) bundle validate $(OUTPUT_DIR)
--- a/config/manifests/kustomization.yaml
+++ b/config/manifests/kustomization.yaml
@@ -1,4 +1,5 @@
 resources:
+- bases
 - ../default
 - ../samples
 - ../scorecard
--- a/config/manifests/bases/kustomization.yaml
+++ b/config/manifests/bases/kustomization.yaml
@@ -1,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ocs-osd-deployer.clusterserviceversion.yaml
+commonAnnotations:
+  olm.skipRange: '>=0.0.1 <2.0.2'
+patches:
+- patch: '[{"op": "replace", "path": "/spec/replaces", "value": "2.0.1"}]'
+  target:
+    kind: ClusterServiceVersion
+    name: ocs-osd-deployer.v0.0.0


Makefile
```diff
@@ -30,6 +30,8 @@ endif
 OS = $(shell go env GOOS)
 ARCH = $(shell go env GOARCH)
 
+OPERATOR_NAMESPACE ?=
```
nb-ohad, line 33: Why do we need `OPERATOR_NAMESPACE` when we already have `NAMESPACE`?

Makefile
```diff
@@ -30,6 +30,8 @@ endif
...
+REPLACES ?=
```
nb-ohad, line 34: Please move this definition immediately after `VERSION` definition (line2) as they are tightly coupled. Also, this should have a default value of the current version we are replacing

Makefile
```diff
@@ -179,11 +181,17 @@ endif
...
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	cd config/manifests/bases && sed -e "s/@REPLACES@/$(REPLACES)/" -e "s/@VERSION@/$(VERSION)/" kustomization-base.yaml > kustomization.yaml
+	cd config/manifests && $(KUSTOMIZE) edit set namespace $(OPERATOR_NAMESPACE)
 	$(KUSTOMIZE) build config/manifests | operator-sdk generate bundle -q --extra-service-accounts prometheus-k8s --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS) $(BUNDLE_FLAGS)
 	cp config/metadata/* $(OUTPUT_DIR)/metadata/
 	operator-sdk bundle validate $(OUTPUT_DIR)
+else
+	@echo Variables 'REPLACES' and 'OPERATOR_NAMESPACE' are mandatory for generating bundle
+endif
```
nb-ohad, line 194: You can skip this check as we the above requests will guarantee default values for both

config/manifests/bases/kustomization-base.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Kustomization
+resources:
+- ocs-osd-deployer.clusterserviceversion.yaml
+commonAnnotations:
+  olm.skipRange: '>=0.0.1 <@VERSION@'
+patches:
+- patch: '[{"op": "replace", "path": "/spec/replaces", "value": "@REPLACES@"}]'
+  target:
+    kind: ClusterServiceVersion
+    name: ocs-osd-deployer.v0.0.0
```
nb-ohad, line 11: Why do we need this `-base` file if we can already work on the actual file?

Makefile
```diff
@@ -179,11 +181,17 @@ endif
 # Generate bundle manifests and metadata, then validate generated files.
 .PHONY: bundle
 bundle: manifests kustomize
+ifneq ($(and $(REPLACES),$(OPERATOR_NAMESPACE)),)
 	operator-sdk generate kustomize manifests -q
 	cd config/manager && $(KUSTOMIZE) edit set image controller=$(IMG)
+	cd config/manifests/bases && sed -e "s/@REPLACES@/$(REPLACES)/" -e "s/@VERSION@/$(VERSION)/" kustomization-base.yaml > kustomization.yaml
```
nb-ohad, line 187: @leelavg I think this change came from a misunderstanding of the intant of my prev request. 
I would prefer to edit the actual file and not create a template to be edited by `sed`. Please use `yq` or any other tool (even `kustomize` itself if you find it easier) to edit the actual patch file.

Makefile
```diff
@@ -179,11 +181,17 @@ endif
...
+	cd config/manifests && $(KUSTOMIZE) edit set namespace $(OPERATOR_NAMESPACE)
```
nb-ohad, line 188: Why is this needed? NAMESPACE is already taken care of in `config/default/kustomization.yaml` and I would like to have a single source of truth regarding the namespace.

I think it will be useful to use this command to set the namespace that is already defined there.

config/manifests/kustomization.yaml
```diff
@@ -1,4 +1,8 @@
 resources:
+- bases
 - ../default
 - ../samples
 - ../scorecard
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: default
```
nb-ohad, line 8: Why do we need this change? It works perfectly without it

nb-ohad (changes requested):

Makefile
leelavg: removed.

Makefile
leelavg: done.

config/manifests/bases/kustomization-base.yaml
leelavg: removed

Makefile
leelavg: removed, not required.

config/manifests/kustomization.yaml
leelavg: removed.

Makefile
leelavg: - Let me explain why I did what I did and I don't want to use neither `sed` nor  introduce a new one `yq`

> 	cd config/manifests/bases && \
- normal cd op

>		rm -rf kustomization.yaml && \
- we can only update below resources/fields with `kustomize edit set`
```
image
label
nameprefix
namespace
namesuffix
replicas
```
- other op `kustomize edit add` will add to the existing resource patch file and since we can't override existing field in the patch file, we are removing it

>		$(KUSTOMIZE) create --resources $(BUNDLE_PACKAGE).clusterserviceversion.yaml && \
- this creates a patch file with resource section `$(BUNDLE_PACKAGE).clusterserviceversion.yaml` on which patches has to be applied not the CSV itself

>		$(KUSTOMIZE) edit add annotation --force 'olm.skipRange':">=0.0.1 <$(VERSION)" && \
- first change we want in a csv, i.e, to update the skipRange

>		$(KUSTOMIZE) edit add patch --name $(BUNDLE_PACKAGE).v0.0.0 --kind ClusterServiceVersion \
		--patch '[{"op": "replace", "path": "/spec/replaces", "value": "$(REPLACES)"}]'
- unlike `kustomize edit add annotation` we can't directly target the fields under spec so we are adding a strategic merge patchable json object

---
- Against sed:
1. If we want inplace replace, the resulting expression will be cryptic and may not be cross-platform
- Against yq: 
1. extra dependency, the field which we want to edit is a json obj inside string and it'll more or less result in same format as kustomize minus 2 steps (remove and create of kustomization.yaml)
2. those two steps may not cause us much friction while generating bundle and can get away with one less installable dependency

---
- All I'm doing is to clarify as much as possible and correct any of my misunderstandings how the tools work and I believe these changes doesn't need a lot of discussions

leelavg:
- rebased

leelavg:
- rebased, @nb-ohad could you pls take a re-look at this PR, thanks.

bindrad (approved):

Makefile
```diff
@@ -33,6 +34,8 @@ endif
 OS = $(shell go env GOOS)
 ARCH = $(shell go env GOARCH)
 
+BUNDLE_PACKAGE ?= ocs-osd-deployer
```
nb-ohad, line 37: @leelavg Why are we allowing the invoker to override the bundle package? Also if someone will run it with a different value the result will be unpredictable, as it will not create a new CSV with the new name, it will just create a patch for a resource that does not exist.

I would suggest just embedding the name `ocs-osd-deployer` in the two commands that use it.

nb-ohad (changes requested):

Makefile
leelavg: done.

nb-ohad (approved):

PR #206, opened 2022-06-01
GitHub actions configuration to run golangci-lint

bindrad:
This PR will enable us to run golangci-lint whenever a PR is raised or updated. The changes to code are based on the output of golangci-lint. 
Error before making changes to the code
```
readinessProbe/readiness/suite_test.go:51:2: `timeout` is unused (deadcode)
        timeout  = time.Second * 10
        ^
readinessProbe/readiness/suite_test.go:52:2: `duration` is unused (deadcode)
        duration = time.Second * 10
        ^
readinessProbe/readiness/suite_test.go:53:2: `interval` is unused (deadcode)
        interval = time.Millisecond * 250
        ^
readinessProbe/readiness/suite_test.go:96:15: Error return value is not checked (errcheck)
                go RunServer(k8sClient, types.NamespacedName{Name: ManagedOCSName, Namespace: TestNamespace}, ctrl.Log.WithName("readiness"))
                            ^
controllers/managedocs_controller.go:704:2: S1008: should use 'return subComponent.StorageCluster.State == v1.ComponentNotFound' instead of 'if subComponent.StorageCluster.State == v1.ComponentNotFound { return true }; return false' (gosimple)
        if subComponent.StorageCluster.State == v1.ComponentNotFound {
        ^
controllers/managedocs_controller.go:1183:3: ineffectual assignment to smtpSecretData (ineffassign)
                smtpSecretData := map[string][]byte{}
                ^
controllers/managedocs_controller.go:1253:7: ineffectual assignment to auth (ineffassign)
                var auth map[string][]byte = nil
                    ^
controllers/managedocs_controller.go:1254:7: ineffectual assignment to err (ineffassign)
                var err error = nil

```

### DIFF
--- a/.github/workflows/dockerbuild.yaml
+++ b/.github/workflows/dockerbuild.yaml
@@ -12,10 +12,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set GO_VERSION
-        run: echo "GO_VERSION=$(grep -m 1 -r 'go' go.mod | awk '{print $2}')" >> $GITHUB_ENV
-      - uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GO_VERSION }}
+      - name: Set up Go environment
+        uses: red-hat-storage/ocs-osd-deployer@main
       - name: Build operator docker image
         run: make docker-build
--- a/.github/workflows/golangci-lint.yaml
+++ b/.github/workflows/golangci-lint.yaml
@@ -1,0 +1,22 @@
+---
+name: golangci-lint
+
+on:
+  pull_request:
+    branches: 
+    - '*'
+
+jobs:
+  golangci-lint:
+    name: golangci-lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Go environment
+        uses: red-hat-storage/ocs-osd-deployer@main
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: v1.46.2
+          args: --timeout 3m --verbose
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -701,10 +701,7 @@ func (r *ManagedOCSReconciler) updateComponentStatus() {
 func (r *ManagedOCSReconciler) verifyComponentsDoNotExist() bool {
 	subComponent := r.managedOCS.Status.Components
 
-	if subComponent.StorageCluster.State == v1.ComponentNotFound {
-		return true
-	}
-	return false
+	return subComponent.StorageCluster.State == v1.ComponentNotFound
 }
 
 func (r *ManagedOCSReconciler) reconcileStorageCluster() error {
@@ -1180,13 +1177,12 @@ func (r *ManagedOCSReconciler) reconcileAlertmanagerConfig() error {
 			}
 		}
 
-		smtpSecretData := map[string][]byte{}
 		if r.smtpSecret.UID == "" {
 			if err := r.get(r.smtpSecret); err != nil {
 				return fmt.Errorf("Unable to get SMTP secret: %v", err)
 			}
 		}
-		smtpSecretData = r.smtpSecret.Data
+		smtpSecretData := r.smtpSecret.Data
 		smtpHost := string(smtpSecretData["host"])
 		if smtpHost == "" {
 			return fmt.Errorf("smtp secret does not contain a host entry")
@@ -1250,9 +1246,8 @@ func (r *ManagedOCSReconciler) reconcileK8SMetricsServiceMonitorAuthSecret() err
 		if err := r.own(r.k8sMetricsServiceMonitorAuthSecret); err != nil {
 			return err
 		}
-		var auth map[string][]byte = nil
-		var err error = nil
-		if auth, err = r.readGrafanaV1Secret(); err != nil {
+		auth, err := r.readGrafanaV1Secret()
+		if err != nil {
 			if errors.IsNotFound(err) {
 				r.Log.Info("Unable to find v1 grafana-datasources secret")
 				auth, err = r.readGrafanaV2Secret()
--- a/readinessProbe/readiness/suite_test.go
+++ b/readinessProbe/readiness/suite_test.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"path/filepath"
 	"testing"
-	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -46,11 +45,6 @@ import (
 const (
 	ManagedOCSName = "test-managedocs"
 	TestNamespace  = "default"
-
-	// Define utility constants for object names and testing timeouts/durations and intervals.
-	timeout  = time.Second * 10
-	duration = time.Second * 10
-	interval = time.Millisecond * 250
 )
 
 var cfg *rest.Config
@@ -93,7 +87,7 @@ var _ = BeforeSuite(func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(k8sClient).ToNot(BeNil())
 
-		go RunServer(k8sClient, types.NamespacedName{Name: ManagedOCSName, Namespace: TestNamespace}, ctrl.Log.WithName("readiness"))
+		go RunServer(k8sClient, types.NamespacedName{Name: ManagedOCSName, Namespace: TestNamespace}, ctrl.Log.WithName("readiness")) //nolint
 
 		ctx := context.Background()
 


.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Set GO_VERSION
+        run: echo "GO_VERSION=$(grep -m 1 -r 'go' go.mod | awk '{print $2}')" >> $GITHUB_ENV
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.45.2
```
Madhu-1, line 24: 1.46.2 is the latest release.

bindrad: Okay, I think I can point it to latest one using `latest`
Something like this
```
        with:
          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
          version: latest
```

Madhu-1: Yes we can do it but the only problem is you might need to fix linter issues if a new linter is added to the latest release of golangci-lint, it can block PRs from getting merged. if it's okay we can point to the latest.

bindrad: Ohh understood! I think we should stick to a specific version, if required we can update the version.

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,43 @@
...
+    name: golangci-lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set GO_VERSION
+        run: echo "GO_VERSION=$(grep -m 1 -r 'go' go.mod | awk '{print $2}')" >> $GITHUB_ENV
+      - uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.GO_VERSION }}
```
nb-ohad, line 19: These setup steps are repeating in multiple actions, can we refactor this out? I believe GitHub Actions has a construct for this

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,43 @@
+---
+name: golangci-lint
```
nb-ohad, line 2: Let's use a better display name, maybe `lint GO code`

controllers/managedocs_controller.go
```diff
@@ -1250,9 +1246,10 @@ func (r *ManagedOCSReconciler) reconcileK8SMetricsServiceMonitorAuthSecret() err
 		if err := r.own(r.k8sMetricsServiceMonitorAuthSecret); err != nil {
 			return err
 		}
-		var auth map[string][]byte = nil
-		var err error = nil
-		if auth, err = r.readGrafanaV1Secret(); err != nil {
+		// var auth map[string][]byte = nil
+		// var err error = nil
```
nb-ohad, line 1250: if we are not using it why not just remove it?

readinessProbe/readiness/suite_test.go
```diff
@@ -93,7 +87,7 @@ var _ = BeforeSuite(func() {
 		Expect(err).ToNot(HaveOccurred())
 		Expect(k8sClient).ToNot(BeNil())
 
-		go RunServer(k8sClient, types.NamespacedName{Name: ManagedOCSName, Namespace: TestNamespace}, ctrl.Log.WithName("readiness"))
+		go RunServer(k8sClient, types.NamespacedName{Name: ManagedOCSName, Namespace: TestNamespace}, ctrl.Log.WithName("readiness")) //nolint
```
nb-ohad, line 90: Is there a reason for `nolint` here?

nb-ohad (changes requested):

nb-ohad:
@bindrad Can we settle on 2 commits, first one to fix the lint issue, a second one to add the lining action

.github/workflows/golangci-lint.yaml
bindrad: Okay, I'll update it to `lint GO code`

nb-ohad: Thinking about it again, Is that even a valid name? maybe we want `lint-go` or `golint`

nb-ohad: Ok, so if there is prior work (blogs, articles) that already established a none name let revert back to it

nb-ohad: yes (resolving this thread)

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+          # Optional: if set to true then the all caching functionality will be complete disabled,
+          #           takes precedence over all other caching options.
+          # skip-cache: true
+
+          # Optional: if set to true then the action don't cache or restore ~/go/pkg.
+          # skip-pkg-cache: true
+
+          # Optional: if set to true then the action don't cache or restore ~/.cache/go-build.
+          # skip-build-cache: true
```
nb-ohad, line 40: Why do we need all of these comments?

readinessProbe/readiness/suite_test.go
nb-ohad: ok

.github/workflows/golangci-lint.yaml
bindrad: We don't need those comments, it is just for reference in case we need to enable something.
It is kind of a template that golangci-lint provides, we can remove it.

controllers/managedocs_controller.go
bindrad: Can we resolve this?

.github/workflows/golangci-lint.yaml
bindrad: Can we resolve this?

bindrad:
@nb-ohad This PR won't pass the golangci-lint and docker-build checks because we are using composite actions to do setup of go env. Composite action requires a file(actions.yaml) already present in the default branch of the repo(in our case main branch). The other actions using composite action will not work unless we have that file in the default branch of the repo.

bindrad:
The checks should pass after this PR #210 gets merged

.github/workflows/dockerbuild.yaml
```diff
@@ -12,10 +12,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set GO_VERSION
-        run: echo "GO_VERSION=$(grep -m 1 -r 'go' go.mod | awk '{print $2}')" >> $GITHUB_ENV
-      - uses: actions/setup-go@v3
-        with:
-          go-version: ${{ env.GO_VERSION }}
+      - name: Setup go env
```
nb-ohad, line 15: The verb for `setup` (none)  is `set up`. 
```suggestion
      - name: Set up Go environment 
```

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,40 @@
+---
+name: golangci-lint
+
+on:
+  pull_request:
+    branches: ['*']
```
nb-ohad, line 6: Strict YAML notation is preferable 
```suggestion
    branches: 
    - '*'
```

.github/workflows/golangci-lint.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  golangci-lint:
+    name: golangci-lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Setup go env
```
nb-ohad, line 15: The verb for setup (none) is set up.
```suggestion
      - name: Set up Go environment 
```

.github/workflows/golangci-lint.yaml
nb-ohad: Please remove them, an up to date reference can be found online

nb-ohad (changes requested):

.github/workflows/golangci-lint.yaml
nb-ohad: @Madhu-1 Please resolve if you are satisfied

.github/workflows/dockerbuild.yaml
bindrad: Updated in latest commit

.github/workflows/golangci-lint.yaml
bindrad: Updated in latest commit

nb-ohad (approved):

nb-ohad:
@bindrad please revert the order of commits. 
I would prefer for the head to always pass tests, in order to achieve that you need to introduce the fixes before introducing the new check

Madhu-1 (approved):

.github/workflows/golangci-lint.yaml
Madhu-1: i dont see any option to resolve it. approved the changes.

nb-ohad (approved):

PR #212, opened 2022-06-07
Fixed incorrect Sendgrid namespace

akshayashu:
Signed-off-by: akkhanna <akkhanna@redhat.com>

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -923,12 +923,19 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
 		}
 
 		alertRelabelConfig := []struct {
-			TargetLabel string `yaml:"target_label,omitempty"`
-			Replacement string `yaml:"replacement,omitempty"`
-		}{{
-			TargetLabel: "namespace",
-			Replacement: r.namespace,
-		}}
+			SourceLabels []string `yaml:"source_labels"`
+			TargetLabel  string   `yaml:"target_label,omitempty"`
+			Replacement  string   `yaml:"replacement,omitempty"`
+		}{
+			{
+				SourceLabels: []string{"namespace"},
+				TargetLabel:  "alertnamespace",
+			},
+			{
+				TargetLabel: "namespace",
+				Replacement: r.namespace,
+			},
+		}
 
 		config, err := yaml.Marshal(alertRelabelConfig)
 		if err != nil {
--- a/templates/customernotification.html
+++ b/templates/customernotification.html
@@ -38,7 +38,7 @@
         {{ range .Alerts.Firing }}
             {{ if .Labels.persistentvolumeclaim }}
                 <li>PVC Name: {{ .Labels.persistentvolumeclaim }}</li>
-                <li>Namespace: {{ .Labels.namespace }}
+                <li>Namespace: {{ .Labels.alertnamespace }}
             {{ end }}
         {{ end }}
 


controllers/managedocs_controller.go
```diff
@@ -911,12 +911,18 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
...
+			Replacement  string   `yaml:"replacement,omitempty"`
+		}{
+			{
+				SourceLabels: []string{"namespace"},
+				TargetLabel:  "alertnamespace",
+			},
+			{
+				TargetLabel: "namespace",
+				Replacement: r.namespace,
+			}}
```
leelavg, line 925:
```suggestion
			}
		}
```

controllers/managedocs_controller.go
```diff
@@ -911,12 +911,18 @@ func (r *ManagedOCSReconciler) reconcileAlertRelabelConfigSecret() error {
 		}
 
 		alertRelabelConfig := []struct {
-			TargetLabel string `yaml:"target_label,omitempty"`
-			Replacement string `yaml:"replacement,omitempty"`
-		}{{
-			TargetLabel: "namespace",
-			Replacement: r.namespace,
-		}}
+			SourceLabels []string `yaml:"source_labels"`
```
leelavg, line 914:
```suggestion
			SourceLabels []string `yaml:"source_labels,omitempty"`
```

leelavg (commented):
- could you pls add a line on how this change fixes the issue or what are you trying to achieve?
- I did some [reading](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) around relabel configs and I believe the correct solution is
``` go
// from 
{
			TargetLabel: "namespace",
			Replacement: r.namespace,
}

// to

{
                        SourceLabels: []string{"namespace"},
			TargetLabel: "alertnamespace",
			Replacement: r.namespace,
}
```
- or if no change to template is required simply adding `SourceLabels` will be good enough
- however I'm not into alerts so I may have misunderstood the required change and pls correct me if I'm wrong

controllers/managedocs_controller.go
akshayashu: Adding `omitreply` was not working, tests were failing and even the deployment too.

controllers/managedocs_controller.go
akshayashu: So the problem was for all the alerts that were raised from a different namespace, other than op-storage also have the same namespace i.e: `openshift-storage`.
Before this Relabel had just
```
TargetLabel: "namespace",
Replacement: r.namespace,
```
which was necessary so that our alertmanager can use the matchLabel of `namespace: openshift-storage`
Now what I have done this, just before relabelling the original namespace(alert original namespace) to openshift-storage(r.namespace), we are creating a new label with `alertnamespace` which can hold the value of original namespace and then we can relabel namespace.

bindrad: @akshayashu Can you add the details to the description of the PR? It would help for reference in the future.

bindrad (approved):

controllers/managedocs_controller.go
nb-ohad, line 925: Why are we moving the source namespace to alertnamespace? Can we just keep the sourcenamespace at `namespace` and have `r.namespace` be in the `alertnamepsace`. To me, it makes much more sense do as little changes to incoming metric data. If we need additional labels we just add them.

akshayashu: Our alertmanager uses the `namespace: openshift-storage`matcher to catch all the alerts, if we want to change `namespace: openshift-storage`, then we will also have to change `alertnamespace: openshift-storage` in alertmanager too

```
matchers:
    - namespace="openshift-storage"
```

nb-ohad: Changing the selector seems to me like a more reasonable change

nb-ohad: But in both cases, i don't think `alertnamespace` is a name that describes the data we are adding. If the only concern we have is that matcher why can't we just add our common= label `app=managed-ocs`?

akshayashu: Right now we don't have this label in the alert, but anyways we can add it if you suggest and we will keep `namespace` as it is

akshayashu: I think we cannot change the match label for alertmanager, we need to make changes in relabelConfig only.

akshayashu:
@nb-ohad rebased

nb-ohad (approved):

nb-ohad:
@akshayashu @leelavg @bindrad 
I have approved and going to merge this PR even though I think we should not rename the default namespace label, mainly as this solves an immediate issue.

We will take a further look into this subject as a tech debt

PR #213, opened 2022-06-07
Enable and configure GitHub action for yamllint

bindrad:


.github/workflows/conf/yamllint.yaml
```diff
@@ -0,0 +1,35 @@
...
+  key-ordering: disable
+  line-length: disable
+  new-line-at-end-of-file: enable
+  new-lines: enable
+  octal-values: disable
+  quoted-strings: disable
+  trailing-spaces: enable
+  truthy:
+    level: warning
+    allowed-values: ['true', 'false', 'on']
```
nb-ohad, line 35: Is there a reason why we are not starting with default configuration?

.github/workflows/yamllint.yaml
```diff
@@ -0,0 +1,17 @@
...
+on:
+  pull_request:
+    branches:
+    - '*'
+jobs:
+  yamllint:
+    name: yamllint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
nb-ohad, line 12: why `actions/checkout@v2` and not `actions/checkout@v3` like the other workflows?

nb-ohad (changes requested):

.github/workflows/yamllint.yaml
bindrad: The default config gives many errors such as the wrong indentation, missing document start "---" and lines too long. So I went through all the rules in the [default config](https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration) and changed the rules.

bindrad: I missed this update, I'll change it.

.github/workflows/conf/yamllint.yaml
bindrad: Please suggest if you want to enable/disable any rules.

nb-ohad: The entire point of a linter is to force guidelines/best practices. If we are using this linter we should match that linter guidelines. Any change to the linter needs to come with a valid reason.
No point to use a linter if we are not going to follow its practices.

leelavg: - I agree with @nb-ohad here 😅, as we are pulling in a new linter to check against best practices a minimal omissions are fine however I see many in here and tracking these many is also a hassle

coming to two points raised above:
- `---`  generally indicates start of a yaml file/section and `***` marks the end
- it's best practice to indent lists with two/four leading empty spaces but not at the same level of it's parent (arguable however yaml is a space de-limited spec, so we better follow it throughout)

ps: IMHO it's not your task alone to fix all linting issues, let me know or some other who can help you on this (or there might be an automated tool to fix linting issues 🤔)

bindrad: Okay, I'll update it to the default configuration

bindrad: @leelavg @nb-ohad Do you think we should line-length rule enabled?
https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.line_length

leelavg: it's arguable, we can have the default length unless we don't have any automated tool to format yaml or else we can relax it to 100 columns or so

bindrad: - I'm not able to find any automated tool to format existing YAML
- Relaxing the limit to 100 won't solve all the errors as there are files that have more than 100 characters in a single line

nb-ohad: @bindrad @leelavg I don't see value in line length limiter mainly because screen resolution differs considerably between monitors and not utilizing the entire monitor width is annoying to anyone who have a hi-res one

I am ok with disabling that one with a comment.

bindrad: The files that are being reverted are
- https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/config/crd/bases/ocs.openshift.io_managedocs.yaml
- https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/config/rbac/role.yaml

nb-ohad: @bindrad Why are we linting files that are produced by an automated process for its own internal use?

bindrad: I thought that we should lint these files as we use them to generate the bundle

leelavg: no, we shouldn't be linting generated files

nb-ohad: Lets skip linting on generate files please

nb-ohad: For now, just skip the config folder

bindrad: Updated in the latest commit

leelavg (approved):

nb-ohad (approved):

PR #214, opened 2022-06-09
Added NetworkPolicyReconcileStrategy to ManagedOCS Spec and Status

ezio-auditore:
When NetworkPolicyReconcileStrategy is set to `none` the deployer will not reconcile any NetworkPolicies as well as set the status of NetworkPolicyReconcileStrategy in ManagedOCS CR as `none`
Sample CR
```
Kind: Managedocs
...
spec:
  networkPolicyReconcileStrategy: none
 ...
 status:
 networkPolicyReconcileStrategy: none
```
Jira: https://issues.redhat.com/browse/RHSTOR-3363

Signed-off-by: Kaustav Majumder <kmajumde@redhat.com>

### DIFF
--- a/api/v1alpha1/managedocs_types.go
+++ b/api/v1alpha1/managedocs_types.go
@@ -35,7 +35,8 @@ const (
 
 // ManagedOCSSpec defines the desired state of ManagedOCS
 type ManagedOCSSpec struct {
-	ReconcileStrategy ReconcileStrategy `json:"reconcileStrategy,omitempty"`
+	ReconcileStrategy              ReconcileStrategy `json:"reconcileStrategy,omitempty"`
+	NetworkPolicyReconcileStrategy ReconcileStrategy `json:"networkPolicyReconcileStrategy,omitempty"`
 }
 
 type ComponentState string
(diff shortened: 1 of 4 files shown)


api/v1alpha1/managedocs_types.go
```diff
@@ -35,7 +35,8 @@ const (
 
 // ManagedOCSSpec defines the desired state of ManagedOCS
 type ManagedOCSSpec struct {
-	ReconcileStrategy ReconcileStrategy `json:"reconcileStrategy,omitempty"`
+	ReconcileStrategy              ReconcileStrategy `json:"reconcileStrategy,omitempty"`
+	NetworkPolicyReconcileStrategy ReconcileStrategy `json:"networkPolicyReconcileStrategy,omitempty"`
```
nb-ohad, line 39: I think this will become very messy very fast. 
One option is to elevate the original `ReconcileStrategy` to a "master switch",  then have a new sub-spec struct containing specific reconcile switches, based on the different aspects that we would like the control on. 
For now, these will be ODFDeployment and NetworkPolicies

nb-ohad (changes requested):

api/v1alpha1/managedocs_types.go
ezio-auditore: Do you mean something like this?
```
...
type ManagedOCSSpec struct{
  MasterReconcileStrategy ReconcileStrategy
  ComponentReconcileStrategy ComponentReconcileStrategy
  }
  ----
  
  type ComponentReconcileStrategy struct{
   ODFDeployment ReconcileStrategy
   NetworkPolicies ReconcileStrategy
 }
 ```
  
  

 ```

nb-ohad: Something like that, but we cannot rename `reconcilePolicy` to `mainReconcilePolicy` as this is a breaking CR change

ezio-auditore:
Not relevant anymore.

PR #216, opened 2022-06-15
Bump from manifests to 2.0.3 

leelavg:

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -31,8 +31,8 @@ rules:
   - clusterversions
   verbs:
   - get
-  - watch
   - list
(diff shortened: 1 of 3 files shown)


config/rbac/role.yaml
```diff
@@ -31,8 +31,8 @@ rules:
   - clusterversions
   verbs:
   - get
-  - watch
   - list
+  - watch
```
leelavg, line 35: - Maybe one of the previous PRs was manually edited and so is the change when roles are generated via `make bundle` now.

bindrad (approved):

nb-ohad (approved):

PR #225, opened 2022-06-28
use bearerTokenFile for federating metrics into addon ns

leelavg:
- remove usage of grafana secret
- remove reconcile phase for creating authentication secret based on
  grafana secret
- remove corresponding unit test mocks and tests
- remove unused `unrestrictedGet` method on reconciler

### DIFF
--- a/controllers/managedocs_controller.go
+++ b/controllers/managedocs_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"math"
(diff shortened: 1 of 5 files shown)


leelavg:
- verified on ocp 4.11 will open for review after verifying in 4.10 (need to check on 4.8 as well?)

bindrad:
I think you need to even remove the creation of grafana secret and monitoring namespace when `make run` command is executed.
https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/makefileutils.yaml#L75-L88

controllers/managedocs_controller.go
```diff
@@ -68,45 +67,41 @@ const (
...
-	providerApiServerNetworkPolicyName     = "provider-api-server-rule"
-	prometheusProxyNetworkPolicyName       = "prometheus-proxy-rule"
-	monLabelKey                            = "app"
-	monLabelValue                          = "managed-ocs"
-	rookConfigMapName                      = "rook-ceph-operator-config"
-	k8sMetricsServiceMonitorName           = "k8s-metrics-service-monitor"
-	grafanaDatasourceSecretName            = "grafana-datasources"
-	grafanaDatasourceSecretKey             = "prometheus.yaml"
-	k8sMetricsServiceMonitorAuthSecretName = "k8s-metrics-service-monitor-auth"
-	openshiftMonitoringNamespace           = "openshift-monitoring"
```
leelavg, line 103: only these are removed.

controllers/managedocs_controller.go
```diff
@@ -127,30 +122,29 @@ type ManagedOCSReconciler struct {
...
-	prometheusService                  *corev1.Service
-	dmsRule                            *promv1.PrometheusRule
-	alertmanager                       *promv1.Alertmanager
-	pagerdutySecret                    *corev1.Secret
-	deadMansSnitchSecret               *corev1.Secret
-	smtpSecret                         *corev1.Secret
-	alertmanagerConfig                 *promv1a1.AlertmanagerConfig
-	alertRelabelConfigSecret           *corev1.Secret
-	k8sMetricsServiceMonitor           *promv1.ServiceMonitor
-	k8sMetricsServiceMonitorAuthSecret *corev1.Secret
```
leelavg, line 148: only this one is removed

controllers/suite_test.go
```diff
@@ -65,20 +65,17 @@ var (
...
-	testAddonParamsSecretName                  = "test-addon-secret"
-	testPagerdutySecretName                    = "test-pagerduty-secret"
-	testDeadMansSnitchSecretName               = "test-deadmanssnitch-secret"
-	testSMTPSecretName                         = "test-smtp-secret"
-	testAddonConfigMapName                     = "test-addon-configmap"
-	testAddonConfigMapDeleteLabelKey           = "test-addon-configmap-delete-label-key"
-	testDeployerCSVName                        = "ocs-osd-deployer.x.y.z"
-	testGrafanaFederateSecretName              = "grafana-datasources"
-	testK8sMetricsServiceMonitorAuthSecretName = "k8s-metrics-service-monitor-auth"
-	testOpenshiftMonitoringNamespace           = "openshift-monitoring"
```
leelavg, line 79: these 3 are removed

leelavg (commented):
- some of the resultant diff is due to `gofmt` aligning vars, so commented about what vars are actually removed

bindrad (approved):

templates/k8smetricsservicemonitor.go
```diff
@@ -35,7 +34,7 @@ var params = map[string][]string{
 	},
 }
 
-var k8sMetricsServiceMonitorAuthSecret = "k8s-metrics-service-monitor-auth"
+var bearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
```
nb-ohad, line 37: First, this is a const, not a var.
Second, this value is used only once in the template, and nowhere else. 
Let's embed it in the template!

nb-ohad (changes requested):

nb-ohad:
@leelavg I see you use the service account token. Does this mean that all service accounts have access to in-cluster monitoring prom?

templates/k8smetricsservicemonitor.go
leelavg: agreed! done.

leelavg:
> @leelavg I see you use the service account token. Does this mean that all service accounts have access to in-cluster monitoring prom?

- yes, seem to be the case. From in-cluster anything including and above `default` sa (for direct internal prom proxy) and from outside of cluster `oc whoami -t` token (for a route) can access metric data.

nb-ohad:
@leelavg The image-builder requirement is failing

leelavg:
> @leelavg The image-builder requirement is failing

- The failure doesn't seem genuine, it's passing for me locally and mostly a contention during update of a resource
- Not related but I just changed below in recent force push
``` diff
diff --git a/controllers/managedocs_controller_test.go b/controllers/managedocs_controller_test.go
index 19e7908..e647f5b 100644
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -290,7 +290,7 @@ var _ = Describe("ManagedOCS controller", func() {
        }
        k8sMetricsServiceMonitorTemplate := promv1.ServiceMonitor{
                ObjectMeta: metav1.ObjectMeta{
-                       Name:      "k8s-metrics-service-monitor",
+                       Name:      k8sMetricsServiceMonitorName,
                        Namespace: testPrimaryNamespace,
                },
        }
```

leelavg:
- now the CI failed with a different error
- nevertheless, two times failure isn't a good sign (was passing before rebase on rhobs pr 🤔)
- will take a re-look and get back at the earliest

leelavg:
> @leelavg I see you use the service account token. Does this mean that all service accounts have access to in-cluster monitoring prom?

- on another note this might be the reason we're able to access the ep https://github.com/red-hat-storage/ocs-osd-deployer/blob/main/config/rbac/k8s_metrics_sm_role.yaml, I didn't see `/federate` ep specifically, so a little bit unsure.

leelavg:
failure 1:
- `servicemonitors.monitoring.coreos.com "k8s-metrics-service-monitor" not found`
- before reconciler tries to create this resource test tried to get it

failure 2:
- `Operation cannot be fulfilled on servicemonitors.monitoring.coreos.com "k8s-metrics-service-monitor": the object has been modified; please apply your changes to the latest version and try again`
- resource creation by reconciler and modification by test occured at same time

---
- earlier we were creating a resource (grafana secret) in test and then wait for monitor to be created by reconciler
- since we do not depend on secret we need to confirm presence of monitor creation of reconciler before we modify the resource
- change made to recent push:
``` diff
diff --git a/controllers/managedocs_controller_test.go b/controllers/managedocs_controller_test.go
index e647f5b..a6194dd 100644
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -1338,6 +1338,9 @@ var _ = Describe("ManagedOCS controller", func() {
                })
                When("k8sMetricsServiceMonitor is modified", func() {
                        It("should revert the changes and bring the resource back to its managed state", func() {
+                               // Wait for the resource to be created
+                               utils.WaitForResource(k8sClient, ctx, k8sMetricsServiceMonitorTemplate.DeepCopy(), timeout, interval)
+
                                sm := k8sMetricsServiceMonitorTemplate.DeepCopy()
                                smKey := utils.GetResourceKey(sm)
                                Expect(k8sClient.Get(ctx, smKey, sm)).Should(Succeed())
```

nb-ohad (approved):

PR #229, opened 2022-07-04
eventually update resource before checking state change

leelavg:

### DIFF
--- a/controllers/managedocs_controller_test.go
+++ b/controllers/managedocs_controller_test.go
@@ -1356,18 +1356,23 @@ var _ = Describe("ManagedOCS controller", func() {
 		When("k8sMetricsServiceMonitor is modified", func() {
 			It("should revert the changes and bring the resource back to its managed state", func() {
 				// Wait for the resource to be created
-				utils.WaitForResource(k8sClient, ctx, k8sMetricsServiceMonitorTemplate.DeepCopy(), timeout, interval)
-
 				sm := k8sMetricsServiceMonitorTemplate.DeepCopy()
-				smKey := utils.GetResourceKey(sm)
-				Expect(k8sClient.Get(ctx, smKey, sm)).Should(Succeed())
+				utils.WaitForResource(k8sClient, ctx, sm, timeout, interval)
+
 				// Update the spec
 				spec := sm.Spec.DeepCopy()
 				sm.Spec = promv1.ServiceMonitorSpec{
 					Endpoints: []promv1.Endpoint{},
 				}
-				Expect(k8sClient.Update(ctx, sm)).Should(Succeed())
+
+				// Contention of test and reconilcer updating the resource is observed
+				// try for successful update of resource before checking for revert
+				Eventually(func() error {
+					return k8sClient.Update(ctx, sm)
+				}, timeout, interval).Should(Succeed())
+
 				// Wait for the spec changes to be reverted
+				smKey := utils.GetResourceKey(sm)
 				Eventually(func() *promv1.ServiceMonitorSpec {
 					sm := k8sMetricsServiceMonitorTemplate.DeepCopy()
 					Expect(k8sClient.Get(ctx, smKey, sm)).Should(Succeed())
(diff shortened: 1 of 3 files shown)


controllers/managedocs_controller_test.go
```diff
@@ -1349,7 +1349,14 @@ var _ = Describe("ManagedOCS controller", func() {
...
 					Endpoints: []promv1.Endpoint{},
 				}
-				Expect(k8sClient.Update(ctx, sm)).Should(Succeed())
+
+				// Contention of test and reconilcer updating the resource is observed
+				// try to get the update from test succeed before polling for resource change
+				Eventually(func() bool {
+					err := k8sClient.Update(ctx, sm)
+					return err == nil
+				}, timeout, interval).Should(BeTrue())
```
nb-ohad, line 1358: I would prefer to have to use `consistently` on the get to validate a stable resource version.

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
leelavg: makes sense, done.

controllers/managedocs_controller_test.go
```diff
@@ -1343,7 +1343,13 @@ var _ = Describe("ManagedOCS controller", func() {
 
 				sm := k8sMetricsServiceMonitorTemplate.DeepCopy()
 				smKey := utils.GetResourceKey(sm)
-				Expect(k8sClient.Get(ctx, smKey, sm)).Should(Succeed())
+
+				// Contention of test and reconilcer updating the resource is observed
+				// try to get the resource consistently before updating the resource
+				Consistently(func() error {
+					return k8sClient.Get(ctx, smKey, sm)
```
nb-ohad, line 1350: I think you need to make sure that the resource version is stable, without it you don't know if updates are done or not. Right now this get will succeed on the first try and is equivalent to the original line.

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
leelavg: - Ack, I'm now reading about the matchers.
- Will make the changes as requested

leelavg: - I think it doesn't exist in first successful try
```
In the case of Consistently, Gomega polls the the input repeatedly and asserts the matcher is satisfied every time. Consistently only exits early if a failure occurs - otherwise it continues polling until the specified interval elapses. This is often the only way to assert that something "does not happen" in an asynchronous system.
```

nb-ohad (changes requested):

controllers/managedocs_controller_test.go
```diff
@@ -1339,17 +1339,22 @@ var _ = Describe("ManagedOCS controller", func() {
 		When("k8sMetricsServiceMonitor is modified", func() {
 			It("should revert the changes and bring the resource back to its managed state", func() {
 				// Wait for the resource to be created
-				utils.WaitForResource(k8sClient, ctx, k8sMetricsServiceMonitorTemplate.DeepCopy(), timeout, interval)
-
 				sm := k8sMetricsServiceMonitorTemplate.DeepCopy()
+				utils.WaitForResource(k8sClient, ctx, sm, timeout, interval)
 				smKey := utils.GetResourceKey(sm)
```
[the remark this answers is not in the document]
nb-ohad: I don't care about diff size. 
As of current implementation, this var is not used until the end of the function and I would like to have vars defined near (and in the scope) of the code that uses them.

nb-ohad (approved):

leelavg:
- @nb-ohad the test reconciler is running too fast, not sure you'll like it or not, we may need to increase the duration https://github.com/red-hat-storage/ocs-osd-deployer/blob/3fda30/controllers/suite_test.go#L194
- I'm thinking of 100ms (ideally should be greater than 250ms used in the test interval)

nb-ohad:
This shouldn't matter! if the reconciler does not introduce a change to a CR, the result is a no-op and there is no contention.
If you see different behavior then the underlying assumption is wrong.

leelavg:
- Not being able to reproduce failure locally is creating a challenge to fix CI failure
- Lastly, I tried the matcher as `BeNil` rather than `BeTrue`

nb-ohad (approved):

nb-ohad:
@leelavg please rebase

leelavg:
- no, I wanted a consistent atleast 5 pass runs in CI, untill then I'll mark it as draft

PR #226, opened 2022-06-28
Add resource limit and request for prom kube-rbac-proxy

bindrad:
Signed-off-by: Dhruv Bindra <dbindra@redhat.com>

PR #232, opened 2022-07-15
drop alert federation in favour of RHOBS remote write

leelavg:
- soft revert of #149, #175, #226


leelavg:
- do we need a starter code to remove unused resources in the running cluster which'll run once on reconcile?

bindrad:
We enabled TLS for Prometheus for two reasons:
1. Federation
2. ODF console
I think we should keep it as it is for the ODF console that will be enabled when we depend on ODF v4.11

ezio-auditore:
@leelavg Alert federation should still be in place for consumer clusters, your PR should reflect only alert federation removal for provider cluster.
The reason, we will not be pushing Alerts from consumer clusters only provider ones.

PR #243, opened 2022-08-22
Use custom rule file.

rexagod:
Use custom rule file for managed-ocs, to allow it to diverge from OCS
from the POV of documentation links (and any such changes that may come
in the future, which would still allow it to be independent of OCS rule
file changes, and only dependent on Ceph's rules).

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

### DIFF
--- a/templates/alertmanagerconfig.go
+++ b/templates/alertmanagerconfig.go
@@ -44,7 +44,7 @@ var pagerdutyAlerts = []string{
 	"CephClusterWarningState",
 	"CephOSDVersionMismatch",
 	"CephMonVersionMismatch",
-	"CephOSDFlapping",
+	"CephOSDFlapping", // Present.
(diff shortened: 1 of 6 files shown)


templates/alertmanagerconfig.go
```diff
@@ -44,7 +44,7 @@ var pagerdutyAlerts = []string{
 	"CephClusterWarningState",
 	"CephOSDVersionMismatch",
 	"CephMonVersionMismatch",
-	"CephOSDFlapping",
+	"CephOSDFlapping", // Present.
```
rexagod, line 47: Need to include OCS alerts in the rule file.

rexagod:
Pivoted: https://github.com/red-hat-storage/ocs-osd-deployer/pull/246.

PR #246, opened 2022-08-22
Remove unwanted annotations from alerts

rexagod:
Remove unwanted annotations from alerts and skip reconciliation of the
prometheus (ceph) rule file by the ceph cluster in OCS.

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>


rexagod:
Also PTAL: https://github.com/red-hat-storage/ocs-operator/pull/1777.

controllers/managedocs_controller.go
```diff
@@ -1338,6 +1338,25 @@ func (r *ManagedOCSReconciler) reconcileMonitoringResources() error {
...
+				group := &groups[groupIndex]
+				for ruleIndex := range group.Rules {
+					rule := &group.Rules[ruleIndex]
+					if rule.Alert != "" {
+						for _, annotation := range unwantedRuleAnnotations {
+							delete(rule.Annotations, annotation)
+						}
+					}
+				}
+			}
```
leelavg, line 1356: 1. What component is setting this annotation in the first place?
2. What happens if we don't remove this annotation, does the side effect of not removing this annotation justify the bulk of reconcile code here for removing this single annotation?

controllers/managedocs_controller.go
```diff
@@ -1338,6 +1338,25 @@ func (r *ManagedOCSReconciler) reconcileMonitoringResources() error {
...
+			// disallow prometheus rule reconciliation by ceph cluster
+			utils.AddAnnotation(obj, "managed", "true")
```
leelavg, line 1358: - we already have a standard key & value for monitoring resources, I believe it's best to re-use that for annotations as well
- https://github.com/red-hat-storage/ocs-osd-deployer/blob/6bba26e7609257de451f66878a3c9ce1fe674170/controllers/managedocs_controller.go#L95-L96

leelavg (commented):
- could you pls add a line or two about why this change is required to the commit msg
- i wasn't following the monitoring meetings and maybe I'm missing the context here 🤔

controllers/managedocs_controller.go
rexagod: Sure, I'll remove this (since its already added in `reconcileMonitoringResources`) and make the commit message more details.

controllers/managedocs_controller.go
rexagod: 1. The `documentation` annotation will be set in the OCS Ceph rules by default, following the Ceph rule format: https://github.com/ceph/ceph/blob/main/monitoring/ceph-mixin/prometheus_alerts.yml#L28.
2. Yep, if this is not removed, the receivers will send out notifications to managed services users having this links in them, which is something we absolutely want to avoid.

controllers/managedocs_controller.go
rexagod: Done.

controllers/managedocs_controller.go
leelavg: - ack, thanks. In that scenario, I believe relabelling the alert is more natural rather than editing the rule itself 🤔
- I do understand that in the linked PR you are trying not to reconcile specific annotated rules in the product, however even for verifying the absence of that annotation we need to drill through multiple stanza's which doesn't sound optimal
- Either come up with a relabel rule just before sending out the alerts or save the state that you removed the annotation and product will not reconcile that
- Nevertheless, I'd like to wait for others to weigh in their thoughts

rexagod: Yeah, relabelling should work as well. +1.

> Either come up with a relabel rule just before sending out the alerts or save the state that you removed the annotation and product will not reconcile that

Makes sense.

templates/prometheus.go
```diff
@@ -155,6 +155,10 @@ var PrometheusTemplate = promv1.Prometheus{
 						Regex:        getRelableRegex(alerts, metrics),
 						Action:       "keep",
 					},
+					{
+						Regex:  "^documentation$",
+						Action: "labeldrop",
```
rexagod, line 160: Note to self: Need to add `documentation` field as a label on in Ceph OCS rules instead of an annotation.

rexagod: Also need to test this out.

templates/prometheus.go
```diff
@@ -155,6 +155,10 @@ var PrometheusTemplate = promv1.Prometheus{
...
+					},
```
nb-ohad, line 161: This relabel will only affect remote writes, we need to drop the documentation from the time series inside our prom to prevent the doc links from being visible in the OCS local dashboard

nb-ohad (changes requested):

templates/prometheus.go
rexagod: Right, relabelling won't work in that case. Should I revert back to the earlier commit?

rexagod:
JFYI Updated: https://github.com/red-hat-storage/ocs-operator/pull/1777/files.

templates/prometheus.go
nb-ohad: Why not relabel upon scraping? I believe we did it somewhere else as well?
@ezio-auditore Can you help here?

rexagod: I'll make the changes based on [this](https://grafana.com/blog/2022/03/21/how-relabeling-in-prometheus-works/).

```
First off, the relabel_configs key can be found as part of a scrape job definition. These relabeling steps are applied before the scrape occurs and only have access to labels added by Prometheus’ Service Discovery. They allow us to filter the targets returned by our SD mechanism, as well as manipulate the labels it sets.

Once the targets have been defined, the metric_relabel_configs steps are applied after the scrape and allow us to select which series we would like to ingest into Prometheus’ storage.

Finally, the write_relabel_configs block applies relabeling rules to the data just before it’s sent to a remote endpoint. This can be used to filter metrics with [high cardinality](https://grafana.com/blog/2022/02/28/how-to-manage-cardinality-with-out-of-the-box-dashboards-in-grafana-cloud/?pg=blog&plcmt=body-txt) or route metrics to specific remote_write targets.
```

ezio-auditore: yes @rexagod you will need to relable while it is being scraped.Otherwise it is only going to affect the remote write data.

PR #248, opened 2022-08-25
Egress Firewall

rewantsoni:


nb-ohad:
@rewantsoni If you want an in-progress PR, please use a draft PR

controllers/managedocs_controller.go
```diff
@@ -255,6 +265,95 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *c
...
+
+	egreeNetworkPolicyCRD := &unstructured.Unstructured{}
+	egreeNetworkPolicyCRD.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apiextensions.k8s.io",
+		Kind:    "CustomResourceDefinition",
+		Version: "v1",
+	})
+	ctx := context.TODO()
+	defer ctx.Done()
+	if err = unrestrictedClient.Get(context.TODO(), client.ObjectKey{Name: "egressnetworkpolicies.network.openshift.io", Namespace: ""}, egreeNetworkPolicyCRD); err != nil {
```
leelavg, line 289:
```suggestion
	var options client.Options
	var err error

	options.Scheme = runtime.NewScheme()
	utilruntime.Must(apiextenstionsv1.AddToScheme(options.Scheme))

	egreeNetworkPolicyCRD := &unstructured.Unstructured{}
	egreeNetworkPolicyCRD.SetGroupVersionKind(schema.GroupVersionKind{
		Group:   "apiextensions.k8s.io",
		Kind:    "CustomResourceDefinition",
		Version: "v1",
	})
	ctx := context.TODO()
	defer ctx.Done()
	if err = r.UnrestrictedClient.Get(context.TODO(), client.ObjectKey{Name: "egressnetworkpolicies.network.openshift.io", Namespace: ""}, egreeNetworkPolicyCRD); err != nil {
```
- did you try above, type already has the field to use unrestricted client?

controllers/managedocs_controller.go
```diff
@@ -255,6 +265,95 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *c
...
+		if strings.Contains(err.Error(), "could not find the requested resource") {
```
leelavg, line 290: - it's not a good idea to have this code duplicated
- `NewControllerManagedBy` returns a builder and you can keep on calling functions off of it, why not use that pattern?
- Add the common resources and based on the error decide b/n `EgressNetworkPolicy` and `EgressFirewall`

controllers/managedocs_controller.go
```diff
@@ -1520,6 +1628,67 @@ func (r *ManagedOCSReconciler) reconcileEgressNetworkPolicy() error {
 	return nil
 }
 
+func (r *ManagedOCSReconciler) reconcileEgressFirewall() error {
+	if r.DeploymentType != convergedDeploymentType {
```
leelavg, line 1632: - a bit confused here, so this reconcile phase only happens for converged mode?
- @renan-campos can you help here? why are we skipping the egresspolicy for provider & consumer mode in https://github.com/red-hat-storage/ocs-osd-deployer/blob/0f906965804ef8d4bfa5f2de463d28bb95e6cf30/controllers/managedocs_controller.go#L1464-L1468 was this due for VPC patch?

controllers/suite_test.go
```diff
@@ -146,7 +146,10 @@ var _ = BeforeSuite(func() {
 		err = opv1a1.AddToScheme(scheme.Scheme)
 		Expect(err).NotTo(HaveOccurred())
 
-		err = openshiftv1.AddToScheme(scheme.Scheme)
+		//err = openshiftv1.AddToScheme(scheme.Scheme)
```
leelavg, line 149: - pls remove if not required

leelavg (commented):
- just released after going through the patch, at the moment we aren't using EgressNetworkPolicy for `provider` and `consumer` mode deployments.

controllers/managedocs_controller.go
renan-campos: @leelavg in provider/consumer mode, the ceph pods are running in the host network namespace, so the ingress/egress network policies need to allow traffic between pods on the node network. However we can't set this without knowing what the CIDR range is for this network is first. By default it is 10.0.0.0/16, but it could be whatever it was initially set as. As a quick fix, the policies were disabled. But we do eventually need it to satisfy security requirements. This [PR](https://github.com/red-hat-storage/ocs-osd-deployer/pull/180) handles this, we just need it merged :pray:

controllers/managedocs_controller.go
rewantsoni: Yes, it works. I have updated the PR

controllers/managedocs_controller.go
```diff
@@ -311,10 +333,24 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *c
 			&source.Kind{Type: &opv1a1.ClusterServiceVersion{}},
 			enqueueManangedOCSRequest,
 			csvPredicates,
-		).
+		)
+
+	err = r.UnrestrictedClient.Get(ctx, client.ObjectKey{Name: "egressnetworkpolicies.network.openshift.io", Namespace: ""}, egreeNetworkPolicyCRD)
```
nb-ohad, line 338: I don't think that we should couple the concern of setting up the reconciliation with figuring out which type of ingress policy we need to use. Please move the check outside and pass that information here via an argument.

controllers/managedocs_controller.go
```diff
@@ -255,7 +263,22 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *c
...
+
+	options.Scheme = runtime.NewScheme()
+	utilruntime.Must(apiextenstionsv1.AddToScheme(options.Scheme))
+
+	egreeNetworkPolicyCRD := &unstructured.Unstructured{}
+	egreeNetworkPolicyCRD.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apiextensions.k8s.io",
+		Kind:    "CustomResourceDefinition",
+		Version: "v1",
+	})
```
nb-ohad, line 277: There should be a very specific reason to use `Unstructured` instead of structured structs. Can you please explain why you choose to use it?

controllers/managedocs_controller.go
```diff
@@ -571,6 +611,11 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			return ctrl.Result{}, err
 		}
 		if err := r.reconcileEgressNetworkPolicy(); err != nil {
+			if errors.IsNotFound(err) {
+				if err := r.reconcileEgressFirewall(); err != nil {
+					return ctrl.Result{}, err
+				}
+			}
```
nb-ohad, line 618: We do not need to fail logic as we already know which CR to use based on the check in `SetupWithManager`

nb-ohad (changes requested):

controllers/managedocs_controller.go
rewantsoni: Using a structured object and a restricted client gives me an error "the cache is not started, can not read objects", and if I use a structured object with an unrestricted client it returns the resource as NotFound.

controllers/managedocs_controller.go
```diff
@@ -87,7 +88,7 @@ const (
 	deployerCSVPrefix                  = "ocs-osd-deployer"
 	ocsOperatorName                    = "ocs-operator"
 	mcgOperatorName                    = "mcg-operator"
-	egressNetworkPolicyName            = "egress-rule"
+	egressPolicyName                   = "default"
```
leelavg, line 91: - nit, some formatting error here? an extra tab 🤔

controllers/managedocs_controller.go
```diff
@@ -104,6 +105,8 @@ const (
 	providerDeploymentType             = "provider"
 	rhobsRemoteWriteConfigIdSecretKey  = "prom-remote-write-config-id"
 	rhobsRemoteWriteConfigSecretName   = "prom-remote-write-config-secret"
+	OVNNetworkType                     = "ovn"
+	SDNNetworkType                     = "sdn"
```
leelavg, line 109: - might be a bit verbose and at the same time I think it's better to have new type only for network 🤔
- cause we have only two possible values and better force them 

``` go
type NetworkType string
// ...
const (
  NetworkOVN NetworkType = "ovn"
  NetworkSDN NetworkType = "sdn"
)
// ...
type ManagedOCSReconciler struct {
// ...
  Network NetworkType
// ...
}
```

main.go
```diff
@@ -142,7 +161,21 @@ func main() {
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			reconciler.NetworkType = controllers.OVNNetworkType
+		} else {
+			setupLog.Error(err, "Error Getting CRD for EgressNetworkPolicy")
+			os.Exit(1)
+		}
+	}
+
+	if err = (reconciler).SetupWithManager(mgr, nil); err != nil {
```
leelavg, line 178: - nit, this method is actually defined on pointer receiver do we need parens around it?
```suggestion
	if err = reconciler.SetupWithManager(mgr, nil); err != nil {
```

leelavg (commented):
- good work I'd say.

controllers/managedocs_controller.go
rewantsoni: In the case of 4.10 cluster with SDN, it will return that the resource is notFound and hence the Network type will be changed to OVN and it will try to deploy egressfirewall instead of egressnetworkpolicy

controllers/managedocs_controller.go
rewantsoni: No, we have to change the PolicyName form egress-rule to default as EgressFirewall doesn't accept any other name. they have a check - ```Invalid value: "egress-rule": metadata.name in body should match '^default$'" for field "metadata.name```

nb-ohad: @rewantsoni that will create a backward compatibility issue for existing deployments, as it will leave non-reconciled existing  egressNetworkPolicy resources on upgraded clusters.

Please use the original name for `egressNetworkPolicy` resource and the name default for `EgressFirewall` resources

main.go
nb-ohad: Let's not mess up with generated code. There is a reason that it was generated that way

controllers/managedocs_controller.go
```diff
@@ -311,10 +317,18 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *c
...
-		).
+		)
+
+	r.Log.Info(fmt.Sprintf("Creating ManagedOCS %s controller", r.NetworkType))
+
+	if r.NetworkType == OVNNetworkType {
+		controller = controller.Owns(&ovnv1.EgressFirewall{})
+	} else {
+		controller = controller.Owns(&openshiftv1.EgressNetworkPolicy{})
+	}
```
nb-ohad, line 328: If we are to use an enum, please use a switch statement that covers all option and a default option if the value is not supported that return an error (or even Panic)

controllers/managedocs_controller.go
```diff
@@ -570,8 +588,14 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
...
-			return ctrl.Result{}, err
+		if r.NetworkType == OVNNetworkType {
+			if err := r.reconcileEgressFirewall(); err != nil {
+				return ctrl.Result{}, err
+			}
+		} else {
+			if err := r.reconcileEgressNetworkPolicy(); err != nil {
+				return ctrl.Result{}, err
+			}
 		}
```
nb-ohad, line 599: Please move the conditional into the reconcile function as an early exist conditional and then just call both functions. For example, at the start of `reconcileEgressFirewall` you will have
```
if r.NetworkType != OVNNetworkType {
    return nil
}
```
Same for the other reconcile func.

Please note we are using a similar pattern for deployment types.

main.go
```diff
@@ -142,7 +161,21 @@ func main() {
 		RHOBSSecretName:              fmt.Sprintf("%v-prom-remote-write", addonName),
 		RHOBSEndpoint:                envVars[rhobsEndpointEnvVarName],
 		RHSSOTokenEndpoint:           envVars[rhssoTokenEndpointEnvVarName],
-	}).SetupWithManager(mgr, nil); err != nil {
+		NetworkType:                  controllers.SDNNetworkType,
```
nb-ohad, line 164: NetworkType is a misleading name, please change to something like CNIType or NetworkInterfaceType.

main.go
```diff
@@ -142,7 +161,21 @@ func main() {
...
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			reconciler.NetworkType = controllers.OVNNetworkType
+		} else {
+			setupLog.Error(err, "Error Getting CRD for EgressNetworkPolicy")
+			os.Exit(1)
+		}
+	}
+
```
nb-ohad, line 177: Why are we introducing this logic between the creation of the controller and `SetupWithManager`?
Please move this logic before the creation of the controller.

Also, it is not recommended to use the reconciler/mgr client for this operation as it will introduce the CRD type into the client cache, and we already know that the client implementation has an issue with cache size.

I would suggest having a util function (in this file) of the form `getNetworkInterfaceType() NetworkInterfaceType` which initializes  a non-cached client which is based on a very minimal schema (you only need the CRD schema) and use it to populate the new the controller field.

controllers/managedocs_controller.go
nb-ohad: The name NetworkType for the enum is misleading, we use the same network in both cases, it is just the CNI implementation that is different. I would suggest to use the name `NetworkInterfaceType` instead

In that case the values will be of the form `NetwortIntefaceOVN` and `NetworkInterfaceSDN`

controllers/suite_test.go
```diff
@@ -92,6 +95,11 @@ func TestAPIs(t *testing.T) {
 			testDeploymentTypeEnvVarName, convergedDeploymentType, providerDeploymentType, consumerDeploymentType))
 	}
 
+	networkType = os.Getenv(testNetworkTypeEnvVarName)
+	if networkType != OVNNetworkType && networkType != SDNNetworkType {
+		panic(fmt.Sprintf("Enviroment var '%s' should be set to one of '%s' or '%s' values", testNetworkTypeEnvVarName, OVNNetworkType, SDNNetworkType))
+	}
+
```
nb-ohad, line 102: We should be able to test even with non-supported values to validate behavior in these cases as well.
Please remove this check.

controllers/suite_test.go
```diff
@@ -187,6 +198,7 @@ var _ = BeforeSuite(func() {
 			RHOBSEndpoint:                os.Getenv(testRHOBSEndpointEnvVarName),
 			RHSSOTokenEndpoint:           os.Getenv(testRHssoTokenEndpointEnvVarName),
 			RHOBSSecretName:              testRHOBSSecretName,
+			NetworkType:                  networkType,
```
nb-ohad, line 201:
```suggestion
			NetworkType:                  os.Getenv(testNetworkTypeEnvVarName),
```

Makefile
```diff
@@ -121,19 +121,31 @@ vet: ## Run go vet against code.
...
+test-provider-ovn: manifests generate fmt vet envtest ## Run provider mode tests with network type ovn.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" DEPLOYMENT_TYPE=provider NETWORK_TYPE=ovn go test ./... -coverprofile cover.out
+
+.PHONY: test-consumer-ovn
+test-consumer-ovn: manifests generate fmt vet envtest ## Run consumer mode tests with network type ovn.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" DEPLOYMENT_TYPE=consumer NETWORK_TYPE=ovn go test ./... -coverprofile cover.out
 
 .PHONY: test
-test: test-converged test-provider test-consumer ## Run tests.
+test: test-converged test-provider test-consumer test-converged-ovn test-provider-ovn test-consumer-ovn ## Run tests.
```
nb-ohad, line 148: It is meaningless to run the entire test suite with both OVN and SDN variants and is a waste of time.
The OVN vs SDN decision is only affecting a very small portion of our codebase.

nb-ohad (changes requested):

main.go
```diff
@@ -142,7 +161,21 @@ func main() {
 		RHOBSSecretName:              fmt.Sprintf("%v-prom-remote-write", addonName),
 		RHOBSEndpoint:                envVars[rhobsEndpointEnvVarName],
 		RHSSOTokenEndpoint:           envVars[rhssoTokenEndpointEnvVarName],
-	}).SetupWithManager(mgr, nil); err != nil {
+		NetworkType:                  controllers.SDNNetworkType,
+	}
+
+	err = reconciler.Client.Get(ctx, client.ObjectKeyFromObject(egressNetworkPolicyCRD), egressNetworkPolicyCRD)
```
nb-ohad, line 167: 1. Please use the following instead of using `unstructured.Unstructured{}` 
2. Please do not mix error handling with program logic
3. have selection logic in one place instead of trying to override a "default" value - it makes following the code simpler and the reader do not need to make an assumption about the value in the non-overriding case.

```
	crd := apiextensionsv1.CustomResourceDefinition{}
	err := r.Client.Get(ctx, types.NamespacedName{Name: "egressnetworkpolicies.network.openshift.io"}, &crd)
	if  err == nil && !errors.IsNotFound(err){
		setupLog.Error(err, "Error Getting CRD for EgressNetworkPolicy")
		os.Exit(1)
        }
        netInterfaceType := ""
        if crd.UID != "" {
                netInterfaceType := NetworkInterfaceSDN
        } else {
                netInterfaceType := NetworkInterfaceOVN
        }
```
(code adjusted to the naming requests from prev comments)

nb-ohad (changes requested):

controllers/managedocs_controller.go
rewantsoni: done

controllers/managedocs_controller.go
rewantsoni: done

controllers/managedocs_controller.go
rewantsoni: done

controllers/managedocs_controller.go
rewantsoni: done

controllers/managedocs_controller.go
rewantsoni: done

controllers/managedocs_controller.go
rewantsoni: Done

controllers/managedocs_controller.go
leelavg: ack

controllers/managedocs_controller.go
leelavg: ack

main.go
rewantsoni: Done

controllers/suite_test.go
leelavg: ack

controllers/managedocs_controller.go
leelavg: ack

controllers/managedocs_controller.go
rewantsoni: Done

main.go
rewantsoni: Done

controllers/managedocs_controller.go
leelavg: ack

main.go
rewantsoni: Done

main.go
leelavg: ack

controllers/managedocs_controller.go
```diff
@@ -171,12 +179,15 @@ type ManagedOCSReconciler struct {
...
 // +kubebuilder:rbac:groups="network.openshift.io",namespace=system,resources=egressnetworkpolicies,verbs=create;get;list;watch;update
+// +kubebuilder:rbac:groups="k8s.ovn.org",namespace=system,resources=egressfirewalls,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="coordination.k8s.io",namespace=system,resources=leases,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="odf.openshift.io",namespace=system,resources=storagesystems,verbs=list;watch;delete
 // +kubebuilder:rbac:groups="config.openshift.io",resources=clusterversions,verbs=get;watch;list
+// +kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;watch;list
 
 // SetupWithManager creates an setup a ManagedOCSReconciler to work with the provided manager
 func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *controller.Options) error {
+	fmt.Println(r.NetworkInterfaceType)
```
leelavg, line 190: - neither a good place nor a correct format for logging the info here.

leelavg (commented):
- minor comment, rest lgtm

controllers/managedocs_controller.go
rewantsoni: done

leelavg (dismissed):

controllers/managedocs_controller.go
```diff
@@ -106,6 +110,10 @@ const (
 	rhobsRemoteWriteConfigSecretName   = "prom-remote-write-config-secret"
 )
 
+func GetCRDNames() []string {
+	return []string{EgressFirewallCRD, EgressNetworkPolicyCRD}
+}
```
leelavg, line 115: - better move this supporting func to end of file

controllers/managedocs_controller.go
```diff
@@ -171,9 +181,11 @@ type ManagedOCSReconciler struct {
 // +kubebuilder:rbac:groups="storage.k8s.io",resources=storageclass,verbs=get;list;watch
 // +kubebuilder:rbac:groups="networking.k8s.io",namespace=system,resources=networkpolicies,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="network.openshift.io",namespace=system,resources=egressnetworkpolicies,verbs=create;get;list;watch;update
+// +kubebuilder:rbac:groups="k8s.ovn.org",namespace=system,resources=egressfirewalls,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="coordination.k8s.io",namespace=system,resources=leases,verbs=create;get;list;watch;update
 // +kubebuilder:rbac:groups="odf.openshift.io",namespace=system,resources=storagesystems,verbs=list;watch;delete
 // +kubebuilder:rbac:groups="config.openshift.io",resources=clusterversions,verbs=get;watch;list
+// +kubebuilder:rbac:groups="apiextensions.k8s.io",resources=customresourcedefinitions,verbs=get;watch;list
```
leelavg, line 188: - this doesn't need to be part of sdk cache as it's being used by a new client in `main.go`, isn't that the case?

controllers/suite_test.go
```diff
@@ -172,6 +175,11 @@ var _ = BeforeSuite(func() {
 		})
 		Expect(err).ToNot(HaveOccurred())
 
+		crdList := map[string]bool{}
+		for _, CRDNames := range GetCRDNames() {
+			crdList[CRDNames] = true
+		}
```
leelavg, line 181: - I'm tending towards to believe the motive for moving from earlier design of using an enum is to enable tests?
- Or there any other reason?

main.go
```diff
@@ -215,3 +221,30 @@ func ensureManagedOCS(c client.Client, log logr.Logger, envVars map[string]strin
...
+	for _, crdName := range crdNames {
+		crd := apiextensionsv1.CustomResourceDefinition{}
+		err = k8sClient.Get(context.Background(), types.NamespacedName{Name: crdName}, &crd)
+
+		if err != nil && !errors.IsNotFound(err) {
+			setupLog.Error(err, fmt.Sprintf("Error Getting CRD for %s", crdName))
+			os.Exit(1)
+		}
+
+		CRDExist[crdName] = crd.UID != ""
```
leelavg, line 247: - is there any scenario where both the CRDs exist?
- I remembering hearing that a previous running cluster with SDN will continue to use SDN even though OVN exists
- however, this needs to be confirmed.

main.go
```diff
@@ -215,3 +221,30 @@ func ensureManagedOCS(c client.Client, log logr.Logger, envVars map[string]strin
...
+	options.Scheme = runtime.NewScheme()
+	utilruntime.Must(apiextensionsv1.AddToScheme(options.Scheme))
+
+	k8sClient, err := client.New(config.GetConfigOrDie(), options)
+	if err != nil {
+		setupLog.Error(err, "error creating client")
+		os.Exit(1)
+	}
+
+	CRDExist := map[string]bool{}
```
leelavg, line 237: - a teeny tiny optimisation here
```suggestion
	CRDExist := map[string]struct{}
```
- and change
``` go
// set op
// from
CRDExist[crdName] = crd.UID != ""

// to
if crd.UID != "" {
  CRDExist[crdName] = struct{}{}
}

// get op
// from
if CRDExist[crdName] {}

// to
if _, exist := CRDExist[crdName]; exist {}
```
- allocation of `struct{}` costs you literally nothing compared to any type
- if you don't feel readability is going to take a hit above change can be made

controllers/managedocs_controller.go
rewantsoni: If we don't have this, the serviceAccount will not have permission to access the CRD resource

controllers/suite_test.go
rewantsoni: Yes, it will also enable us to deploy the resources based on whether or not CRD exists.

main.go
nb-ohad: I think this is a no-gain. 
The common practice is to have a map of bools to implement a set. To enable removing from the set without the need to actually change anything in memory

main.go
```diff
@@ -142,6 +147,7 @@ func main() {
 		RHOBSSecretName:              fmt.Sprintf("%v-prom-remote-write", addonName),
 		RHOBSEndpoint:                envVars[rhobsEndpointEnvVarName],
 		RHSSOTokenEndpoint:           envVars[rhssoTokenEndpointEnvVarName],
+		CRDExist:                     checkIfCRDExist(controllers.GetCRDNames()),
```
nb-ohad, line 150: `controller.GetCRDNames` does not make sense here. 
As a reader I would like to be able to understand which CRDs are we looking for without the need to jump into another file.

main.go
```diff
@@ -215,3 +221,30 @@ func ensureManagedOCS(c client.Client, log logr.Logger, envVars map[string]strin
 		return err
 	}
 }
+
+func checkIfCRDExist(crdNames []string) map[string]bool {
```
nb-ohad, line 225: A function prefixed `check` indicated a boolean result which is not the case.
I would suggest renaming the function to `mapCRDAvailability` which makes sense as the function return a map with a boolean indicating if each of the requested CRD is available or not

nb-ohad (changes requested):

main.go
nb-ohad: No there isn't! but this is a util function taking any number of CRDs and can be useful for other feature that need CRD detection.

controllers/managedocs_controller.go
```diff
@@ -63,7 +64,9 @@ import (
 )
 
 const (
-	ManagedOCSFinalizer = "managedocs.ocs.openshift.io"
+	ManagedOCSFinalizer    = "managedocs.ocs.openshift.io"
+	EgressFirewallCRD      = "egressfirewalls.k8s.ovn.org"
+	EgressNetworkPolicyCRD = "egressnetworkpolicies.network.openshift.io"
 )
```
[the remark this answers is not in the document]
rewantsoni: Moved it under the same const section

controllers/managedocs_controller.go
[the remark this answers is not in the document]
rewantsoni: Done

controllers/managedocs_controller.go
```diff
@@ -126,11 +134,13 @@ type ManagedOCSReconciler struct {
 	RHOBSSecretName              string
 	RHOBSEndpoint                string
 	RHSSOTokenEndpoint           string
+	CRDExist                     map[string]bool
```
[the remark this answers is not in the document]
rewantsoni: Done

PR #250, opened 2022-09-06
Ceph OSDs with Taints and TSC

leelavg:


leelavg:
- Tested locally, however marking it as draft as it needs to be coordinated outside of deployer functionalities.
- So, early reviews will be helpful, thanks.

utils/resources.go
```diff
@@ -57,11 +57,11 @@ var resourceRequirements = map[string]corev1.ResourceRequirements{
 	"sds": {
 		Limits: corev1.ResourceList{
 			"cpu":    resource.MustParse("2000m"),
-			"memory": resource.MustParse("7Gi"),
+			"memory": resource.MustParse("6Gi"),
```
ezio-auditore, line 60: Any reason to decrease the memory limits?

leelavg: - ptal at https://issues.redhat.com/browse/ODFMS-152, it was discussed in great length and the repercussions are well aware
- however, we don't have a final say
- gist is, we want to accommodate 2 OSDs in a single m5.xlarge and 2 OSDs with 7Gi mem won't fit the bill

leelavg:
- depends on #262

leelavg:
- rebased

templates/providerstoragecluster.go
```diff
@@ -31,8 +31,59 @@ import (
 
 const (
 	ProviderOSDSizeInTiB = 4
+	osdNodeLabel         = "node.ocs.openshift.io/osd"
```
leelavg, line 34: - due to OCM limitations, once set changing this field isn't possible
- for reviewers, pls provide an ack whether this is enough or some more info is required in the key, i.e.,
```
osdNodeLabelKey = "node.ocs.openshift.io/reserved-for"
osdNodeLabelValue = "ceph-osd"
```
- and the same will be changed in this file

leelavg:
- Should be merged after #267

templates/providerstoragecluster.go
leelavg: - discussed offline, we are good with having no value for the label

leelavg:
- rebased

nb-ohad (approved):

PR #259, opened 2022-09-30
use correct label for spreading alertmanager

leelavg:
- make sure not all three replicas run on single node


leelavg:
- the existing label combo `app: alertmanager` doesn't exist on the alertmanager pods
- as a result, no extra effort is being put by the kube-scheduler to spread the pods even though it's possible
- use `app.kubernetes.io/name` for label key as it's guaranteed to exist on alertmanager resource
- due to resource constraints, it may so happen that after the spread other workloads in ODF Control Plane will go into pending and so a `maxSkew: 2` is used to guarantee atleast one alertmanager will be not be grouped with other alertmanager pods with `replica: 3`

PR #262, opened 2022-10-27
Run only one alertmanager pod

leelavg:
- Upto now we were using 3 replicas for alertmanager
- Due to wrong usage of labels all of these replicas are ending up on only single node and it doesn't serve redundancy, moreover eating away extra resources
- There'll not be much repercussions due to alertmanager being StatefulSet as it guarantees new pod is Ready before killing old pod (Evict API friendly)


bindrad (approved):

leelavg:
- rebased

nb-ohad (approved):

PR #264, opened 2022-11-03
Fix Egress Firewall by allowing access to OCP join network range

rewantsoni:
- Configure the Egress Firewall to allow access to OCP join network range
- Make sure that the allow rules appear before Deny rules

Ref: [Configure EgressFirewall on OCP 4.11](https://docs.openshift.com/container-platform/4.11/networking/ovn_kubernetes_network_provider/configuring-egress-firewall-ovn.html#nw-egressnetworkpolicy-about_configuring-egress-firewall-ovn)


templates/egressfirewall.go
```diff
@@ -29,6 +29,12 @@ var EgressFirewallTemplate = ovnv1.EgressFirewall{
 				},
 				Type: ovnv1.EgressFirewallRuleAllow,
 			},
+			{
```
leelavg, line 32: - pls do add egress rules for RHOBS as well if not done already
- @ezio-auditore other than SSO & RHOBS do we need any other egress rules wrt alerting?

templates/egressfirewall.go
```diff
@@ -29,6 +29,12 @@ var EgressFirewallTemplate = ovnv1.EgressFirewall{
...
+				To: ovnv1.EgressFirewallDestination{
+					CIDRSelector: "100.64.0.0/16",
```
leelavg, line 34: Any issues if this is added to SDN NetworkPolicy? If there's none, better add it there as well.

rewantsoni: No, there is not any issue if we add it to SDN as well, I will add it there.

templates/egressfirewall.go
rewantsoni: Done

cmd/awsDataGather/main_test.go
```diff
@@ -10,12 +10,9 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
 var _ = Describe("AWS Data Gathering behavior", func() {
-	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
```
bindrad, line 18: Why is this removed?

rewantsoni: It is extra code, we are not logging anything here

bindrad: Can we do this change in another PR?

bindrad (approved):

controllers/managedocs_controller.go
```diff
@@ -1531,10 +1531,6 @@ func (r *ManagedOCSReconciler) reconcileEgressNetworkPolicy() error {
 				return fmt.Errorf("Unable to get DeadMan's Snitch secret: %v", err)
 			}
 		}
-		dmsURL := string(r.deadMansSnitchSecret.Data["SNITCH_URL"])
```
leelavg, line 1534: - just confirming, this is removed as we are checking `snitchURL.Hostname()` afterwards?

leelavg (dismissed):

controllers/managedocs_controller.go
rewantsoni: Yes

ezio-auditore:
LGTM

controllers/managedocs_controller.go
```diff
@@ -1531,10 +1531,6 @@ func (r *ManagedOCSReconciler) reconcileEgressNetworkPolicy() error {
...
-		if dmsURL == "" {
-			return fmt.Errorf("DeadMan's Snitch secret does not contain a SNITCH_URL entry")
-		}
 		snitchURL, err := url.Parse(string(r.deadMansSnitchSecret.Data["SNITCH_URL"]))
 		if err != nil {
 			return fmt.Errorf("Unable to parse DMS url: %v", err)
```
nb-ohad, line 1536: Why are we removing the specific check for an empty entry? I would prefer to have a more targeted error than a generic one. If the desire was to not parse twice then the solution should be:
```
snitchStr := string(r.deadMansSnitchSecret.Data["SNITCH_URL"])
if snitchStr == "" {
	return fmt.Errorf("DeadMan's Snitch secret does not contain a SNITCH_URL entry")
}
snitchURL, err := url.Parse(snitchStr)
if err != nil {
}
```

controllers/managedocs_controller.go
```diff
@@ -1628,14 +1650,13 @@ func (r *ManagedOCSReconciler) reconcileEgressFirewall() error {
...
 		}
+		if snitchURL.Hostname() == "" {
+			return fmt.Errorf("DMS snitch url %q does not have a valid hostname", snitchURL)
+		}
```
nb-ohad, line 1659: Please align with my prev comment

nb-ohad (changes requested):

nb-ohad (approved):

PR #267, opened 2022-11-16
Change resource req and introduce OSD topo changes

leelavg:
- Reduce CPU ask for Prometheus. Use same requests & limits for MDS
- Introdcue topology spread constraints for Ceph OSD placements
- Make Ceph OSDs placement tolerate nodes tainted for OSDs

### DIFF
--- a/templates/providerstoragecluster.go
+++ b/templates/providerstoragecluster.go
@@ -31,7 +31,45 @@ import (
 
 const (
 	ProviderOSDSizeInTiB = 4
+	osdNodeLabel         = "node.ocs.openshift.io/osd"
 )
+
+var osdToleration corev1.Toleration = corev1.Toleration{
+	Key:      osdNodeLabel,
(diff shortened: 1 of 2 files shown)


templates/providerstoragecluster.go
```diff
@@ -31,8 +31,46 @@ import (
 
 const (
 	ProviderOSDSizeInTiB = 4
+	osdNodeLabel         = "node.ocs.openshift.io/osd"
```
leelavg, line 34: - due to OCM limitations, once set changing this field isn't possible
- for reviewers, pls provide an ack whether this is enough or some more info is required in the key, i.e.,
```
osdNodeLabelKey = "node.ocs.openshift.io/reserved-for"
osdNodeLabelValue = "ceph-osd"
```
- and the same will be changed in this file

leelavg:
closing in favour of #250

PR #279, opened 2023-02-06
Upgrade to ODF 4.11.5 and Enabling encryption with KMS feature

ezio-auditore:
Jira Tasks

- [ODFMS-280](https://issues.redhat.com/browse/ODFMS-280)

- [ODFMS-282](https://issues.redhat.com/browse/ODFMS-282)
---
Signed-off-by: Kaustav Majumder <kmajumde@redhat.com>


controllers/managedocs_controller.go
```diff
@@ -920,6 +930,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+			SecretName        string `json:"secretName"`
+		}{
+			EncryptionKMSType: "aws-sts-metadata",
+			SecretName:        "tenant-aws-secret", // user needs to create a secret with this name
+		}
+
+		data, _ := json.Marshal(encryptionMetadata)
+
+		r.csiKMSConnectionDetailsConfigMap.Data = map[string]string{
+			"encryption-with-aws-kms": string(data), // encryption-with-aws-kms will be used by customer to create storageClassClaim
```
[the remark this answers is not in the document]
nb-ohad: What is a backpool storageclassclaim? there is no such a thing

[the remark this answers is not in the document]
bindrad: This feature will enable the encryption only for block storage so I thought mentioning it will be good.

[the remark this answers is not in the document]
bindrad: To create the storageClass that has encryption enabled, we need to create a claim in which we need to mention a type. If that type is blockpool only then encryption will work.

controllers/managedocs_controller.go
```diff
@@ -541,6 +551,9 @@ func (r *ManagedOCSReconciler) reconcilePhases() (reconcile.Result, error) {
 			r.addonParams[storageUnitKey] = "Ti"
 		}
 
+		if err := r.reconcileDesiredODFSubscription(); err != nil {
```
bindrad, line 554:
```suggestion
		if err := r.reconcileODFSubscription(); err != nil {
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
 		deviceSet.Count = currDeviceSetCount
 	}
 }
+
+func (r *ManagedOCSReconciler) reconcileDesiredODFSubscription() error {
```
bindrad, line 2100:
```suggestion
func (r *ManagedOCSReconciler) reconcileODFSubscription() error {
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+	subList := &opv1a1.SubscriptionList{}
+	if err := r.list(subList); err != nil {
+		return fmt.Errorf("Unable to fetch current subscription list %v", err)
```
bindrad, line 2103:
```suggestion
		return fmt.Errorf("Unable to list subscriptions: %v", err)
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		item := subList.Items[index]
+		if item.Spec.Package == odfPackageName {
+			actualSub = &item
+		}
+	}
+	if actualSub.Spec.Channel == desiredODFSubscriptionChannel {
+		r.Log.Info("Current ODF subscription channel is same as desired ODF subscription channel, not updating")
+		return nil
+	}
+	r.Log.Info("Updating ODF subsctiption channel", "current channel", actualSub.Spec.Package, "desired channel", desiredODFSubscriptionChannel)
```
bindrad, line 2116:
```suggestion
	r.Log.Info("Updating ODF subsctiption channel", "current channel", actualSub.Spec.Channel, "desired channel", desiredODFSubscriptionChannel)
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+	actualSub.Spec.Channel = desiredODFSubscriptionChannel
+
+	if err := r.update(actualSub); err != nil {
+		return fmt.Errorf("Unable to update channel %v", err)
```
bindrad, line 2120:
```suggestion
		return fmt.Errorf("Unable to update %s subscription: %v", odfPackageName, err)
```

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+		data, _ := json.Marshal(encryptionMetadata)
+
+		r.csiKMSConnectionDetailsConfigMap.Data = map[string]string{
+			"encryption-with-aws-kms": string(data), // encryption-with-aws-kms will be used by customer to create storageClassClaim
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Unable to create csi-kms-connection-details configMap: %v", err)
```
bindrad, line 968:
```suggestion
		return fmt.Errorf("Unable to create %s configMap: %v", csiKMSConnectionDetailsConfigMapName, err)
```

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
 	return nil
 }
 
+func (r *ManagedOCSReconciler) reconcileCSIKMSConnectionDetailsConfigMap() error {
+	if r.DeploymentType != consumerDeploymentType {
+		r.Log.Info("Non consumer deployment, skipping reconcile for csi-kms-connection-details config map")
+		return nil
+	}
+	r.Log.Info("Reconciling csi-kms-connection-details configMap")
```
bindrad, line 944:
```suggestion
	r.Log.Info(fmt.Sprintf("Reconciling %s configMap", csiKMSConnectionDetailsConfigMapName))
```

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
 	return nil
 }
 
+func (r *ManagedOCSReconciler) reconcileCSIKMSConnectionDetailsConfigMap() error {
+	if r.DeploymentType != consumerDeploymentType {
+		r.Log.Info("Non consumer deployment, skipping reconcile for csi-kms-connection-details config map")
```
bindrad, line 941:
```suggestion
		r.Log.Info(fmt.Sprintf("Non consumer deployment, skipping reconcile for %s config map", csiKMSConnectionDetailsConfigMapName))
```

bindrad (changes requested):

bindrad:
Does the channel in ocs-operator subscription and other dependencies of odf-operator get updated automatically?

config/rbac/role.yaml
```diff
@@ -40,6 +40,15 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
leelavg, line 43: - May I know why we need a handle on subscription resource?
- I assume it's due to minor version change, however just want to understand the background in detail if possible.

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: updated

ezio-auditore:
> Does the channel in ocs-operator subscription and other dependencies of odf-operator get updated automatically after updating channel in odf-operator subscription?

Yes, they do automatically get upgraded to 4.11

config/rbac/role.yaml
ezio-auditore: For upgrading minor versions we need to provide a dependency and code changes in the deployer.
So this change `"4.10.9 || 4.11.5"` updates the deployer with the code changes necessary. odf doesn't get upgraded yet.
The deployer acts on the subscription to change the ODF channel to 4.11after which odf upgrades

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+func (r *ManagedOCSReconciler) reconcileODFSubscription() error {
+	subList := &opv1a1.SubscriptionList{}
+	if err := r.list(subList); err != nil {
+		return fmt.Errorf("Unable to list subscriptions %v", err)
+	}
+	actualSub := &opv1a1.Subscription{}
+	for index := range subList.Items {
+		item := subList.Items[index]
+		if item.Spec.Package == odfPackageName {
+			actualSub = &item
```
leelavg, line 2109: - rest of the function is assuming that you actually found the subscription 🤔 
- in failure case, `r.update` fails incorrectly

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+		if err := r.own(r.csiKMSConnectionDetailsConfigMap); err != nil {
+			return err
+		}
+
+		encryptionMetadata := struct {
+			EncryptionKMSType string `json:"encryptionKMSType"`
+			SecretName        string `json:"secretName"`
+		}{
+			EncryptionKMSType: "aws-sts-metadata",
+			SecretName:        "tenant-aws-secret", // user needs to create a secret with this name
```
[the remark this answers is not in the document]
ezio-auditore: The feature is in Dev preview. So no docs needed

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
 		deviceSet.Count = currDeviceSetCount
 	}
 }
+
+func (r *ManagedOCSReconciler) reconcileODFSubscription() error {
+	subList := &opv1a1.SubscriptionList{}
+	if err := r.list(subList); err != nil {
+		return fmt.Errorf("Unable to list subscriptions %v", err)
+	}
+	actualSub := &opv1a1.Subscription{}
```
nb-ohad, line 2105: Why "actual"? what is "actual" about it?
```suggestion
	odfSub := &opv1a1.Subscription{}
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+	for index := range subList.Items {
+		item := subList.Items[index]
```
nb-ohad, line 2107: This will create a copy of the item in the list, please use a reference here
```suggestion
		item := &subList.Items[index]
```

controllers/managedocs_controller.go
nb-ohad, line 2109: Here you are returning a reference to a copy. I don't think this is what you wanted (see my comment about item).
Also, you are continuing with the loop even after you found the desired subscription
```suggestion
			odfSum = item
			break;
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		if item.Spec.Package == odfPackageName {
+			actualSub = &item
+		}
+	}
+	if actualSub.Spec.Channel == desiredODFSubscriptionChannel {
```
nb-ohad, line 2112: Don't use the subscription before validating that you found one

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		r.Log.Info("Current ODF subscription channel is same as desired ODF subscription channel, not updating")
```
nb-ohad, line 2113: This log message will be visible for every reconcile iteration. Assuming that a minor version upgrade is very rare, do we really need it? 

```suggestion
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		return nil
+	}
+	r.Log.Info("Updating ODF subsctiption channel", "current channel", actualSub.Spec.Channel, "desired channel", desiredODFSubscriptionChannel)
```
nb-ohad, line 2116: Let's give a more meaningful description of the operation. 
```suggestion
	r.Log.Info("Initiating ODF minor version upgrade, current release", odfSub.Spec.Channel, "target release",  desiredODFSubscriptionChannel)
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,28 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+	actualSub.Spec.Channel = desiredODFSubscriptionChannel
+
+	if err := r.update(actualSub); err != nil {
+		return fmt.Errorf("Unable to update %s subscription %v", odfPackageName, err)
```
nb-ohad, line 2120: You are not "unable", you "failed"!
```suggestion
		return fmt.Errorf("Failed to update %s subscription %v", odfPackageName, err)
```

nb-ohad:
@ezio-auditore This PR has a single commit that addresses 2 different concerns 
1. ODF upgrade between minor versions
2. support for encryption with AWS STS 

Please separate the two into different commits

controllers/managedocs_controller.go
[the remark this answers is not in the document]
ezio-auditore: Updated

controllers/managedocs_controller.go
```diff
@@ -68,44 +69,47 @@ const (
...
+	k8sMetricsServiceMonitorName         = "k8s-metrics-service-monitor"
+	alertRelabelConfigSecretName         = "managed-ocs-alert-relabel-config-secret"
+	alertRelabelConfigSecretKey          = "alertrelabelconfig.yaml"
+	onboardingValidationKeySecretName    = "onboarding-ticket-key"
+	convergedDeploymentType              = "converged"
+	consumerDeploymentType               = "consumer"
+	providerDeploymentType               = "provider"
+	rhobsRemoteWriteConfigIdSecretKey    = "prom-remote-write-config-id"
+	rhobsRemoteWriteConfigSecretName     = "prom-remote-write-config-secret"
+	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
```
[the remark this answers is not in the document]
ezio-auditore: Yes, it can be referred here
https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/csi-kms-connection-details.yaml

controllers/managedocs_controller.go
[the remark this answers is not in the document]
ezio-auditore: Yes, it can be referred here
https://github.com/ceph/ceph-csi/blob/7b44054a7ec17e81530d12cffa56070d9bdc6af3/examples/kms/vault/csi-kms-connection-details.yaml#L75

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+
+		encryptionMetadata := struct {
+			EncryptionKMSType string `json:"encryptionKMSType"`
+			SecretName        string `json:"secretName"`
+		}{
+			EncryptionKMSType: "aws-sts-metadata",
+			SecretName:        "tenant-aws-secret", // user needs to create a secret with this name
+		}
+
+		data, _ := json.Marshal(encryptionMetadata)
```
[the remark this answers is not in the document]
ezio-auditore: They use JSON strings inside config map yaml
https://github.com/ceph/ceph-csi/blob/devel/examples/kms/vault/csi-kms-connection-details.yaml

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
```diff
@@ -68,44 +69,47 @@ const (
...
+	alertRelabelConfigSecretKey          = "alertrelabelconfig.yaml"
+	onboardingValidationKeySecretName    = "onboarding-ticket-key"
+	convergedDeploymentType              = "converged"
+	consumerDeploymentType               = "consumer"
+	providerDeploymentType               = "provider"
+	rhobsRemoteWriteConfigIdSecretKey    = "prom-remote-write-config-id"
+	rhobsRemoteWriteConfigSecretName     = "prom-remote-write-config-secret"
+	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
+	desiredODFSubscriptionChannel        = "stable-4.11"
+	odfPackageName                       = "odf-operator"
```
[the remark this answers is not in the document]
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

ezio-auditore:
> @ezio-auditore This PR has a single commit that addresses 2 different concerns
> 
> 1. ODF upgrade between minor versions
> 2. support for encryption with AWS STS
> 
> Please separate the two into different commits

Updated

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
 	return nil
 }
 
+func (r *ManagedOCSReconciler) reconcileCSIKMSConnectionDetailsConfigMap() error {
+	if r.DeploymentType != consumerDeploymentType {
```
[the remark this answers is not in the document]
ezio-auditore: The deployer is creating the config map for csi but the customer needs to create a secret named
`ceph-csi-aws-credentials`
containing the AWS sts role

controllers/managedocs_controller.go
```diff
@@ -68,44 +69,47 @@ const (
...
+	alertRelabelConfigSecretKey          = "alertrelabelconfig.yaml"
+	onboardingValidationKeySecretName    = "onboarding-ticket-key"
+	convergedDeploymentType              = "converged"
+	consumerDeploymentType               = "consumer"
+	providerDeploymentType               = "provider"
+	rhobsRemoteWriteConfigIdSecretKey    = "prom-remote-write-config-id"
+	rhobsRemoteWriteConfigSecretName     = "prom-remote-write-config-secret"
+	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
+	desiredODFSubscriptionChannel        = "stable-4.11"
+	odfOCMBundleName                     = "odf-operator"
```
nb-ohad, line 112: @ezio-auditore There was a question in my last review if this is a package name or a bundle name and you introduced a change to rename it as a bundle name (instead of answering the question). But from the code I can see this is referring to a package name in the subscription. Also, I mentioned in my last review OCM instead of OLM and that was a mistake.

Taking all of that into account this should be named `odfOLMPackageName`

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+		}{
+			EncryptionKMSType: "aws-sts-metadata",
+			SecretName:        "ceph-csi-aws-credentials", // user needs to create a secret with this name
+		}
+
+		data, _ := json.Marshal(encryptionMetadata)
+
+		r.csiKMSConnectionDetailsConfigMap.Data = map[string]string{
+			"aws-sts-metadata-test": string(data), // encryption-with-aws-kms will be used by customer to create storageClassClaim
+		}
```
nb-ohad, line 963: - Why do we need an anonymous struct definition for two string fields? why can't we just use a map?
- If you are ignoring the error for marshal (which makes sense here), why not use `utils.toJsonOrDie()` and inline it? 
```suggestion
		r.csiKMSConnectionDetailsConfigMap.Data = map[string]string{
			"aws-sts-metadata-test": utils.ToJsonOrDie(map[string]string {
				"encryptionKMSType": "aws-sts-metadata",
				"secretName": "ceph-csi-aws-credentials", // user needs to create a secret with this name
			}),
		}
```

controllers/managedocs_controller.go
```diff
@@ -920,6 +936,41 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("Unable to create %s configMap: %v", csiKMSConnectionDetailsConfigMapName, err)
```
nb-ohad, line 968:
```suggestion
		return fmt.Errorf("Failed to create or update %s config map: %v", csiKMSConnectionDetailsConfigMapName, err)
```

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,31 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+	}
+	odfSub := &opv1a1.Subscription{}
+	for index := range subList.Items {
+		item := &subList.Items[index]
+		if item.Spec.Package == odfOCMBundleName {
+			odfSub = item
+			break
+		}
+	}
+	if odfSub == nil {
```
nb-ohad, line 2113: There is a bug here:

This will never be nil because you are initializing it with `odfSub := &opv1a1.Subscription{}` which means it has a default value.

You should initialize it as 
```
var odfSub *opv1a1.Subscription = nil
```
Instead

controllers/managedocs_controller.go
```diff
@@ -2045,3 +2096,31 @@ func (r *ManagedOCSReconciler) setDeviceSetCount(deviceSet *ocsv1.StorageDeviceS
...
+		return fmt.Errorf("Failed to find package name %s", odfOCMBundleName)
+	}
+	if odfSub.Spec.Channel == desiredODFSubscriptionChannel {
+		return nil
+	}
+	r.Log.Info("Initiating ODF minor version upgrade", "current release", odfSub.Spec.Channel, "target release", desiredODFSubscriptionChannel)
```
nb-ohad, line 2119: I think we should be more specific with the information line in the log
```suggestion
	r.Log.Info("Initiating ODF minor version upgrade by switching channels", "current channel", odfSub.Spec.Channel, "new channel", desiredODFSubscriptionChannel)
```

shim/crds/subscriptions.crd.yaml
nb-ohad, line 1: Why do we need this? We already had the assumption that in order to test locally, one needs to install OLM on its k8s cluster

nb-ohad (changes requested):

shim/crds/subscriptions.crd.yaml
ezio-auditore: Without the crd the unit-tests would fail.

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: Updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
ezio-auditore: Sorry, I might have mistook your question. Have updated it accordingly

shim/crds/subscriptions.crd.yaml
nb-ohad: Ok, so the concern is unit tests. In that case, I am ok with this.

controllers/managedocs_controller.go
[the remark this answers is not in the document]
ezio-auditore: Additional details here
https://hackmd.io/66K6Opp8RKGUlRsM62A_rg

controllers/managedocs_controller.go
```diff
@@ -920,6 +929,37 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+		}
+
+		csiKMSConnectionDetailsAsJson := utils.ToJsonOrDie(
+			map[string]string{
+				"encryptionKMSType": "aws-sts-metadata",
+				"secretName":        "ceph-csi-aws-credentials",
+			})
+
+		r.csiKMSConnectionDetailsConfigMap.Data = map[string]string{
+			"aws-sts-metadata-test": string(csiKMSConnectionDetailsAsJson),
```
leelavg, line 957: - it seems whoever creates the storageclassclaim need to use this field? ie, `aws-sts-metadata-test`, if yes, should this be documented anywhere.
- let me know if I misunderstood anything

ezio-auditore: Yes, should be documented, but since this feature is in dev-preview currently there is no requirement for a documentation.

controllers/managedocs_controller.go
```diff
@@ -68,44 +68,47 @@ const (
...
+	alertRelabelConfigSecretKey          = "alertrelabelconfig.yaml"
+	onboardingValidationKeySecretName    = "onboarding-ticket-key"
+	convergedDeploymentType              = "converged"
+	consumerDeploymentType               = "consumer"
+	providerDeploymentType               = "provider"
+	rhobsRemoteWriteConfigIdSecretKey    = "prom-remote-write-config-id"
+	rhobsRemoteWriteConfigSecretName     = "prom-remote-write-config-secret"
+	csiKMSConnectionDetailsConfigMapName = "csi-kms-connection-details"
+	desiredODFSubscriptionChannel        = "stable-4.11"
+	odfOLMPackageName                    = "odf-operator"
```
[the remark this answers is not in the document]
ezio-auditore: Yes, good find. I have added predicates for both.

leelavg (dismissed):

config/metadata/dependencies.yaml
```diff
@@ -6,4 +6,4 @@ dependencies:
 - type: olm.package
   value:
     packageName: odf-operator
-    version: "4.10.9"
+    version: "4.11.5"
```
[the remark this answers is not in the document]
bindrad: This can be resolved

controllers/managedocs_controller.go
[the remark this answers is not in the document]
bindrad: This can be resolved

bindrad (approved):

controllers/managedocs_controller.go
```diff
@@ -263,6 +270,23 @@ func (r *ManagedOCSReconciler) SetupWithManager(mgr ctrl.Manager, ctrlOptions *c
...
+				instance, ok := client.(*opv1a1.Subscription)
+				if !ok {
+					return false
+				}
+
+				// ignore if not a odf-operator subscription
+				if instance.Spec.Package != odfOLMPackageName {
+					return false
+				}
+				return true
```
nb-ohad, line 285: you can simplifyc and prevet ok form being created in func scope
```suggestion
				if instance, ok := client.(*opv1a1.Subscription); ok {
					// ignore if not a odf-operator subscription
					return  instance.Spec.Package != odfOLMPackageName
				} 
				return false
```

controllers/managedocs_controller.go
```diff
@@ -920,6 +959,40 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+
+		awsStsMetadataTestAsString := string(
+			utils.ToJsonOrDie(
+				map[string]string{
+					"encryptionKMSType": "aws-sts-metadata",
+					"secretName":        "ceph-csi-aws-credentials",
+				},
+			),
+		)
+
```
nb-ohad, line 982:
```suggestion
```

utils/utils.go
```diff
@@ -115,3 +116,11 @@ func DeploymentNameFromPodName(podName string) (string, error) {
...
 	return podName[0:i], nil
 }
+
+func ToJsonOrDie(value interface{}) []byte {
+	if bytes, err := json.Marshal(value); err == nil {
+		return bytes
+	} else {
+		panic(err)
+	}
+}
```
nb-ohad, line 126: If you are adding this here, it can be removed from `testsutils` package and have all test files import it from here.
We don't want to maintain 2 copies

nb-ohad (changes requested):

utils/utils.go
ezio-auditore: Have imported testutils package in managedocs_controller.go and used existing toJsonOrDie function. The altrenative would result in a lots of diff

controllers/managedocs_controller.go
ezio-auditore: updated

controllers/managedocs_controller.go
```diff
@@ -920,6 +955,39 @@ func (r *ManagedOCSReconciler) reconcileOnboardingValidationSecret() error {
...
+	}
+	r.Log.Info(fmt.Sprintf("Reconciling %s configMap", csiKMSConnectionDetailsConfigMapName))
+
+	_, err := ctrl.CreateOrUpdate(r.ctx, r.Client, r.csiKMSConnectionDetailsConfigMap, func() error {
+		if err := r.own(r.csiKMSConnectionDetailsConfigMap); err != nil {
+			return err
+		}
+
+		awsStsMetadataTestAsString := string(
+			testutils.ToJsonOrDie(
```
nb-ohad, line 971: `testutils` should not get imported into non-test code. The opposite is ok

nb-ohad (changes requested):

controllers/managedocs_controller.go
ezio-auditore: toJsonOrDie exists in testutils but not used anywhere except sts code.
I have moved the func from testutils to utils.

nb-ohad (approved):

PR #289, opened 2023-05-31
update the egress rules to allow loadBalancer IP on consumer side

rewantsoni:
The consumer needs to access the loadbalancer IP, hence we need to put the rule in the allow list


controllers/managedocs_controller.go
```diff
@@ -1598,6 +1599,27 @@ func (r *ManagedOCSReconciler) reconcileEgressNetworkPolicy() error {
 			)
 		}
 
+		if r.DeploymentType == consumerDeploymentType {
+
+			storageProviderEndpointURL := r.addonParams[storageProviderEndpointKey]
+			storageProviderEndpointHost := strings.Split(storageProviderEndpointURL, ":")[0]
```
nb-ohad, line 1605: This might panic if the split is yielding an empty array in the case of a malformed URL.

controllers/managedocs_controller.go
```diff
@@ -1726,6 +1748,27 @@ func (r *ManagedOCSReconciler) reconcileEgressFirewall() error {
...
```
nb-ohad, line 1754: This might panic if the split is yielding an empty array in the case of a malformed URL.

controllers/managedocs_controller.go
```diff
@@ -1598,6 +1599,27 @@ func (r *ManagedOCSReconciler) reconcileEgressNetworkPolicy() error {
...
+
+			ip := net.ParseIP(storageProviderEndpointHost)
+
+			if ip == nil {
```
nb-ohad, line 1609: I think you wanted `!=` and not `==`.
Also, let's not pollute the func scope if we don't have to.
```suggestion
			if ip := net.ParseIP(storageProviderEndpointHost); ip == nil {
```

controllers/managedocs_controller.go
```diff
@@ -1726,6 +1748,27 @@ func (r *ManagedOCSReconciler) reconcileEgressFirewall() error {
...
```
nb-ohad, line 1758: I think you wanted `!=` and not `==`.
Also, let's not pollute the func scope if we don't have to.
```suggestion
			if ip := net.ParseIP(storageProviderEndpointHost); ip == nil {
```

nb-ohad (changes requested):

controllers/managedocs_controller.go
rewantsoni: I want to check if the provided storageProviderEndpointHost is a nodeIP or a hostName, we want to allow the loadBalancer Endpoint HostName, we can ignore if we have nodeIP as it is already reachable.

controllers/managedocs_controller.go
rewantsoni: I want to check if the provided storageProviderEndpointHost is a nodeIP or a hostName, we want to allow the loadBalancer Endpoint HostName, we can ignore if we have nodeIP as it is already reachable.

controllers/managedocs_controller.go
rewantsoni: Added a condition to return a error if the split is yielding an empty string

controllers/managedocs_controller.go
rewantsoni: Added a condition to return a error if the split is yielding an empty string

